PR #93, opened 2012-12-30
Adds MurmurHashable128 and util-hashing dependency.

sritchie:
This pull request introduces `com.twitter/util-hashing` as a dependency into Algebird, blowing up the jar size by a massive 64k. For you worried aesthetes/goat lovers, `util-hashing` is in maven central and has no extra dependencies.

@johnynek and I rewrote `MurmurHash128` in terms of a new `MurmurHashable128` class.

We also fixed a couple of busted tests and removed an errant println.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -13,6 +13,7 @@ resolvers ++= Seq(
 )
 
 libraryDependencies ++= Seq(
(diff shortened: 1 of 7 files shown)


ianoc:

johnynek:
This has bitrotted quite a lot.

PR #138, opened 2013-03-11
adds algebird-hash module, core depends on it.

johnynek:
This is API breaking, so it it is not a patch version bump when we publish next.

Count-min is still using it's own hash approach. I think we should revisit that, and we should also consider adding something Avi suggested: a HashFamily[T,R] object, which I guess has a Seq[Hashable[T,R]]. The apply on T can return Seq[R]. Comments?


ianoc:

PR #161, opened 2013-05-24
algebird-bufferable

avibryant:
This is just a start, and shouldn't be merged yet, but: defining some Bufferable implementations for Algebird objects so that we don't have to use Kryo for long-term serialized storage.


ianoc:

PR #168, opened 2013-06-21
Linear regression as a group added. It wants to be a vectorspace but I n...

snoble:
...eed to understand how the type system works first. Aaaaand I'll add tests.


ianoc:

mcanlas:
Looks like a stale PR. Can we close?

isnotinvain:
@snoble ?

snoble:
oh totally... (I don't really remember what this does tbh)

isnotinvain:
Thanks! :)

PR #189, opened 2013-08-05
added complement to approximateboolean with property test

davidthomas5412:
Small method added for convenience.


ianoc:

PR #201, opened 2013-09-29
Add the Take semigroup

johnynek:
An operation we probably want in https://github.com/twitter/summingbird


ianoc:

PR #220, opened 2013-10-14
remove seed variable and rename k to hashIndex

sritchie:


ianoc:

PR #222, opened 2013-10-16
Make seed of min hasher repeatable

tsdeng:
Use index of hash function as the seed of murmur hash

Currently, in MinHash we use randomGenerator to generate seed for murmurHash. 

We want stable/repeatable behavior, so that  
As long as the user provide the same parameter for constructing the monoid, the hash functions should be consistent


ianoc:

PR #294, opened 2014-04-04
Adding MutableSumAll and Semigroup#bufferedSumOption,statefulSummer

### LINKED ISSUE #291: implement standard pattern for updateInto
HyperLogLog has an updateInto method which updates a mutable buffer representing its current state, instead of doing immutable operations. This allows for an efficient implementation of sumOption. It would be helpful to standardize this so that we could use it from eg SummingCache. see #290 

https://github.com/twitter/algebird/blob/develop/algebird-core/src/main/scala/com/twitter/algebird/HyperLogLog.scala#L241

jnievelt:
Refactoring some Semigroup#sumOption optimizations to use these.

Closes #291

These should be getting test coverage indirectly, but I'm open to suggestions for additional tests to add.


ianoc:

PR #295, opened 2014-04-05
HyperLogLogSeries

avibryant:
This data structure can produce a HyperLogLog counter for any window into the past, using a constant factor more space than HyperLogLog.

For each hash bucket, rather than keeping a single max RhoW value, it keeps every RhoW value it has seen, and the max timestamp where it saw that value. This allows it to reconstruct an HLL as it would be had it started at zero at any given point in the past, and seen the same updates this structure has seen.

The usage is like this:

````scala
vall hllSeriesMonoid = new HyperLogLogSeriesMonoid(bits)

val series = examples
                       .map{case (bytes, timestamp) => hllSeriesMonoid.create(bytes, timestamp)}
                       .reduce{hllSeriesMonoid.plus(_,_)}

val estimate1 = series.since(timestamp1).toHLL.estimatedSize
val estimate2 = series.since(timestamp2).toHLL.estimatedSize
````

### DIFF
--- a/algebird-core/src/main/scala/com/twitter/algebird/HyperLogLogSeries.scala
+++ b/algebird-core/src/main/scala/com/twitter/algebird/HyperLogLogSeries.scala
@@ -16,7 +16,26 @@ limitations under the License.
 
 package com.twitter.algebird
 
+/**
+ * HLLSeries can produce a HyperLogLog counter for any window into the past,
+ * using a constant factor more space than HyperLogLog.
+ *
+ * For each hash bucket, rather than keeping a single max RhoW value, it keeps
+ * every RhoW value it has seen, and the max timestamp where it saw that value.
+ * This allows it to reconstruct an HLL as it would be had it started at zero at
+ * any given point in the past, and seen the same updates this structure has seen.
+ *
+ * @param bits The number of bits to use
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

**Comment by DanielleSucher** [comment]
Ping! Anything still missing here before y'all would be comfortable merging?

File: algebird-core/src/main/scala/com/twitter/algebird/HyperLogLogSeries.scala [hunk 1]
```diff
@@ -0,0 +1,70 @@
...
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package com.twitter.algebird
+
+case class HLLSeries(bits: Int, rows: Vector[Map[Int, Long]]) {
```
johnynek: We need more docs here. I have to read carefully to remind myself of the intent and use of this code. Can you give some comments that will show up in the scaladoc?

DanielleSucher:
Righto, comments added. ^^

johnynek:
Thanks for the comments and the usage notes!

PR #310, opened 2014-04-22
more elegant implementation of sumOption in eventually monoid

julienledem:
This should be more efficient as it avoids materializing lists and batching.
It will also make batching in other places more predictable.


ianoc:

PR #312, opened 2014-04-24
make SummingCache a trait and provide multiple implementations that can be picked based on the monoid implementation

julienledem:
In particular whether plus or sumOption is more efficient.


ianoc:

PR #327, opened 2014-07-08
Add speculative approach for HLL intersection (in test only)

johnynek:
Rather than taking the byte-wise max, it seems intuitive that byte-wise min should work for intersection. Playing around with it, it seems to give reasonable results in some cases.

There is no code for this outside of the test. We should either:

1) do a careful numerical study of the error of this method (or other related methods).

2) find a proof that the error is bounded. It seems tractable. I'll try to play with it analytically.


ianoc:

PR #333, opened 2014-08-08
Adds Interval Tree Clocks and Vector Clocks with an algebraic approach

johnynek:
I've been geeking out on this for the past few nights.

1) implementing Interval Tree Clock was really fun.

2) I noticed a new concept: a partial semigroup. It seems useful to build operations that are robust to duplication.

3) I think with this, we can make storehaus work with Partial semigroups, which means if we store a value with type (C, T) where we have a clock for C and a semigroup for T we can get duplication tolerance. This is still a little vague to me, but I'd appreciate feedback.


ianoc:

PR #363, opened 2014-11-30
Fixes the base properties like MonoidLaws that weren't running.

### LINKED ISSUE #357: (commutative)MonoidLaws Tests pass when they shouldn't
This monoid:
```scala
class RandomMonoid[T: ClassTag](implicit semi: Semigroup[T]) extends Monoid[Array[T]] {
  override def zero = Array[T]()
  override def plus(left: Array[T], right: Array[T]) = {
    if (scala.util.Random.nextBoolean) right else left
  }
}
```

Passes this test:
```scala
 property("Random Monoid laws") {
    monoidLaws[Array[String]]
  }
```

ianoc:
Got broken in the scalatest conversion. This brings them back. Fixes #357

### DIFF
--- a/algebird-test/src/main/scala/com/twitter/algebird/BaseProperties.scala
+++ b/algebird-test/src/main/scala/com/twitter/algebird/BaseProperties.scala
@@ -17,7 +17,7 @@ limitations under the License.
 package com.twitter.algebird
 
 import org.scalacheck.Arbitrary
-import org.scalacheck.Prop
+import org.scalatest.prop.Checkers
 import org.scalacheck.Prop.forAll
 import scala.math.Equiv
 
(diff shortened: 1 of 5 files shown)


bkirwi:
For reason 2), there's a way to get similarly precise output from ScalaCheck using 'property labels'. You annotate each test in your property roughly like this:

```scala
("commutative" |: a + b == b + a) &&
("associative" |: a + (b + c) == (a + b) + c) 
```

If a condition fails, the offending label should show up in the ScalaCheck output. (I'm not familiar enough with scalatest to know if it surfaces them correctly, though.) The docs are in the [User Guide](https://github.com/rickynils/scalacheck/wiki/User-Guide), under 'Labeling Properties'.

ianoc:

sritchie:
Closed - I went through in #584 and made sure all tests are running.

PR #385, opened 2014-12-19
update qtree to compress more aggressively

jinlee:
The current qtree compress implementation is too conservative. In the current implementation new nodes with small counts can exist as long as the neighboring subtree has a big enough count.

I've updated the implementation to be more aligned with the algorithm outlined here: http://www.cs.virginia.edu/~son/cs851/papers/ucsb.sensys04.pdf

I've checked my solution by playing around in `./sbt algebird-core/console` and via `./sbt algebird-test/test`

This is also my first pull request for algebird, please let me know if I should have taken some extra steps.

### DIFF
--- a/algebird-core/src/main/scala/com/twitter/algebird/QTree.scala
+++ b/algebird-core/src/main/scala/com/twitter/algebird/QTree.scala
@@ -210,7 +210,7 @@ case class QTree[A](
     val (lowerResult, upperResult) =
       mapChildrenWithDefault[(Option[QTree[A]], Boolean)]((None, false)){ _.prune(minCount) }
     val parent =
-      if (lowerResult._2 || lowerResult._2)
+      if (lowerResult._2 || upperResult._2)
         copy(lowerChild = lowerResult._1, upperChild = upperResult._1)
       else
         this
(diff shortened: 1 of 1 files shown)


johnynek:
If we can get a benchmark using our benchmarking here, and compare the new code to the old code, we can merge this.

We definitely want it to be faster, but without benchmarks we are conservative on perf changes.

jinslee:
Sorry I haven't had much time to follow up on this.

But yes, as Avi mentioned, I was trying to get the QTree implementation to mirror the q-digest closely. In particular, I was hoping to get the same space constraint outlined in the paper. I am pretty sure that my implementation will run slower than the original code.

If keeping the QTree implementation fast is crucial, I'll be closing this request.

avibryant:
Can I suggest that instead of just closing it, we do some benchmarking? I'm already pretty worried about the performance of QTree, but it may be that this doesn't slow it down much compared to some more aggressive optimizations we could do (in particular, I suspect we shouldn't be compressing on every addition to the tree). So there might be a balance that trades off this change against something else which is preferable.

ianoc:

PR #387, opened 2014-12-29
Freshen up Applicative instances a bit

johnynek:
Cleaning up MonadSemigroup and friends, which always just used Applicative, but we didn't have that type class then.


johnynek:
@isnotinvain please review.

isnotinvain:
LGTM!

Issue #345, opened 2014-09-12
A generic version of CountMinSketch[T] to decouple CMS from Long?


miguno:
_This is a question about a) whether this is a good idea in general and b) whether you'd be interested in a PR with draft code._

**"Why should I care about this issue?"**

If we had a way to use CMS with "bigger" number types than `Long` we would be able to cover use cases that require counting more distinct elements than can fit into the range of Long (which is `2^64` bits).  This would allow, for instance, to probabilistically count IPv6 addresses, which is currently not possible.

**Background**

The current CMS implementation ([CountMinSketch.scala](https://github.com/twitter/algebird/blob/develop/algebird-core/src/main/scala/com/twitter/algebird/CountMinSketch.scala)) is backed by a `Long`, which means that CMS will count the occurrences of Long instances.  As a user I must ensure that I can properly encode as well as decode my "real" data entities to and from `Long` (and this mapping should be bijective).

What does this mean in practice?  `Long` has `2^64` bits (= you can count up to 1.8E19 distinct elements), which means means you can't use the current CMS implementation if what you are trying to count has more than `2^64` distinct elements.

Unfortunately, there are such use cases where the address space of `Long` is not sufficient.  One example is counting IPv6 addresses, which have an address space of `2^128`.  (IPv4 addresses can be counted as they are `2^32` bits).  Similarly, you may want to count the occurrences of unique strings, where one option is to encode a string as a `BigInt` (e.g. via `BigInt(s.getBytes("UTF-8")`), and then perform the probabilistic counts via a `CMS[BigInt]` instance (though this is but one option to count strings of course).

**Alternatives to CMS in Algebird**

The one option I found is `SketchMap`, which is "a generalized version of Count-Min Sketch":

> A Sketch Map is a generalized version of the Count-Min Sketch that is an approximation of Map[K, V] that stores reference to top heavy hitters. The Sketch Map can approximate the sums of any summable value that has a monoid.

Unfortunately, SketchMap seems to be really slow.  In a simple benchmark -- all other things being equal -- the updates/s were `0.3k/s` (SketchMap with K=BigInt and V=Long) vs. `150k/s` (CMS with hardcoded V=Long).

Code snippet for SketchMap (similar setup for CMS):

```scala
    type K = BigInt
    type V = Long

    val SM_MONOID = {
      val eps = 0.005
      val delta = 1E-8
      val seed = 1
      val heavyHittersCount = 1000
[...]

PR #390, opened 2014-12-31
CMSItems

avibryant:
Adds a `CMSItems[K]` implementation of `CMS[K]` which, like `CMSItem[K]`, is exact, but for a small set of keys instead of just one.

You can add `CMSItem` and `CMSItem` or `CMSItems` and `CMSItems` and get a `CMSItems`. At a certain threshold (currently set to 100 items), this will transition to a `CMSInstance[K]`.

I haven't updated the TopN or TopPct variants, just the basic CMS.

I have a specific need for this, but it seems like a generally useful optimization. It could *probably* entirely replace `CMSItem` but I wouldn't want to do that without benchmarking first, which I have not done.

Thoughts @johnynek @miguno?

### DIFF
--- a/algebird-core/src/main/scala/com/twitter/algebird/CountMinSketch.scala
+++ b/algebird-core/src/main/scala/com/twitter/algebird/CountMinSketch.scala
@@ -76,6 +76,7 @@ import scala.collection.immutable.SortedSet
  *        (an interval that depends on `eps`) around the truth.
  * @param seed  A seed to initialize the random number generator used to create the pairwise independent
  *        hash functions.
+ * @param maxExactCount The largest number of keys that will be stored directly, instead of in a sketch.
  * @tparam K The type used to identify the elements to be counted.  For example, if you want to count the occurrence of
  *           user names, you could map each username to a unique numeric ID expressed as a `Long`, and then count the
  *           occurrences of those `Long`s with a CMS of type `K=Long`.  Note that this mapping between the elements of
@@ -87,11 +88,11 @@ import scala.collection.immutable.SortedSet
  *           include Spire's `SafeLong` and `Numerical` data types (https://github.com/non/spire), though Algebird does
  *           not include the required implicits for CMS-hashing (cf. [[CMSHasherImplicits]].
  */
-class CMSMonoid[K: Ordering: CMSHasher](eps: Double, delta: Double, seed: Int) extends Monoid[CMS[K]] {
+class CMSMonoid[K: Ordering: CMSHasher](eps: Double, delta: Double, seed: Int, maxExactCount: Int = 100) extends Monoid[CMS[K]] {
 
   val params = {
     val hashes: Seq[CMSHash[K]] = CMSFunctions.generateHashes(eps, delta, seed)
-    CMSParams(hashes, eps, delta)
+    CMSParams(hashes, eps, delta, maxExactCount)
   }
 
   val zero: CMS[K] = CMSZero[K](params)
@@ -114,7 +115,7 @@ class CMSMonoid[K: Ordering: CMSHasher](eps: Double, delta: Double, seed: Int) e
   /**
    * Creates a sketch out of multiple items.
    */
-  def create(data: Seq[K]): CMS[K] = CMSItems[K](data.toList, params).toCMS
+  def create(data: Seq[K]): CMS[K] = CMSItems[K](data.toList, data.size, params).toCMS
 
 }
 
@@ -138,14 +139,14 @@ case class CMSAggregator[K](cmsMonoid: CMSMonoid[K]) extends MonoidAggregator[K,
  * @param eps One-sided error bound on the error of each point query, i.e. frequency estimate.
  * @param delta A bound on the probability that a query estimate does not lie within some small interval
  *              (an interval that depends on `eps`) around the truth.
+ * @param maxExactCount The largest number of keys that will be stored directly, instead of in a sketch.
  * @tparam K The type used to identify the elements to be counted.
  */
-case class CMSParams[K](hashes: Seq[CMSHash[K]], eps: Double, delta: Double) {
+case class CMSParams[K](hashes: Seq[CMSHash[K]], eps: Double, delta: Double, maxExactCount: Int = 0) {
 
   require(0 < eps && eps < 1, "eps must lie in (0, 1)")
   require(0 < delta && delta < 1, "delta must lie in (0, 1)")
   require(hashes.size >= CMSFunctions.depth(delta), s"we require at least ${CMSFunctions.depth(delta)} hash functions")
-
 }
 
 /**
@@ -436,7 +437,7 @@ case class CMSItem[K: Ordering](item: K, override val params: CMSParams[K]) exte
   override def ++(other: CMS[K]): CMS[K] = {
     other match {
       case other: CMSZero[_] => this
-      case other: CMSItem[K] => CMSItems(List(item, other.item), params)
+      case other: CMSItem[K] => CMSItems(List(item, other.item), 2L, params)
       case other: CMSItems[K] => other + item
       case other: CMSInstance[K] => other + item
     }
@@ -451,17 +452,26 @@ case class CMSItem[K: Ordering](item: K, override val params: CMSParams[K]) exte
 /**
  * Used for holding a small set of elements.
  */
-case class CMSItems[K: Ordering](items: List[K], override val params: CMSParams[K]) extends CMS[K](params) {
+case class CMSItems[K: Ordering](items: List[K], override val totalCount: Long, override val params: CMSParams[K]) extends CMS[K](params) {
 
-  override val totalCount: Long = items.size
-
-  override def +(x: K, count: Long): CMS[K] = CMSItems(x :: items, params).toCMS
+  override def +(x: K, count: Long): CMS[K] = CMSItems(x :: items, totalCount + 1, params).toCMS
 
   override def ++(other: CMS[K]): CMS[K] = {
     other match {
       case other: CMSZero[_] => this
       case other: CMSItem[K] => this + other.item
-      case other: CMSItems[K] => CMSItems(items ++ other.items, params).toCMS
+      case other: CMSItems[K] => {
+        val (larger, smaller) =
+          if (totalCount > other.totalCount)
+            (this, other)
+          else
+            (other, this)
+
+        CMSItems(
+          larger.items.reverse ::: (smaller.items),
+          totalCount + other.totalCount,
+          params).toCMS
+      }
       case other: CMSInstance[K] => toCMSInstance ++ other
     }
   }
@@ -471,23 +481,16 @@ case class CMSItems[K: Ordering](items: List[K], override val params: CMSParams[
   override def innerProduct(other: CMS[K]): Approximate[Long] =
     items.map{ item => other.frequency(item) }.reduce{ _ + _ }
 
-  def toCMS: CMS[K] = if (items.size > CMSItems.MaxSize) toCMSInstance else this
+  def toCMS: CMS[K] = if (items.size > params.maxExactCount) toCMSInstance else this
   def toCMSInstance: CMSInstance[K] = {
     val vectors = params.hashes.map{ hash =>
-      items.foldLeft(Vector.fill[Long](width)(0L)){
-        case (vector, item) =>
-          val pos = hash(item)
-          vector.updated(pos, vector(pos) + 1L)
-      }
+      val array = Array.fill(width)(0L)
+      items.foreach{ item => array(hash(item)) += 1L }
+      array.toVector
     }
 
     CMSInstance(CMSInstance.CountsTable[K](vectors.toVector), items.size, params)
   }
-}
-
-object CMSItems {
-  //arbitrarily chosen
-  val MaxSize = 100
 }
 
 /**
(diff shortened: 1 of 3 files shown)


miguno:
> by all means let's benchmark, is there a HOWTO for algebird-caliper somewhere? If not I'm sure I can dive in and figure it out.

We do not have a full HOWTO for algebird-caliper yet but I did write a brief README:
https://github.com/twitter/algebird/blob/develop/algebird-caliper/README.md

See [CMSBenchmark](https://github.com/twitter/algebird/blob/develop/algebird-caliper/src/test/scala/com/twitter/algebird/caliper/CMSBenchmark.scala) for an example.  I'd suggest to first run this benchmark from the CLI (see README) to see what it is actually doing, and then take a look at the benchmark code on how it's done.  It's essentially a simple, parameterizable micro-benchmark ("Call this function N-times with these combinations of parameters").

Note: algebird-caliper uses the sbt plugin [cappi](https://github.com/softprops/capp) to run caliper benchmarks.  Cappi does not support the latest caliper version yet (1.0-beta-1), so we must use the feature set provided by IIRC caliper 0.6-something ([see further details](https://github.com/twitter/algebird/blob/develop/algebird-caliper/src/test/scala/com/twitter/algebird/caliper/CMSBenchmark.scala#L11-L14)).

HTH!

algebird-core/src/main/scala/com/twitter/algebird/CountMinSketch.scala
```diff
@@ -448,6 +449,48 @@ case class CMSItem[K: Ordering](item: K, override val params: CMSParams[K]) exte
 }
 
 /**
+ * Used for holding a small set of elements.
+ */
+case class CMSItems[K: Ordering](items: List[K], override val params: CMSParams[K]) extends CMS[K](params) {
+
+  override val totalCount: Long = items.size
```
avibryant: Items is not a sparse counting vector (although that might make sense), but simply a list of keys, which may contain duplicates. So `items.size` is the correct way to get the totalCount - this will be equivalent to the L1-norm of the counting vector once it's transformed into that.

avibryant:
@johnynek, I think this latest commit addresses your performance concerns.
@miguno I moved MaxItems to CMSParams and called it `maxExactCount`. The API involving configuring CMS and its variations has a very large surface area and I'm not super happy with how this new param is integrated: the only reasonable way to set it explicitly is to create a `new CMSMonoid`, otherwise you'll generally get the default. OTOH, I think I'm not very happy with CMSParams in general - if we're pulling out a params object, why aren't we exposing it in the API for constructing these monoids and aggregators etc? But I don't want to get embroiled in a major refactoring of that right now, so this seems like the minimal-impact way to start.

I'll look at benchmarks next.

algebird-core/src/main/scala/com/twitter/algebird/CountMinSketch.scala
```diff
@@ -448,6 +450,50 @@ case class CMSItem[K: Ordering](item: K, override val params: CMSParams[K]) exte
 }
 
 /**
+ * Used for holding a small set of elements.
+ */
+case class CMSItems[K: Ordering](items: List[K], override val totalCount: Long, override val params: CMSParams[K]) extends CMS[K](params) {
+
+  override def +(x: K, count: Long): CMS[K] = CMSItems(x :: items, totalCount + 1, params).toCMS
```
miguno: It looks as if `count` is not being taken into account.  Shouldn't `+(x, 4)` result in the count of `x` being increased by 4?

This was the reason why I asked the `totalCount` question about `items.size` vs. `items.sum` (the L1-norm) -- size would be equivalent to sum only if we restrict `+` to `+(x, 1)`.

miguno:
CMS configuration issues:  Yeah, I definitely agree with your comments that the current way of configuring CMS could be better.  The main reason (on my side) this hasn't really changed from the original code was to limit the changes being introduced when we turned `CMS[Long]` into `CMS[K]` (see #345 and #354) which is a similar "I'm not touching it right now" reason as you mentioned;  well, at least it is a bit clearer in the latest CMS version about which parameters are needed where, which should help with deciding on how to improve the setup.  Also, CMS up to this point has had the built-in assumption that the widths and depths of CMS instances are identical when they are being merged via `++` (the only "free" parameter has been related to heavy hitters, and that parameter is not a "core" CMS parameter), so from what I understand we haven't have a need yet to fiddle with CMS parameters after instances had been constructed.

If you have any ideas on how to simplify and/or improve the CMS configuration, I'd definitely appreciate it.  IMHO the configuration and usability aspects of this code is as important as the magic that's going on behind the scenes, so personally I wouldn't mind spending some time to make sure the new functionality you have been working on is made available to the user in a better fashion -- but that could be a separate issue/PR as you are pointing out.

algebird-test/src/test/scala/com/twitter/algebird/CountMinSketchTest.scala
```diff
@@ -21,25 +21,25 @@ class CmsLaws extends PropSpec with PropertyChecks with Matchers {
   }
 
   property("CountMinSketch[Short] is a Monoid") {
-    implicit val cmsMonoid = CMS.monoid[Short](EPS, DELTA, SEED)
+    implicit val cmsMonoid = new CMSMonoid[Short](EPS, DELTA, SEED, 0)
```
miguno: Can we use a const for `0` to explain the magic number, just like we do for EPS, DELTA, and SEED?

algebird-core/src/main/scala/com/twitter/algebird/CountMinSketch.scala
```diff
@@ -138,14 +139,14 @@ case class CMSAggregator[K](cmsMonoid: CMSMonoid[K]) extends MonoidAggregator[K,
  * @param eps One-sided error bound on the error of each point query, i.e. frequency estimate.
  * @param delta A bound on the probability that a query estimate does not lie within some small interval
  *              (an interval that depends on `eps`) around the truth.
+ * @param maxExactCount The largest number of keys that will be stored directly, instead of in a sketch.
  * @tparam K The type used to identify the elements to be counted.
  */
-case class CMSParams[K](hashes: Seq[CMSHash[K]], eps: Double, delta: Double) {
+case class CMSParams[K](hashes: Seq[CMSHash[K]], eps: Double, delta: Double, maxExactCount: Int = 0) {
```
miguno: We should add a `require(maxExactCount >= 0, "maximum exact count must not be negative")`.

algebird-core/src/main/scala/com/twitter/algebird/CountMinSketch.scala
```diff
@@ -436,7 +437,8 @@ case class CMSItem[K: Ordering](item: K, override val params: CMSParams[K]) exte
   override def ++(other: CMS[K]): CMS[K] = {
     other match {
       case other: CMSZero[_] => this
-      case other: CMSItem[K] => CMSInstance[K](params) + item + other.item
+      case other: CMSItem[K] => CMSItems(List(item, other.item), 2L, params)
```
miguno: When `maxExactCount == 0` what should happen to this `case`?  That is, when it is zero my understanding is that the user wants to disable the exact counting optimization based on `CMSItems`, so this `case` statement should rather be "if (zero) oldCode else newCode" I think.

That being said, it looks to me as if we'd have to thread such if-else switches for maxExactCount through multiple parts of the CMS code, which is an indication that there might be a better pattern to enable/disable the new optimization functionality?

avibryant:
Some initial benchmarks: it makes relatively little difference when summing the first 1000 integers (though using a size 100 CMSItems is a modest performance gain):

```
[info]       1000        PlusOfFirstNIntegersWithItemsCms   0.1        0  2787.1 =
[info]       1000        PlusOfFirstNIntegersWithItemsCms   0.1       10  2822.8 =
[info]       1000        PlusOfFirstNIntegersWithItemsCms   0.1      100  2539.4 =
[info]       1000        PlusOfFirstNIntegersWithItemsCms 0.005        0  3653.2 ==
[info]       1000        PlusOfFirstNIntegersWithItemsCms 0.005       10  3656.0 ==
[info]       1000        PlusOfFirstNIntegersWithItemsCms 0.005      100  3337.8 ==
```

Not too surprisingly, if you're only summing 100 integers, and you have a size 100 CMSItems, it makes a huge difference:

```
[info]        100        PlusOfFirstNIntegersWithItemsCms   0.1        0   278.5 =
[info]        100        PlusOfFirstNIntegersWithItemsCms   0.1       10   257.7 =
[info]        100        PlusOfFirstNIntegersWithItemsCms   0.1      100    14.5 =
[info]        100        PlusOfFirstNIntegersWithItemsCms 0.005        0   436.0 =
[info]        100        PlusOfFirstNIntegersWithItemsCms 0.005       10   411.5 =
[info]        100        PlusOfFirstNIntegersWithItemsCms 0.005      100    13.8 =
```

algebird-core/src/main/scala/com/twitter/algebird/CountMinSketch.scala
avibryant: We should probably just add `.toCMS` here. If `maxExactCount == 0`, that will trigger a conversion to `CMSInstance`.

algebird-core/src/main/scala/com/twitter/algebird/CountMinSketch.scala
avibryant: Thanks, good catch.

johnynek:
So the speed win here is likely when you have a significant number of CMS instances that are going to be very small, which is common in heavy-tailed distributions. In that case, you cut the small instance cost down by a factor of 10 or more. Seems worth it for that.

Also, for the accuracy win you mention in dot-product.

algebird-core/src/main/scala/com/twitter/algebird/CountMinSketch.scala
```diff
@@ -448,6 +450,50 @@ case class CMSItem[K: Ordering](item: K, override val params: CMSParams[K]) exte
...
+      case other: CMSItem[K] => this + other.item
+      case other: CMSItems[K] => {
+        val (larger, smaller) =
+          if (totalCount > other.totalCount)
+            (this, other)
+          else
+            (other, this)
+
+        CMSItems(
+          larger.items.reverse ::: (smaller.items),
```
johnynek: 1. don't we want the method `reverse_:::` which combines these two to save cost?
2. don't we want the smaller one on the left so we push the fewer items onto the bigger?

johnynek:
Can you commit the benchmarks?

algebird-core/src/main/scala/com/twitter/algebird/CountMinSketch.scala
miguno: >  (2) don't we want the smaller one on the left so we push the fewer items onto the bigger?

The code is correctly pushing the smaller list onto the larger list:

```scala
scala> val larger = List(1,2,3,4,5,6,7,8,9)
larger: List[Int] = List(1, 2, 3, 4, 5, 6, 7, 8, 9)

scala> val smaller = List(10,20,30)
smaller: List[Int] = List(10, 20, 30)

scala> larger.reverse_:::(smaller)
res4: List[Int] = List(30, 20, 10, 1, 2, 3, 4, 5, 6, 7, 8, 9)
```

> (1) don't we want the method reverse_::: which combines these two to save cost?

And I was surprised to learn that apparently a List's `reverse_:::` is faster than `++` or `:::` because it leverages structural sharing (its loop does [`::`](https://github.com/scala/scala/blob/v2.11.4/src/library/scala/collection/immutable/List.scala#L110)) whereas e.g. [`:::`](https://github.com/scala/scala/blob/v2.11.4/src/library/scala/collection/immutable/List.scala#L124) creates a `ListBuffer` via [`++=`](https://github.com/scala/scala/blob/v2.11.4/src/library/scala/collection/mutable/ListBuffer.scala#L181) and [`prependToList`](https://github.com/scala/scala/blob/v2.11.4/src/library/scala/collection/mutable/ListBuffer.scala#L316).  I didn't expect that, though I haven't yet verified my understanding with a quick caliper benchmark. :-)

miguno:
@johnynek wrote:
> So the speed win here is likely when you have a significant number of CMS instances that are going to be very small, which is common in heavy-tailed distributions. In that case, you cut the small instance cost down by a factor of 10 or more. Seems worth it for that.

First let me say that I agree, too, but I have a question:  Could you elaborate on why heavy-tailed distributions would commonly imply a significant (large) number of CMS instances?  Doesn't this depend on how you "partition" the input data stream into CMS instances?

@ianoc made a similar reference to how this optimization would help to aggregate the long tail;  I don't see yet how you guys are segregating the long tail from the heavy hitters (if they are lumped together then the number of elements is quickly `> 100`, or you need to create so many N partitions that 1/N-th of the data is `<= 100`) -- I assume you are talking about scenarios where one _is not_ using e.g. random shuffling to spread the load of processing?

ianoc:
@miguno the primary win for space + perf:
-> Adding Maps of few elements is much faster than combining the CMS structures
-> Those maps wind up being smaller serialized than the CMS structures too


And the reference/difference here i think what I was getting at is:

Often we in a dataset have a primary key in a space O(10^8) , say user ID's. Then we have a secondary index, in a space O(10^12), say url's. So we might use a setup like this to track the main referring URL's for all users, with a type of (Long, CMS[String]). For most users there will in a given time window before we age them out probably very few of these URL's. So moving to a CMSInstance at size == 2 is very expensive if the size never gets bigger than say 5.

johnynek:
We're about to publish. Is this ready to merge?

@avibryant can you merge with develop? @miguno what do you say? does this have your :+1: ?

avibryant:
IMO let's get this in the next one, I think I might want to take another run at it when I have time that is more invasive but makes me happier with the result.

Unless someone really wants this in the meantime?

miguno:
I remember there's at least one functional issue remaining where the new code only supports incrementing by 1 but not by N (see my comment above).  We also don't have the tests yet AFAIK.  So as Avi said maybe we should wait one more round and give it some more work?

johnynek:
@avibryant you want to get this synced and try to get it in 0.10?

ianoc:

johnynek:
I think this was superseded by #464

PR #391, opened 2015-01-02
add averageFrom to DecayedValue

avibryant:
DecayedValue's `average` method is useful but it gives artificially low results if the counter is only a small number of half lives old. `averageFrom` accounts for the fact that you might have a fixed bound in the past for your oldest value.

It currently requires passing an explicit endTime. I'm not sure if it's worth having a version which derives the endTime from the scaledTime (or possibly *only* having that version, and maybe adding a separate `asOf` method which can be used in conjunction... this seems better than having `valueAsOf` on the monoid, anyway).

### DIFF
--- a/algebird-core/src/main/scala/com/twitter/algebird/DecayedValue.scala
+++ b/algebird-core/src/main/scala/com/twitter/algebird/DecayedValue.scala
@@ -73,4 +73,19 @@ case class DecayedValue(value: Double, scaledTime: Double) extends Ordered[Decay
     val normalization = halfLife / math.log(2)
     value / normalization
   }
+
+  /*
+  * Moving average assuming the signal started at zero a fixed point in the past.
+  * This normalizes by the integral of exp(-t(ln(2))/halflife) from 0 to (endTime - startTime).
+  */
+  def averageFrom(halfLife: Double, startTime: Double, endTime: Double) = {
+    if(endTime > startTime) {
+      val asOfEndTime = DecayedValue.scale(DecayedValue.build(0, endTime, halfLife), this, 0.0)
+      val timeDelta = startTime - endTime
+      val normalization = halfLife * (1 - math.pow(2, timeDelta / halfLife)) / math.log(2)
+      asOfEndTime.value / normalization
+    } else {
+      0.0
+    }
+  }
 }
--- a/algebird-core/src/main/scala/com/twitter/algebird/DecayedValue.scala
+++ b/algebird-core/src/main/scala/com/twitter/algebird/DecayedValue.scala
@@ -79,7 +79,7 @@ case class DecayedValue(value: Double, scaledTime: Double) extends Ordered[Decay
   * This normalizes by the integral of exp(-t(ln(2))/halflife) from 0 to (endTime - startTime).
   */
   def averageFrom(halfLife: Double, startTime: Double, endTime: Double) = {
-    if(endTime > startTime) {
+    if (endTime > startTime) {
       val asOfEndTime = DecayedValue.scale(DecayedValue.build(0, endTime, halfLife), this, 0.0)
       val timeDelta = startTime - endTime
       val normalization = halfLife * (1 - math.pow(2, timeDelta / halfLife)) / math.log(2)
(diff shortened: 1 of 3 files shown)


avibryant:
I suppose one natural next step is to add a `BoundedDecayedValue` which explicitly carries around its min timestamp.

algebird-core/src/main/scala/com/twitter/algebird/DecayedValue.scala
```diff
@@ -73,4 +73,19 @@ case class DecayedValue(value: Double, scaledTime: Double) extends Ordered[Decay
     val normalization = halfLife / math.log(2)
     value / normalization
   }
+
+  /*
+  * Moving average assuming the signal started at zero a fixed point in the past.
+  * This normalizes by the integral of exp(-t(ln(2))/halflife) from 0 to (endTime - startTime).
+  */
+  def averageFrom(halfLife: Double, startTime: Double, endTime: Double) = {
```
johnynek: can we add some tests for this?

algebird-core/src/main/scala/com/twitter/algebird/DecayedValue.scala
```diff
@@ -73,4 +73,19 @@ case class DecayedValue(value: Double, scaledTime: Double) extends Ordered[Decay
...
+    if(endTime > startTime) {
+      val asOfEndTime = DecayedValue.scale(DecayedValue.build(0, endTime, halfLife), this, 0.0)
```
johnynek: can't you use `valueAsOf(this, halfLife, endTime)` here to be a bit more clear?

algebird-core/src/main/scala/com/twitter/algebird/DecayedValue.scala
johnynek: so `averageFrom(h, 0.0, e) == average(h)` right?

avibryant: working on tests, but as I'm doing them I'm becoming suspicious of the original math. For example: the average (whether using `average` or `averageFrom`) of, say, `f(t) = t` is > t, and it diverges as `t` increases. That seems like a strange thing for an average to do.

avibryant: `averageFrom(h, 0.0, e)` can be > `average(h)` if `e < n*h` for small `n`. That's intended; the `average` goes to time `-Inifnity`, whereas `averageFrom` does not.

However, `averageFrom(h, 0.0, e)` will converge to `average(h)` as `e` goes to infinity, which I think is the actual property you want to test.

algebird-core/src/main/scala/com/twitter/algebird/DecayedValue.scala
avibryant: That's a method on the monoid and you don't have access to that here.

algebird-core/src/main/scala/com/twitter/algebird/DecayedValue.scala
avibryant: (re being suspicious of the math - I definitely come up with the same answer when I do it from first principles, but I struggle with the intuition of an average being higher than any values it's seen so far; I'm sure this is something to do with discrete vs. continuous...)

avibryant:
@johnynek ok, so after doing a little playing around numerically: in the discrete case (by which I mean, when your half-life is some small multiple of the resolution of your timestamps... eg you only have data at 1min resolution and you have 1hr halflives, etc), it seems to be better to multiply by `1.0 - math.pow(2, -1.0/halfLife)` as the normalization. This converges to the current normalization of dividing by `halfLife/math.log(2)` as `halfLife` gets large (that is: `1.0 - math.pow(2, -1.0/halfLife) - math.log(2)/halfLife` goes to 0.0 as `halfLife` goes to infinity).

That would be for `discreteAverage`, I haven't figured out what the right `discreteAverageFrom` would be.

Does that seem sensible to add, to you?

avibryant:
@johnynek added `discreteAverage` and some tests that demonstrate where the three types of average method are useful.

avibryant:
@johnynek oops, somehow lost `discreteAverage` but I've put it back in. Anything more you'd like to see before a merge?

johnynek:
@isnotinvain let's get this in before the next version.

isnotinvain:
got a +1 from @johnynek so I'll merge this, thanks!

Issue #393, opened 2015-01-07
CMS: Do we support negative counts and/or is computing inner products broken?


miguno:
## Background

I came across this question while debugging a failing test (computing the inner product) for CMS with the proposed new hashing code for `BigInt` discussed at #392.  The failing test in question is ["estimate inner products"](https://github.com/twitter/algebird/blob/develop/algebird-test/src/test/scala/com/twitter/algebird/CountMinSketchTest.scala#L198).

I noticed that with the old/legacy CMS hashing for BigInt's -- and because they hash in the same way this means the issue I am talking about is also relevant for `CMS[Long]`, `CMS[Int]`, etc. -- that computing inner products (as part of the test above) returned `Approximate` instances with a negative minimum value:

```scala
case class Approximate[N](min: N, estimate: N, max: N, probWithinBounds: Double)(implicit val numeric: Numeric[N]) { ...}
```

Example using current code and current BigInt hashing:

```
Approximate(-2544,24850,24850,0.9999999999)  // <<< note the negative minimum of `-2544`
```

This also meant that the test ended up computing weird (?) error bounds because "minus a negative value" is effectively addition.  But the test passed, and has done so for I think many months.

```
approxA.min:  -2544
approx:       24850  (approx = approxA.estimate)
exact:         1192
estimError:   23658  (estimationError = approx - exact)
maxError:     27394  (maxError = approx - approxA.min)  // <<< IMHO maxError should rather be smaller than approx, not larger!
```

> Btw, the same negative value for `Approximate.min` happens for `CMS[Long]`, for instance.

When testing the proposed Murmur3-based hashing for `CMS[BigInt]`, the same tests fails, because here Approximate.min happened to be a positive value.

> Note: I am not claiming (yet?) that the values are negative all the time etc. -- I can only confirm that they happened to be negative in my limited testing.  Same thing for the "positive" values for the minimums returned with the new hashing code.

```
Approximate(6270,33664,33664,0.9999999999) // <<< min is positive, `6270`

approxA.min: 6270
approx:       33664  (approxA.estimate)
exact:         1305
estimError:   32359  (approx - exact)
maxError:     27394  (approx - approxA.min) // <<< this makes more sense, but now the test fails!
```

At this point I got confused.

I checked the [inner product definition](https://github.com/twitter/algebird/blob/develop/algebird-core/src/main/scala/com/twitter/algebird/CountMinSketch.scala#L484) of CMSInstance[K]:

```scala
[...]

Issue #392, opened 2015-01-07
New CMS: Improve hashing of BigInt


miguno:
## Summary

The latest implementation of CMS ships with a default hashing implementation for `BigInt` that is based on the previous hashing approach for `Long`.  Unfortunately, this implementation for `BigInt` is "broken" in my opinion.

I think I have found a clean fix, which I will send in a separate pull request.  It's actually only a few lines of code, and the reason I am writing this ticket is simply to justify these lines and solicit peer review. :-)

## What is broken in CMS[K] for BigInt's?

**TL;DR: The current hashing causes more collisions "than expected", leading to counter-intuitive CMS counting behavior because similar but different keys end up being treated the same.  For instance, "twitter.com" is hashed to the same counting table slot as "www.twitter.com" and "mail.twitter.com".**

Currently we use the following hashing function for `BigInt`:

```scala
  implicit object CMSHasherBigInt extends CMSHasher[BigInt] {

    def hash(a: Int, b: Int, width: Int)(x: BigInt) = {
      val unModded: BigInt = (x * a) + b
      val modded: BigInt = (unModded + (unModded >> 32)) & PRIME_MODULUS
      val h = modded.toInt % width
      assert(h >= 0, "hash must not be negative")
      h
    }

  }
```

I will now show a few examples that illustrate what is wrong with the current hashing implementation for `BigInt`.

```
$ ./sbt "project algebird-core" console
```
```scala
scala> import com.twitter.algebird._
scala> import com.twitter.algebird.CMSHasherImplicits._
scala> val a = 5; val b = 0; val width = 999;

// Example: We hash domain names.
// Result: Different domain names are assigned to the same slot.
Seq("twitter.com", "www.twitter.com", "mail.twitter.com").map { s => CMSHasherBigInt.hash(a, b, width)(BigInt(s.getBytes("UTF8"))) }
res133: Seq[Int] = List(889, 889, 889)   //  <<< again, all inputs hash to the same slot!

// Let's see how bad it is.  Ok, once we have only "ter.com" to hash we get a different slot.
scala> Seq("twitter.com", "witter.com", "itter.com", "tter.com", "ter.com").map { s => CMSHasherBigInt.hash(a, b, width)(BigInt(s.getBytes("UTF8"))) }
res209: Seq[Int] = List(889, 889, 889, 889, 490)

// Also, I don't think that these domains should be considered the same...
[...]

PR #394, opened 2015-01-08
GH-392: Improve hashing of BigInt

miguno:
This PR addresses the issue described in #392.

In addition, this PR includes several CMS code fixes as well as CMS test/spec fixes that I described at #393.  The reason these fixes are included in the PR is that the hashing improvements for BigInt exposed these code bugs / test bugs, and excluding those fixes would have meant that we end up with non-working code.

On a high-level, the hashing improvement for `CMS[BigInt]` covers (cf. #392):

- We now use Murmur3-based hashing for `BigInt` instead of the "default" hashing implementation we use for `Long` and friends.  See #392 for the justification of using Murmur3 and for why this change in general is important for `BigInt` use cases.

On a high-level, the additional fixes cover (cf. #393):

- Fixes to how we estimate inner products of CMS.
- Fixes and improvements (we use ScalaCheck now) to how we test the estimation of inner products.
- Fix to ensure `TopCMSItem` instances (which represent a single item with count 1) are created with a correct counting table.  The previous code was buggy and resulted in miscomputation of inner products.

### DIFF
--- a/algebird-caliper/src/test/scala/com/twitter/algebird/caliper/CMSHashingBenchmark.scala
+++ b/algebird-caliper/src/test/scala/com/twitter/algebird/caliper/CMSHashingBenchmark.scala
@@ -60,9 +60,9 @@ class CMSHashingBenchmark extends SimpleBenchmark {
   private def murmurHashScala(a: Int, b: Int, width: Int)(x: BigInt) = {
     val hash: Int = scala.util.hashing.MurmurHash3.arrayHash(x.toByteArray, a)
     val h = {
-      // We only want positive integers for the subsequent modulo.  This method mimics Java's Hashtable implementation,
-      // and it requires `hash` to be an `Int` = have 32 bits (to match with `0x7FFFFFFF`).
-      val positiveHash = hash & 0x7FFFFFFF
+      // We only want positive integers for the subsequent modulo.  This method mimics Java's Hashtable
+      // implementation.  The Java code uses `0x7FFFFFFF` for the bit-wise AND, which is equal to Int.MaxValue.
+      val positiveHash = hash & Int.MaxValue
       positiveHash % width
     }
     assert(h >= 0, "hash must not be negative")
--- a/algebird-caliper/src/test/scala/com/twitter/algebird/caliper/CMSHashingBenchmark.scala
+++ b/algebird-caliper/src/test/scala/com/twitter/algebird/caliper/CMSHashingBenchmark.scala
@@ -1,0 +1,98 @@
+package com.twitter.algebird.caliper
+
+import com.google.caliper.{Param, SimpleBenchmark}
+
+/**
+ * Benchmarks the hashing algorithms used by Count-Min sketch for CMS[BigInt].
+ *
+ * The input values are generated ahead of time to ensure that each trial uses the same input (and that the RNG is not
+ * influencing the runtime of the trials).
+ */
+// Once we can convince cappi (https://github.com/softprops/capp) -- the sbt plugin we use to run
+// caliper benchmarks -- to work with the latest caliper 1.0-beta-1, we would:
+//     - Let `CMSHashingBenchmark` extend `Benchmark` (instead of `SimpleBenchmark`)
+//     - Annotate `timePlus` with `@MacroBenchmark`.
+class CMSHashingBenchmark extends SimpleBenchmark {
+
+  /**
+   * The `a` parameter for CMS' default ("legacy") hashing algorithm: `h_i(x) = a_i * x + b_i (mod p)`.
+   */
+  @Param(Array("5123456"))
+  val a: Int = 0
+
+  /**
+   * The `b` parameter for CMS' default ("legacy") hashing algorithm: `h_i(x) = a_i * x + b_i (mod p)`.
+   *
+   * Algebird's CMS implementation hard-codes `b` to `0`.
+   */
+  @Param(Array("0"))
+  val b: Int = 0
+
+  /**
+   * Width of the counting table.
+   */
+  @Param(Array("11" /* eps = 0.271 */ , "544" /* eps = 0.005 */ , "2719" /* eps = 1E-3 */ , "271829" /* eps = 1E-5 */))
+  val width: Int = 0
+
+  /**
+   * Number of operations per benchmark repetition.
+   */
+  @Param(Array("100000"))
+  val operations: Int = 0
+
+  /**
+   * Maximum number of bits for randomly generated BigInt instances.
+   */
+  @Param(Array("128", "1024", "2048"))
+  val maxBits: Int = 0
+
+  var random: scala.util.Random = _
+  var inputs: Seq[BigInt] = _
+
+  override def setUp() {
+    random = new scala.util.Random
+    // We draw numbers randomly from a 2^maxBits address space.
+    inputs = (1 to operations).view.map { _ => scala.math.BigInt(maxBits, random)}
+  }
+
+  private def murmurHashScala(a: Int, b: Int, width: Int)(x: BigInt) = {
+    val hash: Int = scala.util.hashing.MurmurHash3.arrayHash(x.toByteArray, a)
+    val h = {
+      // We only want positive integers for the subsequent modulo.  This method mimics Java's Hashtable implementation,
+      // and it requires `hash` to be an `Int` = have 32 bits (to match with `0x7FFFFFFF`).
+      val positiveHash = hash & 0x7FFFFFFF
+      positiveHash % width
+    }
+    assert(h >= 0, "hash must not be negative")
+    h
+  }
+
+  private val PRIME_MODULUS = (1L << 31) - 1
+
+  private def brokenCurrentHash(a: Int, b: Int, width: Int)(x: BigInt) = {
+    val unModded: BigInt = (x * a) + b
+    val modded: BigInt = (unModded + (unModded >> 32)) & PRIME_MODULUS
+    val h = modded.toInt % width
+    assert(h >= 0, "hash must not be negative")
+    h
+  }
+
+  def timeBrokenCurrentHashWithRandomMaxBitsNumbers(operations: Int): Int = {
+    var dummy = 0
+    while (dummy < operations) {
+      inputs.foreach { input => brokenCurrentHash(a, b, width)(input)}
+      dummy += 1
+    }
+    dummy
+  }
+
+  def timeMurmurHashScalaWithRandomMaxBitsNumbers(operations: Int): Int = {
+    var dummy = 0
+    while (dummy < operations) {
+      inputs.foreach { input => murmurHashScala(a, b, width)(input)}
+      dummy += 1
+    }
+    dummy
+  }
+
+}
--- a/algebird-caliper/src/test/scala/com/twitter/algebird/caliper/CMSHashingBenchmark.scala
+++ b/algebird-caliper/src/test/scala/com/twitter/algebird/caliper/CMSHashingBenchmark.scala
@@ -7,6 +7,8 @@ import com.google.caliper.{Param, SimpleBenchmark}
  *
  * The input values are generated ahead of time to ensure that each trial uses the same input (and that the RNG is not
  * influencing the runtime of the trials).
+ *
+ * More details available at https://github.com/twitter/algebird/issues/392.
  */
 // Once we can convince cappi (https://github.com/softprops/capp) -- the sbt plugin we use to run
 // caliper benchmarks -- to work with the latest caliper 1.0-beta-1, we would:
--- a/algebird-core/src/main/scala/com/twitter/algebird/CountMinSketch.scala
+++ b/algebird-core/src/main/scala/com/twitter/algebird/CountMinSketch.scala
@@ -479,7 +479,7 @@ case class CMSInstance[K: Ordering](countsTable: CMSInstance.CountsTable[K],
    * Let X be a CMS, and let count_X[j, k] denote the value in X's 2-dimensional count table at row j and column k.
    * Then the Count-Min sketch estimate of the inner product between A and B is the minimum inner product between their
    * rows:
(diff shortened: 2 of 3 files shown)


algebird-caliper/src/test/scala/com/twitter/algebird/caliper/CMSHashingBenchmark.scala
```diff
@@ -0,0 +1,100 @@
...
+      // We only want positive integers for the subsequent modulo.  This method mimics Java's Hashtable implementation,
+      // and it requires `hash` to be an `Int` = have 32 bits (to match with `0x7FFFFFFF`).
+      val positiveHash = hash & 0x7FFFFFFF
+      positiveHash % width
+    }
+    assert(h >= 0, "hash must not be negative")
+    h
+  }
+
+  private val PRIME_MODULUS = (1L << 31) - 1
```
johnynek: Int.MaxValue is another way to say this, right?

algebird-caliper/src/test/scala/com/twitter/algebird/caliper/CMSHashingBenchmark.scala
```diff
@@ -0,0 +1,100 @@
...
+    // We draw numbers randomly from a 2^maxBits address space.
+    inputs = (1 to operations).view.map { _ => scala.math.BigInt(maxBits, random)}
+  }
+
+  private def murmurHashScala(a: Int, b: Int, width: Int)(x: BigInt) = {
+    val hash: Int = scala.util.hashing.MurmurHash3.arrayHash(x.toByteArray, a)
+    val h = {
+      // We only want positive integers for the subsequent modulo.  This method mimics Java's Hashtable implementation,
+      // and it requires `hash` to be an `Int` = have 32 bits (to match with `0x7FFFFFFF`).
+      val positiveHash = hash & 0x7FFFFFFF
```
johnynek: the number on the right is Int.MaxValue, right? Any thoughts on why to say it one way or another?

algebird-caliper/src/test/scala/com/twitter/algebird/caliper/CMSHashingBenchmark.scala
```diff
@@ -0,0 +1,100 @@
...
+      positiveHash % width
+    }
+    assert(h >= 0, "hash must not be negative")
```
johnynek: this really can't ever throw due to how it is constructed. Is it worth a check that never succeeds? My thinking is that this is better in a test than in the run path.

algebird-caliper/src/test/scala/com/twitter/algebird/caliper/CMSHashingBenchmark.scala
```diff
@@ -0,0 +1,100 @@
...
+  var inputs: Seq[BigInt] = _
+
+  override def setUp() {
+    random = new scala.util.Random
+    // We draw numbers randomly from a 2^maxBits address space.
+    inputs = (1 to operations).view.map { _ => scala.math.BigInt(maxBits, random)}
+  }
+
+  private def murmurHashScala(a: Int, b: Int, width: Int)(x: BigInt) = {
+    val hash: Int = scala.util.hashing.MurmurHash3.arrayHash(x.toByteArray, a)
```
johnynek: not using b worries me. Can you add a comment as to why this is safe? Why would we not combine a and b in some interesting way (a * Int.MaxValue + b) or something.

algebird-core/src/main/scala/com/twitter/algebird/CountMinSketch.scala
```diff
@@ -1079,14 +1087,23 @@ object CMSHasherImplicits {
...
+     */
+    override def hash(a: Int, b: Int, width: Int)(x: BigInt) = {
+      val hash: Int = scala.util.hashing.MurmurHash3.arrayHash(x.toByteArray, a)
+      val h = {
+        // We only want positive integers for the subsequent modulo.  This method mimics Java's Hashtable implementation,
+        // and it requires `hash` to be an `Int` = have 32 bits (to match with `0x7FFFFFFF`).
+        val positiveHash = hash & 0x7FFFFFFF
+        positiveHash % width
+      }
       assert(h >= 0, "hash must not be negative")
```
johnynek: see comments above about this.

algebird-core/src/main/scala/com/twitter/algebird/CountMinSketch.scala
```diff
@@ -1079,14 +1087,23 @@ object CMSHasherImplicits {
...
-      val unModded: BigInt = (x * a) + b
-      val modded: BigInt = (unModded + (unModded >> 32)) & PRIME_MODULUS
-      val h = modded.toInt % width
+    /**
+     * Implementation detail:  This hash function is based upon Murmur3.  Note that the original CMS paper requires
+     * `d` (depth) pair-wise independent hash functions;  in the specific case of Murmur3 we argue that it is sufficient
+     * to pass `d` different seed values to Murmur3 to achieve a similar effect.
+     */
+    override def hash(a: Int, b: Int, width: Int)(x: BigInt) = {
+      val hash: Int = scala.util.hashing.MurmurHash3.arrayHash(x.toByteArray, a)
```
johnynek: see comments about my anxiety lack of using b.

algebird-test/src/test/scala/com/twitter/algebird/CountMinSketchTest.scala
```diff
@@ -196,23 +214,26 @@ abstract class CMSTest[K: Ordering: CMSHasher: Numeric] extends WordSpec with Ma
...
+      val ranges = Gen.choose(100, 2000)
+
+      forAll((totalCounts, "totalCount"), (ranges, "range"), minSuccessful(50)) { (totalCount: Int, range: Int) =>
+        val data1 = createRandomStream(totalCount, range)
+        val data2 = createRandomStream(totalCount, range)
+        val cms1 = COUNTING_CMS_MONOID.create(data1)
+        val cms2 = COUNTING_CMS_MONOID.create(data2)
+
+        val approxA = cms1.innerProduct(cms2)
+        val approx = approxA.estimate
```
johnynek: can we also check that `approxA.min >= 0`?

algebird-core/src/main/scala/com/twitter/algebird/CountMinSketch.scala
```diff
@@ -964,7 +972,7 @@ class TopNCMSMonoid[K: Ordering](cms: CMS[K], heavyHittersN: Int = 100) extends
   /**
    * Creates a sketch out of a single item.
    */
-  def create(item: K): TopCMS[K] = TopCMSItem[K](item, cms, params)
+  def create(item: K): TopCMS[K] = TopCMSItem[K](item, cms + item, params)
```
johnynek: I guess this was a bug before. Does the new test find this bug? Why didn't we see a test failure in the past?

algebird-caliper/src/test/scala/com/twitter/algebird/caliper/CMSHashingBenchmark.scala
miguno: My reasoning was alomg those lines: http://stackoverflow.com/questions/3956478/understanding-randomness.  As long as the seed of Murmur is random we should get what we need.  My understanding is that "randomVal2 <op> randomVal2" will not improve the "randomness" (uniform distribution) as long as they are generated by the same source, which is the case for us because we use Scala's PRNG only.  According to the discussion above, doing any such combinazion of a and b might actually worsen the outcome.

Also, the current CMS implementation hardcodes b to 0 anyway.  See top of the file where the random hashes are generated.  Given the code comments this was apparently safe for the "ax + b mod p" approach we are still using for Long, Int, Short.

Does that make sense? Please correct me if I'm wrong.

miguno:
I will respond to the rest of your comments tomorrow, Oscar!

algebird-core/src/main/scala/com/twitter/algebird/CountMinSketch.scala
miguno: We do not have a direct test for this (may or may not be helpful, I am fine either way), but the new property-based tests for estimating inner product has been triggering the bug pretty reliably before ai fixed it.  The reason we didn't see this bug before was that the previous, non-fuzzy test for estimating inner products happened to use a test fixture (e.g. totalCount setting) so that the test passed kinda by pure luck.

Maybe we should actually add a direct test for this anyways. What do you think?

algebird-test/src/test/scala/com/twitter/algebird/CountMinSketchTest.scala
miguno: Ah, good catch. Will do.

algebird-caliper/src/test/scala/com/twitter/algebird/caliper/CMSHashingBenchmark.scala
miguno: PS: And in case my reasoning does make sense I will add a ccode comment of course.

algebird-test/src/test/scala/com/twitter/algebird/CountMinSketchTest.scala
miguno: Done.

algebird-core/src/main/scala/com/twitter/algebird/CountMinSketch.scala
miguno: I added two tests for each Top-% CMS and Top-N CMS that test for this bug.  Note that because the counting table has private access I use "identity" inner products (= the inner product of a CMS instance with itself) to verify correct behavior instead of e.g. verifying that the L1-norm of each `w` vector is 1.

algebird-caliper/src/test/scala/com/twitter/algebird/caliper/CMSHashingBenchmark.scala
miguno: Yes.  I changed the line from `(1L << 31) - 1` to `Int.MaxValue1`.  (Note that I intentionally did not change the legacy code before because that change would have been unrelated to the BigInt-hashing change.)

algebird-caliper/src/test/scala/com/twitter/algebird/caliper/CMSHashingBenchmark.scala
miguno: No special thoughts here -- I simply used the same code as we see in Java's Hashtable implementation.  According to your suggestion I changed `0x7FFFFFFF` to `Int.MaxValue` because it clarifies the intent better than the hex numeral.

I compared benchmark runs before/after this change and I didn't observe any noticable performance difference.  Hence I'd prefer the more readable variant (`Int.MaxValue`), too.

algebird-caliper/src/test/scala/com/twitter/algebird/caliper/CMSHashingBenchmark.scala
miguno: I replaced the `assert` for all hash functions with a fuzzy test that verifies that hashes (i.e. slots) are always positive.

miguno:
All your comments should be addressed, Oscar.  And thanks for the review, of course!

miguno:
Note: One of the Travis build jobs failed but for an unrelated test (BloomFilter):

```
[info] BloomFilter
[info] - should identify all true positives
[info] - should have small false positive rate *** FAILED ***
[info] 0.0021 was not less than or equal to 0.002 (BloomFilterTest.scala:141)
```

algebird-test/src/test/scala/com/twitter/algebird/CountMinSketchTest.scala
```diff
@@ -272,6 +283,46 @@ abstract class CMSTest[K: Ordering: CMSHasher: Numeric] extends WordSpec with Ma
...
+        // Poor man's way to come up with an item that is not x and that is very unlikely to hash to the same slot.
+        val otherItem = x + 1
+        topCms.frequency(otherItem.toK[K]).estimate should be(0)
+        // The following assert indirectly verifies whether the counting table is not all-zero (cf. GH-393).
+        topCms.innerProduct(topCms).estimate should be(1)
+      }
+    }
+
+    "create correct sketches out of a single-item stream" in {
+      forAll{ (x: Int, y: Int) =>
```
johnynek: this is weird. Isnt' this x shadowed below?

algebird-test/src/test/scala/com/twitter/algebird/CountMinSketchTest.scala
```diff
@@ -405,6 +456,46 @@ abstract class CMSTest[K: Ordering: CMSHasher: Numeric] extends WordSpec with Ma
...
```
johnynek: shadowing again?

algebird-test/src/test/scala/com/twitter/algebird/CountMinSketchTest.scala
miguno: You were right, this shadowing should not have been there.  The outer x/y loop was a relic from the previous test code.  Dumb copy-paste mistake of mine!

johnynek:
Thank you sir!

miguno:
Thanks for the review and the merge!

ianoc:
We probably need to make sure to call this out big in the release since any serialized CMS's using Big Int right now won't remain valid after the new version right? I don't think we internally use the ability at all so it won't effect us

johnynek:
Good point, Ian. They will want to override the hasher to be the same as the old version.

That said, it looks like the old hashing was so broken they will be getting terrible results with their data.

miguno:
Yes, good point, Ian.

Looking at the git tag of the latest released version (0.8.2) tells me that the CMS[K] pull request -- which introduced BigInt support -- was not included in 0.8.2.  So as long as there was no custom build in use internally on your side then there shouldn't have been anyone using the BigInt CMS.  (Note: We did use a custom build because CMS[K] was not in 0.8.2.)

And as Oscar said if their use case was similar to ours then they would surely like to reprocess the data anyways...

PR #397, opened 2015-01-10
Algebird support for spark

johnynek:
This needs tests to make sure the implicits are wired up correctly, but it seems like it is time to make Algebird-spark to make it easy to use Algebird with spark.


johnynek:
@ianoc and @tsdeng care to look at this?

algebird-spark/src/main/scala/com/twitter/algebird/spark/package.scala
```diff
@@ -0,0 +1,46 @@
...
+ */
+package object spark {
+  /**
+   * This adds methods to Spark RDDs to use Algebird
+   */
+  implicit class AlgebirdRDD[T](val rdd: RDD[T]) extends AnyVal {
+    /**
+     * Apply an Aggregator to return a single value for the whole RDD
+     */
+    def aggregatorOnAll[B: ClassTag, C](agg: Aggregator[T, B, C]): C =
```
avibryant: why `aggregatorOnAll` and `aggregatorByKey` rather than `aggregateAll` and `aggregateByKey`? The latter seem a closer fit to spark's `reduceByKey`

johnynek: There is already `aggregateByKey`. I will add tests and see if the implicit enrichment can differentiate based on the arguments. It may, or it may give the user a very confusing message. Let's see.

ianoc:
any tests? looks fine modulo Avi's comment about naming would be nice to be closer to sparks.

algebird-spark/src/main/scala/com/twitter/algebird/spark/package.scala
```diff
@@ -0,0 +1,46 @@
...
+      agg.present(rdd.map(agg.prepare).reduce(agg.reduce))
+
+    def aggregatorByKey[K: ClassTag, A: ClassTag, B: ClassTag, C](agg: Aggregator[A, B, C])(implicit ev: T <:< (K, A), ordK: Ordering[K] = null): RDD[(K, C)] =
```
miguno: - Where/how is `ev` / `T` used?  Is it needed by Spark or Algebird behind the scenes?

- Any particular reason to give a default of `null` to `ordK` and/or to use an implicit parameter instead of `K: ClassTag: Ordering`?

    ```scala
    implicit ordK: Ordering[K] = null
    ```

johnynek: seems like spark follows this style, which I assume that it means it can try to work even if there is no Ordering. Scalding requires the ordering, and uses the approach you suggest.

adelbertc:
Interested in this effort, especially with the [algebra](https://github.com/non/algebra) stuff going on. I've talked with a few people briefly about a more open-source effort to share common Spark things that are related to Algebird, e.g. having functions that takes a `Monoid` instead of `A => (A => A => A)`, etc.

ianoc:
@johnynek This looks good to me, if we can fix the merge conflict looks good to merge

johnynek:
@ianoc we need tests too, right?

virtualirfan:
An example would be nice too.

ianoc:
Tests would be great but as some tiny implicits id happily ship if manually verified and follow up with those later.

adelbertc:
Latest version of Spark is now 1.3.0

reconditesea:
+1 for a small tutorial job of how to make spark work with algebird.

johnynek:
@adelbertc can you review?

@ianoc this has tests now, and I think reasonably efficient implementations, especially for sum, sumOption, aggregate, aggregateOption.

I don't see a way to use sumOption in sumByKey or aggregateByKey with reimplementing or skipping map-side combining. I wish spark had something like scalding's sumByLocalKeys.

project/Build.scala
```diff
@@ -160,6 +161,10 @@ object AlgebirdBuild extends Build {
   lazy val algebirdBijection = module("bijection").settings(
     libraryDependencies += "com.twitter" %% "bijection-core" % "0.8.0"
   ).dependsOn(algebirdCore, algebirdTest % "test->test")
+
+  lazy val algebirdSpark = module("spark").settings(
+    libraryDependencies += "org.apache.spark" %% "spark-core" % "1.3.0" % "provided"
```
adelbertc: Latest Spark version is 1.4.0 now, not sure how you want to handle versioning

algebird-spark/src/main/scala/com/twitter/algebird/spark/AlgebirdRDD.scala
```diff
@@ -0,0 +1,73 @@
...
+  def sum(implicit mon: Monoid[T], ct: ClassTag[T]): T =
+    sumOption.getOrElse(mon.zero)
+
+  def sumOption(implicit sg: Semigroup[T], ct: ClassTag[T]): Option[T] = {
+    val partialReduce: RDD[T] = rdd.mapPartitions({ itT => sg.sumOption(itT).toIterator },
+      preservesPartitioning = true)
+
+    // my reading of the docs is that we do want a shuffle at this stage to
+    // to make sure the upstream work is done in parallel.
+    val results = partialReduce.coalesce(1, shuffle = true).mapPartitions({ it =>
```
adelbertc: I'm not too familiar with how the underlying operations work, but you may or may not need a commutative Semigroup for this to work out... it's very possible I'm wrong though.

algebird-spark/src/main/scala/com/twitter/algebird/spark/AlgebirdRDD.scala
```diff
@@ -0,0 +1,73 @@
...
+    case (None, m: Monoid[B]) => agg.present(m.zero)
+    case (None, _) => None.get // no such element
+  }
+
+  def aggregateByKey[K: ClassTag, V1, U: ClassTag, V2](agg: Aggregator[V1, U, V2])(implicit ev: T <:< (K, V1), ordK: Ordering[K] = null): RDD[(K, V2)] = {
+    val prepared = new PairRDDFunctions(rdd.map { t =>
+      /*
+       * We use the cast to avoid having to serialize the ev, which could also be applied
+       */
+      val tupl = t.asInstanceOf[(K, V1)]
```
adelbertc: Use `keyed` to be consistent? ... minor :-)

adelbertc:
:+1: LGTM, only thing I'm concerned about is whether or not commutativity is needed in the `sumOption` case

algebird-spark/src/main/scala/com/twitter/algebird/spark/AlgebirdRDD.scala
```diff
@@ -0,0 +1,73 @@
+package com.twitter.algebird
+
+import com.twitter.algebird._
+import org.apache.spark.rdd.{ RDD, PairRDDFunctions }
+import scala.reflect.ClassTag
+/**
+ * import com.twitter.algebird.spark.toAlgebird
```
jnievelt: ToAlgebird?

project/Build.scala
ianoc: Its provided so the user brings their own. Once the API's ABI compatible in spark this uses it should work fine for both

algebird-spark/src/main/scala/com/twitter/algebird/spark/AlgebirdRDD.scala
ianoc: it does need a commutative semigroup for both of these operations(in sumOption) to work. In practice thats often what users have/operations they want to do (take scalding's sumByKey). 

Not sure if it should be indicated clearer here

algebird-core/src/main/scala/com/twitter/algebird/SummingCache.scala
```diff
@@ -58,7 +58,7 @@ class SummingCache[K, V](capacity: Int)(implicit sgv: Semigroup[V])
   }
   override def flush: Option[Map[K, V]] = {
     // Get a copy of the cache, since it is mutable
-    val res = optNonEmpty(Map(cache.toSeq: _*))
+    val res = optNonEmpty(cache.iterator.toMap)
```
ianoc: was there a particular reason to make this change? It looks like its probably more efficient, but curious if you had a benchmark/profile you saw this in?

algebird-spark/src/main/scala/com/twitter/algebird/spark/AlgebirdRDD.scala
ianoc: Could probably unwrap this in place here if the compiler doesn't mind to avoid the _1 and _2 references below

algebird-spark/src/main/scala/com/twitter/algebird/spark/AlgebirdRDD.scala
```diff
@@ -0,0 +1,73 @@
...
+  /**
+   * This will throw if you use a non-MonoidAggregator with an empty RDD
+   */
+  def aggregate[B: ClassTag, C](agg: Aggregator[T, B, C]): C = (aggregateOption[B, C](agg), agg.semigroup) match {
+    case (Some(c), _) => c
+    case (None, m: Monoid[B]) => agg.present(m.zero)
+    case (None, _) => None.get // no such element
+  }
+
+  def aggregateByKey[K: ClassTag, V1, U: ClassTag, V2](agg: Aggregator[V1, U, V2])(implicit ev: T <:< (K, V1), ordK: Ordering[K] = null): RDD[(K, V2)] = {
```
ianoc: minor, but line is too long here for github viewing with scrolling, probably good to wrap it

algebird-spark/src/main/scala/com/twitter/algebird/spark/AlgebirdRDD.scala
```diff
@@ -0,0 +1,73 @@
...
+      .map { case (k, b) => (k, agg.present(b)) }
+  }
+
+  private def keyed[K, V](implicit ev: T <:< (K, V)): RDD[(K, V)] = rdd.asInstanceOf[RDD[(K, V)]]
+
+  /**
+   * Use the implicit semigroup to sum by keys
+   */
+  def sumByKey[K: ClassTag, V: ClassTag: Semigroup](implicit ev: T <:< (K, V), ord: Ordering[K] = null): RDD[(K, V)] =
+    (new PairRDDFunctions(keyed)).reduceByKey(implicitly[Semigroup[V]].plus)
```
ianoc: Shame no way to wire in sumOption here...but oh well.

johnynek:
I think this addresses all review concerns. I also added Partitioner support for the cases where you want to control partitioning.

algebird-core/src/main/scala/com/twitter/algebird/SummingCache.scala
johnynek: I didn't benchmark. I was thinking about using this to implement mapSide caching, which could have allowed at least using sumOption on the "reducer" side of the computation, but I back-peddled on that since we'd have to have parameters for the cachesize, etc...

When I looked at this, I noticed the existing algorithm is going through the cache twice: once to create the Seq. a second time to create the map. Using the iterator, you go through it only one time. Seems hard to imagine how this can hurt, but it probably does not help much as flush does not occur very often.

ianoc:
lgtm, merge when green

ianoc:
@johnynek incase you haven't seen, failing with
[info] Compiling 1 Scala source to /home/travis/build/twitter/algebird/algebird-bijection/target/scala-2.11/classes...
[error] /home/travis/build/twitter/algebird/algebird-spark/src/main/scala/com/twitter/algebird/spark/AlgebirdRDD.scala:14: in class AlgebirdRDD, multiple overloaded alternatives of method sumByKey define default arguments.
[error] class AlgebirdRDD[T](val rdd: RDD[T]) extends AnyVal {
[error]       ^
[error] one error found

johnynek:
@ianoc yeah. I fixed it by adding a class that @non , @avibryant and I (I think it was) discussed and @non put in algebra.

still need to get algebird to depend on algebra.

ianoc:
Cool, lgtm anyway with this in place

johnynek:
coveralls is keeping this yellow. Need to get this sorted out.

Issue #398, opened 2015-01-14
We should have a default hasher for Strings for CMS


johnynek:
Here are two approaches:

1) hash using Murmur to Long, then use the Long universal hash functions.

2) append the a and b we get for the family of hashes and use Murmur on the whole block.

We could also add `Array[Byte]` while we are at it since we will probably go through that anyway.

miguno:
That's a good suggestion.  I have been thinking about the very same as we have been using `CMS[BigInt]` to effectively do `CMS[String]` (you can easily convert BigInt<->String with built-in methods). :-)

> 1) hash using Murmur to Long, then use the Long universal hash functions.

Hmm.  Why would you want to hash twice, i.e. `hashLong(hashMurmur(s))` to get the desired `Int`?

Murmur already gives us the desired `Int`, hence doing only `hashMurmur(s)` is faster.  Furthermore, I *think* the Murmur-only approach would provide us with a more uniform distribution of `Int`s then a double-hash approach;  at least in the Murmur-only case we *know* based on prior work (I don't mean #392 necessarily, I mean studies of Murmur3 in general) that the `Int` output is well-formed, whereas we'd have to do our own research/plotting/etc. to understand the impact of the proposed double-hashing.  For instance, `(ax + b) mod p` is said to sort-of [produce predictable patterns](http://stackoverflow.com/a/12120625/1743580) -- to me this sounds like it would lower the quality of the Murmur hashing, and we'd have to do the latter anyways.

Similarly, my experience with #392 makes me feel that we should rely on just Murmur3, as the `(ax + b) mod p` hashing resulted in counter-intuitive (read: wrong, IMHO) behavior when hashing `BigInt`s.  And as I said above, we have been using `CMS[BigInt]` to effectively do `CMS[String]`, so I can claim limited practical experience with it. :-)

> 2) append the a and b we get for the family of hashes and use Murmur on the whole block.

Not sure I understand -- why would you want to append `a` and/or `b` to the string before hashing?

> We could also add Array[Byte] while we are at it since we will probably go through that anyway.

The `BigInt` hashing is effectively hashing `Array[Byte]` already, so that would be a starting point.  

Here's a snippet for hashing `BigInt`, `String`, `Array[Byte]`:

```scala
  implicit object CMSHasherArrayByte extends CMSHasher[Array[Byte]] {
    override def hash(a: Int, b: Int, width: Int)(x: Array[Byte]): Int = {
[...]

Issue #353, opened 2014-10-15
CMS computes heavy hitters incorrectly when combining CMS instances with ++


miguno:
_This is a follow-up on a conversation I had with @johnynek and @avibryant on Twitter.  Many thanks to both of you for your help on narrowing this down.  And please do let me know if I'm mistaken with what I'm writing here._

From what I understand today, I think I was partly right and partly wrong in my Twitter comments from yesterday:

* I was right right in the sense there currently _is_ a bug in the Algebird's CMS that miscomputes heavy hitters under certain (data) conditions.  I describe these conditions in detail below and also propose a fix, and this fix is exactly what @avibryant and I briefly talked about on Twitter.
* I was (partly) wrong though when I said this is related to the merging of heavy hitters not being associative -- here I over-generalized.  In fact, the merging of heavy hitters (once fixed) _is_ associative for top-% CMS variants (the current, pre GH-345 implementation of CMS is such a variant) but it is not associative for top-N CMS variants.  On the positive side, this means once this fix is included, the current CMS variant works correctly.  On the negative side, it means that a top-N CMS variant may still (and always?) have a problem when merging heavy hitters, and SketchMap might be similarly affected (and if SketchMap is not, which would be great, then we may have just learned we can't simply replace SketchMap with a top-N CMS variant).

In summary:

* We seem to have a legitimate bug in the current (top-% based) CMS implementation that miscomputes heavy hitters under certain conditions.  There's a fix for that, see below.
* We seem to have a general limitation to correctly compute heavy hitters for any top-N based CMS.

## Background

As part of my refactoring work in GH-345 I decoupled the counting/querying functionality of CMS from the heavy hitters functionality (as per @jnievelt's suggestion), and once I reached that point I could easily provide both top-% and top-N CMS variants (here, having a top-N CMS variant brings us one step closer to eventually retire SketchMap).  And while testing the top-N variant I first discovered the heavy hitters issue I describe here.

## Problem statement

_In the discussion below I assume the two CMS instances we want to merge are called `left` and `right`, respectively._

The computation to combine two sets of heavy hitters is not correct right now when you use `CMSInstance#++(other: CMS): CMS` for combing two CMS instances.  The problem itself is two-fold, though both aspects are related and can be fixed in one swing.

[...]

PR #399, opened 2015-01-15
CMS: add contramap to convert CMS[K] to CMS[L], add support for String and Bytes, remove Ordering context bound for K

miguno:
Here's a first draft of `contramap` for the CMS monoid as per our discussion in #398.

Three things stand out:

- It seems to work. :-)  See `CMSStringTest`.
- We need another implicit value in scope, which is an instance of `Ordering[Array[Byte]]`.  I put this into a separate `CMSOrderingImplicits` object.
- I added the default hashing functions for `Array[Byte]`, `BigInt`, and `String` as per my snippet above.  However, I disabled the `String` hashing function because I lacked the creativity to come up with a non-BigInt, non-String example to showcase how to translate `CMS[K]` into `CMS[L]` (hence I did `CMS[Array[Byte]]` into `CMS[String]`).

Let me know what you think.  (I do not consider this code to be production-ready yet.)


miguno:
My biggest open question is how to test the contramap functionality in an elegant and effective way.

johnynek:
So, looking at this indeed your concern about ordering on `Array[Byte]` is legit. For UTF-8 strings, they are ordered on unsigned bytes.

That said, I'm not sure why we need an Ordering on Keys at all. I was looking back through the code, and it seems to me we are only using the Ordering when we do heavy-hitters, and even then it is just to break ties if there are items with the same count.  Seems to me we could have used something like the hash code to do that since it is hard to imagine that ordering is truly meaningful given it will only matter if one is evicted from the heavy hitters and the other is not, but that only happens for the TopK version, which is not associative anyway.

I wonder if we are making a mistake by requiring `Ordering[K]` at  all.

miguno:
You raise a good point.  I will reply in further detail tomorrow.  For starters though I have revisited the core CMS code (counting-only) and successfully removed the `Ordering[K]` requirement.  That is, we only require `K: CMSHasher` for `CMS` (I haven't fiddled with the Top-CMS variants yet).

miguno:
> I was looking back through the code, and it seems to me we are only using the Ordering when we do heavy-hitters, and even then it is just to break ties if there are items with the same count. Seems to me we could have used something like the hash code to do that since it is hard to imagine that ordering is truly meaningful given it will only matter if one is evicted from the heavy hitters and the other is not, but that only happens for the TopK version, which is not associative anyway.

To add to what you said, technically the requirement on `Ordering` is caused only by CMS' usage of `SortedSet` for heavy hitters:

```scala
case class HeavyHitters[K: Ordering](hhs: SortedSet[HeavyHitter[K]]) extends java.io.Serializable { ... }
```

However, the API contract of CMS only guarantees `Set` for heavy hitters to the user.  There has been a related discussion about SortedSet vs. Set for SketchMap in #277, and I also had a question about the very same thing for CMS some time back at https://github.com/twitter/algebird/issues/353#issuecomment-59190169.

> Seems to me we could have used something like the hash code to do that since it is hard to imagine that ordering is truly meaningful given it will only matter if one is evicted from the heavy hitters and the other is not [...]

Hmm. Given that we're allowing the user to throw any type `K` at us, `Ordering` allows the user to control how such tie-breaking should be performed.  One use case I can think of is similar to a priority queue (the analogy works if you consider heavy hitters to be a ranked/sorted list):  Let's say you have paying customers and non-paying customers, represented via a type `C`, and you use CMS to track misbehavior such as bandwidth usage in order to enforce throttling (users in heavy hitters will be throttled).  Here, if the CMS must break ties between a paying customer and a non-paying customer F, then the client can ensure via `Ordering[C]` that the non-paying customer will be moved/kept in heavy hitters and thus throttled so that the service will not degrade for the paying customer.

I admit that this use case might be a bit contrived.  Why not tracking paying and non-paying customers in separate CMS?  That being said, it is quite hard to anticipate or predict how clients will use an API.

But:

> [...] but that only happens for the TopK version, which is not associative anyway.

How important such a contrived use case is in practice is debatable if it's applicable to the TopK version only anyways (which I agree is much less attractive than its associative Top-% sibling).

johnynek:
Could we easily move the demand for Ordering closer to where it is needed? For instance, only in the heavy hitter variants?

miguno:
I already did that in fact -- see the last three commits in this PR, specifically https://github.com/miguno/algebird/commit/e6158973cd043d96d3f09c8c1e30d15aca42d358. 

So the core CMS is free of Ordering now.  Still, maybe we want to think a bit more about whether we truly need Ordering for the heavy hitters variants.  What's your opinion?  Personally I am still debating pro and con.

miguno:
See also my comment about the Ordering removal at https://github.com/twitter/algebird/pull/399#issuecomment-70156234.

miguno:
Again, the build failed only because of a single unstable test that is not related to the CMS code change.

miguno:
Based @johnynek's comments and concerns about "Do we really need `Ordering[K]`?" I'd propose the following cleanup to CMS.  I think this would adequately address these concerns while improving the code base at the same time.

Note: This cleanup would not be required but very helpful to achieve the original goal of this pull request, which is a) providing a `contramap` functionality for `CMS[K]` to `CMS[L]` auto-translation, which leads up to b) supporting String-based CMS out of the box.  Both (a) and (b) are discussed in #398.  The goal of #398 is to improve the UX of Algebird for users interested in String-based use cases.

#### Proposal

1. Remove the `Ordering` context bound for `K`, and
2. To keep the code simple, remove the Top-N CMS variant.

#### Why we should accept this proposal

- Ordering is not really required given the API contract of CMS.  Getting rid of Ordering simplifies the API (good for devs) as well as using the API (good for users).
    - We have always guaranteed "only" a `Set[K]` for heavy hitters, even though we used `SortedSet[K]` behind the scenes.  As Oscar mentioned we'd need `Ordering` only to break ties when comparing/sorted iff two items have the same count AND we're running a Top-N CMS (= we don't care about this type of tie breaking for the Top-% CMS).
- The Top-N CMS variant is not associative, and one could argue that "Top-N CMS is considered harmful".  Also, it this variant is the only real reason we can't get rid of `Ordering` in a clean way.

#### Why we should NOT accept this proposal

- It is a backwards incompatible change for anyone who is already using the Top-N CMS variant in the wild.  Note that this would have to be a user running a non-official Algebird verison because the latest public version (0.8.2) does not include the new `CMS[K]` code yet (i.e. 0.8.2 has neither `TopPctCMS` nor `TopNCMS`).
- If Top-N use cases are important enough to be supported out of the box, then of course we should keep the Top-N CMS in the code base.
    - In our case, for instance, we are not using the Top-N variant at all because, well, it is not associative, which means we can never trust the outcome of a merge operations of two or more CMS instances.

#### Current status

I have a version of CMS and TopPctCMS that works without `Ordering`.  Behind the scenes we stop using `SortedSet` (as well as `dropWhile` during the heavy hitters purge step) and replace it with `Set` (and a straight `filter`).

Depending on your feedback "Top-N CMS yes/no" I would take another stab at the Top-N variant but here things will get messy because we _must_ break ties in some way, and this will mean either a) to continue to require `Ordering` (at least for Top-N) or b) to come up with a way to break ties without ordering, although that would require some thinking (e.g. if we used a "pick an item at random" strategy -- which is a reasonable strategy IMHO -- to break ties then we'd have a hard time to write deterministic tests;  Oscar suggested hashing to break up ties, which would be better than random tie-breaking but still we'd end up having to kinda "special-craft" test data for the code).

What do you think?

jnievelt:
I'd prefer not to cut TopK, if for no other reason than to maintain parity with the SketchMap implementation.

Instead, my strong preference would be to simply remove the key from the HH ordering and leave it at that.  I would go as far as to say that we don't need to worry about consistently sorting two HH with the same count at all; I can't imagine it happening in practice enough to be meaningful.

avibryant:
@jnievelt why is parity with SketchMap important? Can't we just say "if you want top-k, use SketchMap"?

I'd also be curious to hear how SketchMap gets used in practice. Is it mostly (entirely?) for HLL values? Because if so I think it's not the best way of achieving that (but that's a discussion for another issue/PR).

miguno:
Yeah, I think it would help to understand the reasoning behind keeping feature parity between CMS and SketchMap (not saying I disagree, I simply don't know).

That being said I pushed two new commits.

- The first commit (https://github.com/miguno/algebird/commit/876786085e4bbe7d3794504ae7474cb1af9c57d2) gets rid of the `Ordering` requirement for `K` and turns the `SortedSet` we use behind the scenes into a plain `Set`. Functionality wise we do not need a `SortedSet` to do what we want.  However:
- The second commit (https://github.com/twitter/algebird/commit/6b29c42f5cdb9bfe75508ad2bb51877bca705ee4) brings back `SortedSet` (but not `Ordering`!) to minimize the time complexity of write operations (e.g. `count()`).  It's beneficial to track heavy hitters in a sorted data structure.  In order for this to work, I needed to create a "generic" `Ordering` for heavy hitter elements so that `SortedSet` works.  It is quite important that this ordering works correctly, hence I include it here:

```scala
object HeavyHitter {

  // Note that we can't define the ordering on `count` only.  We use a set to track heavy hitters, where equality means
  // two elements are identical.  If we were to use `count` only, we would label different items in heavy hitters as
  // identical simply because they have the same count (and such false duplicates may happen easily in practice).  For
  // this reason we need to add another dimension to the ordering to prevent such false duplicates.  For this second
  // dimension we use an item's default Java hash code because we do not want to introduce an `Ordering` constraint.
  def ordering[K]: Ordering[HeavyHitter[K]] =
    Ordering.by[HeavyHitter[K], (Long, Int)] { hh => (hh.count, hh.item.hashCode()) }

}
```

jnievelt:
@avibryant I would actually prefer to deprecate `SketchMap` and use `CMS` everywhere.  To answer your usage question, I believe `SketchMap[String, Long]` is very common (including with my team), and these could use `CMS` today.

Of course, there would be significant work required to really make this happen (generalize `CMS[K,V]`, double check performance, etc.).  In the meantime, I like to at least move in that direction.  At least as I see it, we need compelling reason not to do so.

johnynek:
:+1: from me. Any objections or should we put this out in the next version?

miguno:
I'd like to make a few improvements to the current PR (e.g. re-enabling the default hasher for Strings, adding tests, run benchmarks, Joe's ordering variant).

When to you want to create the release?  I will be able to finish this until tomorrow.

johnynek:
@miguno were hoping for today, but maybe not until tomorrow. I'd really like to have this in.

miguno:
Same here.  I’d also like to have this patch included.  As I said I’ll have an updated PR for review by tomorrow your time.  Today won’t work unfortunately. :-)

johnynek:
@miguno missed the train for this one, but if we can make it binary compatible, we can ship it as minor update.

miguno:
No problem.  Since you released already I wasn't in a rush today.  My plan is to submit an updated PR by Friday.

miguno:
Most of the code is done, but I have some weird issues with the tests for `Array[Byte]`.  I'll have to do some further work on that.  Stay tuned!

miguno:
Further progress has been made, including filing a [ScalaTest issue](https://github.com/scalatest/scalatest/issues/491) regarding equality/matcher issues when using e.g. `Set[Array[Byte]]`.  I solved this with a custom `Equality[Set[S]]` definition in algebird-test; once ScalaTest 3.x is released we can get rid again of this custom equality definition.

There are still two failing test cases for `Array[Byte]` that I need to work on.

miguno:
The code is ready for review.

Notes:

* I changed plans slightly regarding `contramap`.  The PR provides `contramap` only for the `CMSHasher[K]` typeclass because that makes the API much cleaner (otherwise we'd have to spread `contramap` variants all of the place -- `CMSMonoid`, `TopPctCMSMonoid`, and so on).  I added docs to explain how `contramap` is supposed to be used, and why it is helpful.
* We have built-in support for `Short`, `Int`, `Long`, `Array[Byte]`, `BigInt`, `String`.  Here, `BigInt` and `String` use `contramap` to derive from `Array[Byte]`.
* I needed to make several changes to the test spec because of our new support of `Array[Byte]`.  The reason is that, essentially, `Array#equal()` is totally unusable.
* I improved general docs and also improved the existing test cases a little as a result of my `Array`-based "bug" hunting.  (There was no new bug in CMS, but I improved the tests while looking for the reasons that the new code wasn't working, and I didn't want to throw away these improvements now that the code including `Array` does work.)

Example `contramap` usage to create a `TopPctCMSMonoid` for `Double` based on the built-in `Array[Byte]` support:

```scala
def f(d: Double): Array[Byte] = {
  val l: Long = java.lang.Double.doubleToLongBits(d)
  java.nio.ByteBuffer.allocate(8).putLong(l).array()
}

implicit val cmsHasherDouble: CMSHasher[Double] = CMSHasherArrayByte.contramap((d: Double) => f(d))

val monoid: TopPctCMSMonoid[Double] = TopPctCMS.monoid[Double](eps, delta, seed, 0.01)
```

miguno:
The Travis build failed due to an unrelated error:

```
[error] (*:update) sbt.ResolveException: unresolved dependency: com.typesafe.sbt#sbt-git;0.6.1: not found

The command "sbt ++2.11.4 test" exited with 1.
```

algebird-core/src/main/scala/com/twitter/algebird/CountMinSketch.scala
```diff
@@ -1027,16 +1082,58 @@ case class TopNCMSAggregator[K](cmsMonoid: TopNCMSMonoid[K])
...
   def hash(a: Int, b: Int, width: Int)(x: K): Int
 
+  /**
+   * Given `f`, a function from `L` into `K`, creates a `CMSHasher[L]` whose hash function is equivalent to:
+   *
+   * {{{
+   * def hash(a: Int, b: Int, width: Int)(x: L): CMSHasher[L] = CMSHasher[K].hash(a, b, width)(f(x))
+   * }}}
+   */
+  def on[L](f: L => K)(implicit hasher: CMSHasher[K]) = new CMSHasher[L] {
```
johnynek: why do we need the implicit hasher here rather than doing `self.hash(a, b, width)(f(x))` below? This method, with the implicit parameter, seems appropriate for the `object` but not the trait. I don't see where this method uses `this` or `self`, which is what makes me think something should change.

algebird-core/src/main/scala/com/twitter/algebird/CountMinSketch.scala
```diff
@@ -1027,16 +1082,58 @@ case class TopNCMSAggregator[K](cmsMonoid: TopNCMSMonoid[K])
...
+   * {{{
+   * def f(d: Double): Array[Byte] = {
+   *   val l: Long = java.lang.Double.doubleToLongBits(d)
+   *   java.nio.ByteBuffer.allocate(8).putLong(l).array()
+   * }
+   *
+   * implicit val cmsHasherDouble: CMSHasher[Double] = CMSHasherArrayByte.contramap((d: Double) => f(d))
+   * }}}
+   */
+  def contramap[L](f: L => K)(implicit hasher: CMSHasher[K]) = on(f)
```
johnynek: same comment as above.

algebird-core/src/main/scala/com/twitter/algebird/CountMinSketch.scala
```diff
@@ -812,18 +827,44 @@ case class HeavyHitter[K](item: K, count: Long) extends java.io.Serializable
...
 
-  def ordering[K: Ordering]: Ordering[HeavyHitter[K]] = Ordering.by { hh: HeavyHitter[K] => (hh.count, hh.item) }
+  import scala.reflect._
+
+  // Note that we can't define the ordering on `count` only.  We use a set to track heavy hitters, where equality means
+  // two elements are identical.  If we were to use `count` only, we would label different items in heavy hitters as
+  // identical simply because they have the same count (and such false duplicates may happen easily in practice).  For
+  // this reason we need to add another dimension to the ordering to prevent such false duplicates.  For this second
+  // dimension we use an item's default Java hash code because we do not want to introduce an `Ordering` constraint.
+  def ordering[K: ClassTag]: Ordering[HeavyHitter[K]] = {
```
johnynek: damn scala. Equiv and scala.util.hashing.Hashing should be what we want here, but they are so broken and impoverished I guess we should think about it.

I really don't like using ClassTag in general, and I especially don't like it here when it is in fact a round-about way to get at what we really want, which is a hashcode.

So, I guess the least ugly solution is to keep the `Ordering[K]` bound on cases with the TopK heavy hitters. What do you think?

algebird-core/src/main/scala/com/twitter/algebird/CountMinSketch.scala
miguno: Fixed.

algebird-core/src/main/scala/com/twitter/algebird/CountMinSketch.scala
miguno: Fixed.  (And good catch.)

algebird-core/src/main/scala/com/twitter/algebird/CountMinSketch.scala
miguno: I'd be more inclined to say "Damn Java Array!" but otherwise I totally agree. ;-)

> Equiv and scala.util.hashing.Hashing should be what we want here, but they are so broken and impoverished I guess we should think about it.

I haven't used either of those yet, but I understand that you would not recommend them.  (Did you mean "I guess we should /not/ think about it"?)

> I really don't like using ClassTag in general, and I especially don't like it here when it is in fact a round-about way to get at what we really want, which is a hashcode.
> 
> So, I guess the least ugly solution is to keep the Ordering[K] bound on cases with the TopK heavy hitters. What do you think?

I'm torn.  While I don't like `ClassTag` shenanigans that much either -- particularly when they are just a workaround for a different issue (read: Array's messed up equality) -- there's one reason I'd prefer `ClassTag` over `Ordering`:  Not having the `Ordering[K]` bound clarifies the contract.  There is no reason the user should have to provide an `Ordering` -- we only use an Ordering internally to benefit from better update performance of a `SortedSet`, and as you pointed out this is mostly because of the TopK use case.  We do not guarantee any order(ing) at all in the API, which only guarantees a `Set` when retrieving the heavy hitters from a CMS instance.  In other words, requiring an `Ordering` kinda leaks internal, implementation-specific information to the user (well, at least in the current scenario where the API does not guarantee a `SortedSet`).
    Another reason would be to keep the code and the look'n'feel as identical as possible between TopPct and TopK.

On the other hand, keeping the Ordering bound for TopK only might be a sneaky transition path to eventually get rid of TopK altogether, though @jnievelt will not like to hear that. ;-)

> So, I guess the least ugly solution is to keep the Ordering[K] bound on cases with the TopK heavy hitters.

There's also a tiny performance benefit of using an `Ordering` internally for TopPct, as we don't have to traverse the full list of heavy hitters every time we perform an update because `dropWhile` can terminate earlier than `filter`:

```scala
// TopPct with Ordering
override def purgeHeavyHitters(cms: CMS[K])(hitters: HeavyHitters[K]): HeavyHitters[K] = {
  val minCount = heavyHittersPct * cms.totalCount
  // with Ordering/SortedSet: HeavyHitters[K](hitters.hhs.dropWhile { _.count < minCount })
  HeavyHitters[K](hitters.hhs.filter { _.count >= minCount })
}
```

Whether that's an important optimization to warrant keeping `Ordering` I can't tell.  Typically, the list of heavy hitters shouldn't be that huge that terminating early is that big of a benefit, for example.

johnynek: Why is Equiv bad news?

The reason is that in the `object Equiv` there is an implicit Equiv for every type that just defaults to `a.equals(b)`. So, with that implicit hanging out, taking an `Equiv[T]` is not a guarantee of any kind of type-safe equivalence, since if the user does nothing, the default equals gets called. So, in the case of array `Equiv[Array[Int]].equiv(Array(1), Array(1)) == false`. The correct approach, is my view, would be define Equiv for all the primitive types, String and Array in the `object Equiv` and then auto generate them in case class companions (along with Ordering and Hashing).

The reason why Hashing is lame is that it is unknown to most (I just learned of it somewhat recently) and it also has almost no good default instances, so it does not buy users much: they still have to write a lot of code, they just write it a certain way.

I really wish the typeclass story in scala were better.

About pushing Ordering onto only TopK: big fan of that. Only those using the variant that needs it should pay the price. The benefit you note with dropWhile is even more of a win.

miguno: Thanks for the explanation.  Then I don't have to feel to bad about not having heard of `Hashing` before. :-)

> The benefit you note with dropWhile is even more of a win.

Just to be clear:  If we push Ordering to only TopK, then we lose `dropWhile` for TopPct (so that benefit, whether it's tiny or actually noticeable, would be gone) and we'd have to use `filter` instead.  (I made a mistake in the first comment of the Scala snippet above.)

miguno: So I guess the conclusion is:  Revert the `ClassTag` and bring back `Ordering`?

johnynek: I think that's the most correct approach (where correct means that the typeclasses in play are the ones closest to what the code actually requires).

@jnievelt do you have any concerns?

Thank you, sir.

miguno:
Since we haven't heard back from @jnievelt I think I'll count this as silent agreement. :-)  I will revamp the code according to what we discussed above.

miguno:
The equality misbehavior of `Array[Byte]` is really becoming a nuisance to deal with -- to the point where I am considering to drop `Array[Byte]` support from CMS.

Why is that? We can't even track heavy hitters for `Array[Byte]` easily because both `Set` and `SortedSet` -- i.e. regardless of whether we use or do not use `Ordering` -- break for `Array[Byte]` due to its equality semantics.

Example:

```scala
$ ./sbt algebird-core/console

scala> val a1 = HeavyHitter[Array[Byte]](Array(1.toByte), 123)
scala> val a2 = HeavyHitter[Array[Byte]](Array(1.toByte), 123)

// The two objects are NOT equal, "thanks" to the equality semantics of Java arrays.
scala> a1 == a2
res0: Boolean = false
 
// We must first convert/wrap `Array[Byte]` to achieve the correct equality behavior.
scala> val s1 = HeavyHitter[Seq[Byte]](Array(1.toByte).toSeq, 123)
scala> val s2 = HeavyHitter[Seq[Byte]](Array(1.toByte).toSeq, 123)
scala> s1 == s2
res1: Boolean = true

// This also means that HeavyHitters[K] (plural), i.e. the container
// we use to track heavy hitters, will not work correctly either.
```

Unfortunately these Array-related issues are affecting many places in Scala/CMS, it's not just about `Set` and `SortedSet`.  For instance, even the simple function `CMSItem#frequency()` will break for Arrays because the comparison `if (item == x)` will not work correctly for Arrays:

```scala
/**
 * Used for holding a single element, to avoid repeatedly adding elements from sparse counts tables.
 */
case class CMSItem[K](item: K, override val params: CMSParams[K]) extends CMS[K](params) {

  // ...other code removed...

  override def frequency(x: K): Approximate[Long] = if (item == x) Approximate.exact(1L) else Approximate.exact(0L)

}
```

This also has the implication that you can't use a "safe" source `CMS[K]` such as `CMS[String]`, and then use `contramap` to create a equality-safe wrapper for `CMS[Array[Byte]]`.

In general I am not aware of an easy and/or clean way to address these Array equality issues, and from what I understand the attempts to address this at the Scala language level failed in Scala 2.8.

Also: I would rather not want to track heavy hitters by a hash code of the tracked elements (think: using `hashCode` instead of `equals`;  the idea would be we do have good hashing in place, so why not use a hash to determine equality?).  We should really stick to the "true" equality semantics of the various types `K`, otherwise `Set`/`SortedSet` will incorrectly track heavy hitters due to hash collisions for instances of `K`.  Since our hashing is based on 32-bit integers such collisions are pretty likely, particularly for large domains such as variable-length `String` instances.

--

To recap what we have been trying to achieve in this PR and the related discussions:

1. Add contramap functionality to CMS.
2. Also, add built-in `String` support.
3. Consolidate/re-use code to support `String`, `BigInt`, and `Array[Byte]` via a base hasher (Murmur3) for `Array[Byte]` that is re-used by the hashers for `String` and `BigInt`.
4. At some point we also began adding `Array[Byte]` support, kinda because we already had been working on (2) and (3).

So what I am proposing is to skip (4) and "only" do (1), (2), and (3).  (We'd still keep the `Array[Byte]` base hashing in the code for use with `String` and `BigInt` hashing, but we wouldn't expose it as an importable implicit to the user.)

What do you think?  Of course I'd also appreciate any ideas how to address the Array equality misbehavior.

miguno:
PS: I forgot to add:  My comment above refers to us trying to achieve the various goals without being able to leverage `ClassTag` to special-case Arrays.

johnynek:
@miguno what about adding something like:

```scala
object BytesWrapper {
  implicit val ordering: Ordering[BytesWrapper] = new Ordering[BytesWrapper] {
    def compare(a: BytesWrapper, b: BytesWrapper) = {
      ByteBuffer.wrap(a.get).compareTo(ByteBuffer.wrap(b.get))
    }
  }
}
case class BytesWrapper(get: Array[Byte]) extends AnyVal {
  override def hashCode = /* murmur here */
  override def equals(t: Any) = t match {
    case BytesWrapper(that) => java.util.Arrays.equals(get, that)
  }
}
```

Then we can side step the issue and add a comment as to why. What do you think of that?

miguno:
So you mean we'd tell CMS users* to not try to work on a `CMS[Array[Byte]]` directly, but rather convert their original `Array[Byte]` instances to a `BytesWrapper`, and use a `CMS[BytesWrapper]` instead?

If that's the case, would it make sense to add such an `Array[Byte]` wrapper to Bijection?

Either way (Bijection or not) we could proceed with the current PR modifications:  we'd exclude direct `Array[Byte]` support as described above, and tell CMS users to rely on the wrapper to work on `Array[Byte]` if that's what they need.

*If I understand correctly we can't automagically use such a wrapper behind the scenes, given the previous findings.

johnynek:
Sounds right to me. I don't know of a way to automatically wrap. But we could provide the right wrapper so that it is easier.

Note that MinHashSignature is basically the same thing, but it does not get the hashCode and equality right. If we put the hashCode/equals/toString methods in an object we could call them there too.

miguno:
Ok. I'll take a stab at this when I'm back early next week. Thanks for the quick feedback, Oscar!

miguno:
PS: Note that the wrapper cannot be a value class because we must override `hashCode` and `equals`. Doh!

miguno:
To play devil's advocate: Apart from a Murmur3-based `hashCode` the user would achieve the same effect by converting `Array[Byte]` into a `Seq[Int]` (backed by a `WrappedArray`) via `array.toSeq` and then use a `CMS[Seq[Int]]`.

Also, in CMS we use a separate hash code setup that is decoupled from the `hashCode` of type `K`, so whether or not the `Array[Byte]` wrapper would have a sane, Murmur3-based `hashCode` (compared to normal `Array[Byte]`) would not matter for the CMS use case.

miguno:
Please take a look at the latest changes.

- We do not use `ClassTag` anymore.
- We drop direct `Array` (and notably `Array[Byte]`) support.
- For the convenience of Algebird users we include a `Bytes` wrapper class that provides sane implementations of `equals`, `hashCode`, and `toString` as well as an `Ordering` typeclass.  CMS ships with an implicit CMSHasher for `Bytes` so that you do not have to use `contramap` to get a `CMS[Bytes]`.  (using contramap results in performance penalties, which I could confirm during benchmarking)
- I enhanced the docs to explain the use of `contramap` and how to workaround the Array-related issues (e.g. via the new `Bytes` wrapper).

I also did some benchmarking/profiling to compare the performance (and memory impacts etc.) of `Bytes` vs. `Seq[Byte]` to work on `Array[Byte]`.  In a nutshell, both are in the same ballpark.  `Bytes` was slightly faster than `Seq[Int]`, hence I decided to keep it in the code (instead of adding docs that would tell users to manually convert `Array[Byte]` into `Seq[Byte]` via `toSeq`.)

miguno:
Also, we use a `Set` instead of a `SortedSet` now.  The `TopPctCMS` does not care either way, but the `TopNCMS` currently needs to perform a `sort` of the heavy hitters whenever an instance is being updated (read: a count is performed).

algebird-test/src/test/scala/com/twitter/algebird/BytesSpec.scala
```diff
@@ -0,0 +1,89 @@
...
+import org.scalatest.prop.GeneratorDrivenPropertyChecks
+import org.scalatest.{Matchers, WordSpec}
+
+class BytesSpec extends WordSpec with Matchers with GeneratorDrivenPropertyChecks {
+
+  "has a sane equals function" in {
+    val random = new scala.util.Random
+    forAll((Gen.choose(1, 4096), "wordLength")) { (wordLength: Int) =>
+      // Given two different array instances that have the same content.
+      val word = random.nextString(wordLength)
```
johnynek: I think we need some tests to exercise the case that they are not all equal. Right now, equals(t: Any) == (t != null) and compare(a, b) == 0 would pass your tests

miguno: Good catch regarding the `equals` test.  In the latest commit I enhanced the test spec of Bytes in general.

> compare(a, b) == 0 would pass your tests

As part of the changes I also enhanced the `compare` tests, although `compare(a, b) == 0` was already failing the tests prior to the enhancements.  Hence I'm not sure but hope that the enhancements will still cover any gap you found yesterday for `compare`.

johnynek:
Looks good. I'm going to wait on merging this until we publish 0.10 just in case we find some bugs or something on 0.9.

Thanks for doing this.

miguno:
Ok, sounds like a plan.  Thanks from my side for your patience on finalizing this pull request, I do appreciate it.

johnynek:
I guess 0.10 is coming down the pipe, so I'm going to merge this.

PR #400, opened 2015-01-16
Preparer

avibryant:
This is a trait that makes it more convenient to produce complex compositions of Aggregators. For example, if you want to aggregate over a stream of values of type `List[Int]`, and produce both the total (flattened) sum and the largest individual item (if it exists, which it might not if all lists are empty), then with standard aggregators you would need:

````scala
Aggregator
   .fromMonoid[Int]
   .sumBefore
   .join(Aggregator.max.lift.sumBefore)
````

whereas with Preparer you would use:

````scala
 Preparer[List[Int]].flatten.split{p => (p.sum, p.maxOption)}
````
I propose that this is a lot easier to understand.

### DIFF
--- a/algebird-core/src/main/scala/com/twitter/algebird/Preparer.scala
+++ b/algebird-core/src/main/scala/com/twitter/algebird/Preparer.scala
@@ -47,6 +47,8 @@ case class FlatMapPreparer[A, T](prepareFn: A => TraversableOnce[T])
 
   //alias for convenience
   def aggregate[B, C](aggregator: MonoidAggregator[T, B, C]) = monoidAggregate(aggregator)
+
+  def sum(implicit monoid: Monoid[T]) = monoidAggregate(Aggregator.fromMonoid(monoid))
 }
 
 trait HasMonoidAggregate[A, T] {
@@ -84,23 +86,29 @@ trait HasAggregate[A, T] {
     implicit val ordT = Ordering.by(fn)
     aggregate(Aggregator.min[T])
   }
+
+  def sum(implicit sg: Semigroup[T]) = aggregate(Aggregator.fromSemigroup(sg))
+  def reduce(fn: (T, T) => T) = aggregate(Aggregator.fromReduce(fn))
 }
 
 trait HasLift[A, T] extends HasMonoidAggregate[A, T] {
   def lift[B, C](aggregator: Aggregator[T, B, C]): MonoidAggregator[A, Option[B], Option[C]] =
     monoidAggregate(aggregator.lift)
 
-  def head = lift(Aggregator.head)
-  def last = lift(Aggregator.last)
-  def max(implicit ord: Ordering[T]) = lift(Aggregator.max)
-  def maxBy[U: Ordering](fn: T => U) = {
+  def headOption = lift(Aggregator.head)
+  def lastOption = lift(Aggregator.last)
+  def maxOption(implicit ord: Ordering[T]) = lift(Aggregator.max)
+  def maxOptionBy[U: Ordering](fn: T => U) = {
     implicit val ordT = Ordering.by(fn)
     lift(Aggregator.max[T])
   }
 
-  def min(implicit ord: Ordering[T]) = lift(Aggregator.min)
-  def minBy[U: Ordering](fn: T => U) = {
+  def minOption(implicit ord: Ordering[T]) = lift(Aggregator.min)
+  def minOptionBy[U: Ordering](fn: T => U) = {
     implicit val ordT = Ordering.by(fn)
     lift(Aggregator.min[T])
   }
+
+  def sumOption(implicit sg: Semigroup[T]) = lift(Aggregator.fromSemigroup(sg))
+  def reduceOption(fn: (T, T) => T) = lift(Aggregator.fromReduce(fn))
 }
--- a/algebird-core/src/main/scala/com/twitter/algebird/Preparer.scala
+++ b/algebird-core/src/main/scala/com/twitter/algebird/Preparer.scala
@@ -1,0 +1,37 @@
+package com.twitter.algebird
+
+case class Preparer[A, T](prepareFn: A => T) {
+  def map[U](fn: T => U) =
+    Preparer[A, U](fn.compose(prepareFn))
+
+  def flatMap[U](fn: T => TraversableOnce[U]) =
+    FlatPreparer[A, U](fn.compose(prepareFn))
+
+  def aggregate[B, C](aggregator: Aggregator[T, B, C]): Aggregator[A, B, C] =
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: algebird-core/src/main/scala/com/twitter/algebird/Preparer.scala [hunk 1]
```diff
@@ -0,0 +1,132 @@
...
+  def minBy[U: Ordering](fn: T => U) = {
+    implicit val ordT = Ordering.by(fn)
+    aggregate(Aggregator.min[T])
+  }
+
+  def sum(implicit sg: Semigroup[T]) = aggregate(Aggregator.fromSemigroup(sg))
+  def reduce(fn: (T, T) => T) = aggregate(Aggregator.fromReduce(fn))
+}
+
+trait HasLift[A, T] extends HasMonoidAggregate[A, T] {
```
johnynek: needs comments to explain the usage.

algebird-core/src/main/scala/com/twitter/algebird/Preparer.scala
```diff
@@ -0,0 +1,132 @@
...
+
+  def sortedReverseTake(count: Int)(implicit ord: Ordering[T]) =
+    monoidAggregate(Aggregator.sortedReverseTake(count))
+
+  def toList = monoidAggregate(Aggregator.toList)
+  def toSet = monoidAggregate(Aggregator.toSet)
+  def uniqueCount = monoidAggregate(Aggregator.uniqueCount)
+}
+
+trait HasAggregate[A, T] {
```
johnynek: comment please.

algebird-core/src/main/scala/com/twitter/algebird/Preparer.scala
```diff
@@ -0,0 +1,132 @@
+package com.twitter.algebird
+
+trait Preparer[A, T, +This[A, T] <: Preparer[A, T, This]] extends HasMonoidAggregate[A, T] {
```
johnynek: need tests for this to exercise and demo the code.

Also, the scala docs should exists.

avibryant: Yep, just wanted to get feedback on the API before I spent the time documenting it.

algebird-core/src/main/scala/com/twitter/algebird/Preparer.scala
```diff
@@ -0,0 +1,134 @@
...
+
+  //really ought to generate N versions of this...
+  def split[B1, B2, C1, C2](fn: This[A, T] => (Aggregator[A, B1, C1], Aggregator[A, B2, C2])): Aggregator[A, (B1, B2), (C1, C2)] = {
+    val (a1, a2) = fn(this.asInstanceOf[This[A, T]])
+    a1.join(a2)
+  }
+}
+
+object Preparer {
+  def apply[A] = MapPreparer[A, A](identity)
```
johnynek: I wonder if it is worth making `IdentityPreparer[A]` that does not bother calling compose. It might help perf a bit since the jit does not have to realize identity does nothing.

algebird-core/src/main/scala/com/twitter/algebird/Preparer.scala
johnynek: So this is a one pass transformer: compose `A => TraversableOnce[B]` as many times as we like, then aggregate.

This kind of composition comes up so much, I wonder if we should have that as a thing by itself. These things form a category, and combine is `{ a => fn(a).flatMap(gn) }`

avibryant:
@johnynek added some scaladoc and tests for you

algebird-core/src/main/scala/com/twitter/algebird/Preparer.scala
```diff
@@ -0,0 +1,235 @@
...
+  /**
+   * count and following methods all just call monoidAggregate with one of the standard Aggregators.
+   * see the Aggregator object for more docs.
+   */
+
+  def count(pred: T => Boolean) = monoidAggregate(Aggregator.count(pred))
+  def exists(pred: T => Boolean) = monoidAggregate(Aggregator.exists(pred))
+  def forall(pred: T => Boolean) = monoidAggregate(Aggregator.forall(pred))
+  def size = monoidAggregate(Aggregator.size)
+
```
johnynek: shouldn't this have `def sum(implicit m: Monoid[T])`?

johnynek: nevermind, I see it below in both subclasses.

algebird-core/src/main/scala/com/twitter/algebird/Preparer.scala
```diff
@@ -0,0 +1,235 @@
...
+  def split[B1, B2, C1, C2](fn: FlatMapPreparer[TraversableOnce[T], T] => (MonoidAggregator[TraversableOnce[T], B1, C1], MonoidAggregator[TraversableOnce[T], B2, C2])): Aggregator[A, (B1, B2), (C1, C2)] = {
+    val (a1, a2) = fn(FlatMapPreparer.identity[T])
+    a1.join(a2).composePrepare(prepareFn)
+  }
+
+  /**
+   * transform a given Aggregator into a MonoidAggregator by lifting the reduce and present stages
+   * into Option space
+   */
+  def lift[B, C](aggregator: Aggregator[T, B, C]): MonoidAggregator[A, Option[B], Option[C]] =
```
johnynek: can't all these till the end be in Preparer?

avibryant: They could. Originally I had them without the 'Option' suffix (ie, `head`, not `headOption`) and so they would have conflicted with the semigroup versions. But I think it's better to be explicit.

avibryant: (They are now moved)

algebird-core/src/main/scala/com/twitter/algebird/Preparer.scala
```diff
@@ -0,0 +1,235 @@
...
+  def sum(implicit monoid: Monoid[T]) = monoidAggregate(Aggregator.fromMonoid(monoid))
+
+  /**
+   * Split the processing into two parallel aggregations.
+   * You provide a function which produces two different aggregators from this preparer,
+   * and it will return a single aggregator which does both aggregations in parallel.
+   * (See also Aggregator's join method.)
+   *
+   * We really need to generate N versions of this for 3-way, 4-way etc splits.
+   */
```
avibryant: It's irritating that this returns an Aggregator rather than a MonoidAggregator. See issue #403.

johnynek:
:+1:

PR #401, opened 2015-01-16
Add getValueQuantile and unit-test.

reconditesea:
Add getValueQuantile and unit-test.
Fix a bug (at least I think it is) in trueQuantile.

### DIFF
--- a/algebird-core/src/main/scala/com/twitter/algebird/QTree.scala
+++ b/algebird-core/src/main/scala/com/twitter/algebird/QTree.scala
@@ -200,6 +200,20 @@ case class QTree[A](
       (0L, 0L)
     }
   }
+  /**
+   * For a given value and given quantiles list, compute which quantile it fails to in this QTree
+   * @param value The specified value to compute quantile.
+   * @param quantiles The specified quantiles list, need to be sorted.
+   */
+  def getValueQuantile(value: Double, quantiles: Seq[Double]): (Double, Double) = {
+    val quantilePoints = quantiles.map { q: Double =>
+      val (lower, upper) = quantileBounds(q)
+      (lower, upper, q)
+    }
+    val lowerBound = quantilePoints.find(_._2 >= value).map(_._3).getOrElse(1.0)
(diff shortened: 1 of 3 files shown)


algebird-core/src/main/scala/com/twitter/algebird/QTree.scala
```diff
@@ -200,6 +200,20 @@ case class QTree[A](
...
+  /**
+   * For a given value and given quantiles list, compute which quantile it fails to in this QTree
+   * @param value The specified value to compute quantile.
+   * @param quantiles The specified quantiles list, need to be sorted.
+   */
+  def getValueQuantile(value: Double, quantiles: Seq[Double]): (Double, Double) = {
+    val quantilePoints = quantiles.map { q: Double =>
+      val (lower, upper) = quantileBounds(q)
+      (lower, upper, q)
+    }
```
johnynek: can you add a few comments:

1. how do you know there is only one item in this list?
2. I am confused about the algorithm have `>=` and `1.0` in both cases. Can you comment a bit?

It seems to me that you want something like:
`quantilePoints.filter { case @b(l, u) => (l <= value && value <= u) }`
Why is that not what we want?

reconditesea: I'm thinking situations like this:
Given an ordered and unique list of quantiles, e.g. [0, 0.1, 0.2, 0.3,... 0.9, 1.0] and specific value x. This method will return the estimated quantile for x in this QT.  Wondering if this is a valid and common use case?

Since the list is ordered and unique, there will be only one item in the list.  Also It's possible that x's quantile is somewhere b/w the two items in the list. For example, x's quantile is 0.25 which is in the middle of [0.2, 0.3]. In this case, I think it should return the the interval as lower and upper bound. So that's why the algorithm uses `>=` to do the filtering. The `1.0` is because if no bounds in the QT is larger than value x, then I think its quantile should be 1.0 since it's larger than all values in this QT.
Does this make sense?

reconditesea: @johnynek  Does the above make sense to you?

ianoc:

PR #407, opened 2015-01-24
EventuallyAggregator and variants

avibryant:
An Aggregator that wraps the `EventuallySemigroup` (and one for `EventuallyMonoid`), as well as a `SetSizeAggregator` that starts with a `Set` and ends with a `HLL`.

### DIFF
--- a/algebird-core/src/main/scala/com/twitter/algebird/HyperLogLog.scala
+++ b/algebird-core/src/main/scala/com/twitter/algebird/HyperLogLog.scala
@@ -451,3 +451,15 @@ case class HyperLogLogAggregator(val hllMonoid: HyperLogLogMonoid) extends Monoi
   def prepare(value: Array[Byte]) = monoid.create(value)
   def present(hll: HLL) = hll
 }
+
+case class SetSizeAggregator[A](hllBits: Int, maxSetSize: Int = 10)(implicit toBytes: A => Array[Byte])
+  extends EventuallyMonoidAggregator[A, HLL, Set[A], Long] {
+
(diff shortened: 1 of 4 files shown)


algebird-core/src/main/scala/com/twitter/algebird/HyperLogLog.scala
```diff
@@ -451,3 +451,15 @@ case class HyperLogLogAggregator(val hllMonoid: HyperLogLogMonoid) extends Monoi
   def prepare(value: Array[Byte]) = monoid.create(value)
   def present(hll: HLL) = hll
 }
+
+case class SetSizeAggregator[A](hllBits: Int, maxSetSize: Int = 10)(implicit toBytes: A => Array[Byte])
```
johnynek: it might make sense to make the default maxSetSize to be something that depends on hllBits. At the extreme you could serialize every item coming in and increment a byte counter (although the in-memory may be much less or more, but still), then when it exceeds `2^hllBits`, switch over.  I would think something like `maxSetSize: Int = math.max(10, 1 << (hllBits - 5))` might be a decent default.

johnynek:
This looks great, but you're pushing our test coverage down here. Can we get at least a simple test to exercise this a bit?

Also, can you add some comments to the type so the scaladoc is not bare?

johnynek:
ping for at least a test or two that exercises this.

johnynek:
ping again for a test.

avibryant:
Sorry, been busy - I'll try to get one in, but there's always the next release.

johnynek:
@DanielleSucher thanks for the tests!

Issue #369, opened 2014-12-05
Add MultiAggregator.apply


johnynek:
```scala
MultiAggregator(agg1, agg2, agg3)
```
is nicer than ```GeneratedTupleAggregator.from3(agg1, agg2, agg3)```

Add an object with 22 apply methods for ergonomic usage.

PR #408, opened 2015-01-29
Add MultiAggregator.apply

DanielleSucher:
Because it'll be nicer to use than GeneratedTupleAggregator.fromN

Finishes issue #369 

@johnynek - still want this? I was just feeling insomniac tonight and glancing through algebird issues to fill for the hell of it.


johnynek:
Nice! Thanks so much. This is a real usability improvement!

joshualande:
This is awesome!

Issue #403, opened 2015-01-17
Joining together MonoidAggregators should produce a MonoidAggregator


avibryant:
Right now if you do any kind of join on multiple instances of MonoidAggregator you get an Aggregator back. It would be nice if, in the case that everything being joined is itself a MonoidAggregator, you got something that was still a MonoidAggregator.

johnynek:
this would be good.

PR #409, opened 2015-01-30
Return a MonoidAggregator from MultiAggregator when possible

DanielleSucher:
Given MonoidAggregators as input, return a MonoidAggregator
instead of just a plain old Aggregator.

cc @avibryant


avibryant:
This closes #403. Yay!

PR #410, opened 2015-02-05
Add SummingWithHitsCache class to also track key hits.

reconditesea:


algebird-core/src/main/scala/com/twitter/algebird/SummingCache.scala
```diff
@@ -76,3 +76,18 @@ class SummingCache[K, V] private (capacity: Int)(implicit sgv: Semigroup[V])
...
+  def apply[K, V: Semigroup](cap: Int): SummingWithHitsCache[K, V] = new SummingWithHitsCache[K, V](cap)
+}
+/**
+ * A SummingCache that also tracks the number of key hits
+ */
+class SummingWithHitsCache[K, V] private (capacity: Int)(implicit sgv: Semigroup[V])
+  extends SummingCache[K, V](capacity)(sgv) {
+
+  def putWithHits(m: Map[K, V]): (Int, Option[Map[K, V]]) = {
+    val keyHits = m.keys.filter(cache.get(_).isDefined).size
```
johnynek: `m.keys.count(cache.contains)`

johnynek:
Let's add some tests, but otherwise this looks good. What do you think about the private constructor? That was unusual before and caused a problem here. Should we just make the constructor public? What do we gain from private constructor? (I did it before, but I think it was a mistake now).

reconditesea:
I actually don't know the reason of why it's private in the first place. So I just changed it to protected to make this work. If you feel the private accessor is not needed any more, I think let's go ahead change it to public.

reconditesea:
@johnynek Added several tests for it.

algebird-core/src/main/scala/com/twitter/algebird/SummingCache.scala
```diff
@@ -76,3 +76,18 @@ class SummingCache[K, V] private (capacity: Int)(implicit sgv: Semigroup[V])
...
   }).asScala
 }
+
+object SummingWithHitsCache {
+  def apply[K, V: Semigroup](cap: Int): SummingWithHitsCache[K, V] = new SummingWithHitsCache[K, V](cap)
+}
+/**
+ * A SummingCache that also tracks the number of key hits
+ */
+class SummingWithHitsCache[K, V] private (capacity: Int)(implicit sgv: Semigroup[V])
```
johnynek: let's make this one public too

johnynek:
after we fix the public issue, LGTM.

algebird-core/src/main/scala/com/twitter/algebird/SummingCache.scala
reconditesea: Updated.
Just out of curiosity, what's the original reason to make the constructor private?

PR #411, opened 2015-02-07
Add MapAggregator to compose tuples of (key, agg) pairs

DanielleSucher:
Add a MapAggregator to join tuples of (K, Aggregator[A, Bn, C]) pairs
into a composed Aggregator[A, TupleN[B1,...,BN], Map[K, C]].

cc @avibryant @mlmanapat

### DIFF
--- a/algebird-test/src/test/scala/com/twitter/algebird/TupleAggregatorsTest.scala
+++ b/algebird-test/src/test/scala/com/twitter/algebird/TupleAggregatorsTest.scala
@@ -227,110 +227,110 @@ class TupleAggregatorsTest extends WordSpec with Matchers {
     }
 
     val longData = data.map{ _.toLong }
-    val MinMonoidAgg = Aggregator.size
+    val SizeAgg = Aggregator.size
 
     "Create a MonoidAggregator from a tuple of 2 MonoidAggregators" in {
-      val agg: MonoidAggregator[Long, Tuple2[Long, Long], Tuple2[Long, Long]] = MultiAggregator(MinMonoidAgg, MinMonoidAgg)
+      val agg: MonoidAggregator[Long, Tuple2[Long, Long], Tuple2[Long, Long]] = MultiAggregator(SizeAgg, SizeAgg)
       assert(agg(longData) == Tuple2(6, 6))
     }
 
     "Create a MonoidAggregator from a tuple of 3 MonoidAggregators" in {
-      val agg: MonoidAggregator[Long, Tuple3[Long, Long, Long], Tuple3[Long, Long, Long]] = MultiAggregator(MinMonoidAgg, MinMonoidAgg, MinMonoidAgg)
+      val agg: MonoidAggregator[Long, Tuple3[Long, Long, Long], Tuple3[Long, Long, Long]] = MultiAggregator(SizeAgg, SizeAgg, SizeAgg)
       assert(agg(longData) == Tuple3(6, 6, 6))
     }
 
     "Create a MonoidAggregator from a tuple of 4 MonoidAggregators" in {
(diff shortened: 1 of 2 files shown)


avibryant:
I could be wrong about this but I think it would be more usable if the `apply` methods took N Tuple2 parameters instead of a single TupleN[Tuple2...]. As a practical matter, that means you wouldn't need the extra set of params and constructing a `MapAggregator` would look almost exactly like constructing a `Map`: `MapAggregator("min" -> Aggregator.min, "max" -> Aggregator.max)`.

algebird-test/src/test/scala/com/twitter/algebird/TupleAggregatorsTest.scala
```diff
@@ -334,4 +334,638 @@ class TupleAggregatorsTest extends WordSpec with Matchers {
       assert(agg(longData) == Tuple22(6, 6, 6, 6, 6, 6, 6, 6, 6, 6, 6, 6, 6, 6, 6, 6, 6, 6, 6, 6, 6, 6))
     }
   }
+
+  "MapAggregator" should {
+    import MapAggregator._
+
+    "Create an aggregator from 2 (key, aggregator) pairs" in {
```
avibryant: these tests would be more informative if there were different aggregators used for each key. It would also be a better test if the aggregators had different intermediate types (like, the `B` in the agg's A, B, C type params).

algebird-test/src/test/scala/com/twitter/algebird/TupleAggregatorsTest.scala
```diff
@@ -334,4 +334,638 @@ class TupleAggregatorsTest extends WordSpec with Matchers {
...
+        ("key2", MinAgg),
+        ("key3", MinAgg)))
+      val expectedMap = Map(
+        "key1" -> 0,
+        "key2" -> 0,
+        "key3" -> 0)
+      assert(agg(data) == expectedMap)
+    }
+
+    "Create an aggregator from 4 (key, aggregator) pairs" in {
```
avibryant: I suspect that if it works up to 3 we don't need to test 4..22.

johnynek:
+1 to Avi's suggestion. Also, not that it is super important: but are you planning to use this or is it just a nice generalization? The reason I ask is that I'm surprised that the result type (V is the map) is homogenous across a wide variety of aggregations. Of course, Double might be common enough.

algebird-test/src/test/scala/com/twitter/algebird/TupleAggregatorsTest.scala
johnynek: We're adding code coverage metrics. Keep it. :)

avibryant:
Yeah, `Double` is quite common for us. Also, you can imagine using an `Either` type or something similar (like brushfire's 4-way `Dispatched`) to unify disparate aggregation results into a single `V`.

DanielleSucher:
Totally convinced, made the suggested changes!  cc @mlmanapat

johnynek:
Thanks, @DanielleSucher!

vidma:
For consistency/simplicity-of-use, we're missing the ability to generate Tuple or Map aggregator from one entry!

at the moment we'he hacked up this:
```scala
  /**
   * makes an aggregator returning a Tuple1[Metric]
   * this allows to keep aggregator definition consistent with GeneratedTupleAggregator.fromN
   */
  def tupleAggrFrom1[A, B, C](agg: Aggregator[A, B, C]): Aggregator[A, B, scala.Tuple1[C]] = {
    agg.andThenPresent(Tuple1(_))
  }

  // for the MapAggregator
  def mapAggrFrom1[A, B, C <: AnyVal](aggDef: (String, Aggregator[A, B, C])): Aggregator[A, B, Map[String, AnyVal]] = {
    val (key, agg) = aggDef
    agg.andThenPresent(value => Map(key -> value))
  }

```

PR #414, opened 2015-02-09
Add coverageAggregate

johnynek:
We need this to get better, but still not accurate, results due to: https://github.com/scoverage/sbt-scoverage/issues/87


ianoc:
lgtm, merge when green

caniszczyk:
This is good but we fail when we upload results to coveralls:

https://travis-ci.org/twitter/algebird/jobs/50114430#L1321

ianoc:

johnynek:
It looks like the tests take too long to run when doing the coverage tracking.

Maybe we can turn down the number of trials in the CMS tests. They seem to be the slow ones in my check:
https://github.com/twitter/algebird/blob/develop/algebird-test/src/test/scala/com/twitter/algebird/CountMinSketchTest.scala#L267

Maybe we can turn down the number of instances on some of these?

johnynek:
I think it is actually working now: https://coveralls.io/github/twitter/algebird
/cc @sriramkrishnan

johnynek:
@ianoc I guess this is green now, and we can start showing code coverage. Merge?

sriramkrishnan:
Nice. Does it also work if the tests are in a separate package now?

sriramkrishnan:
https://github.com/scoverage/sbt-scoverage/issues/87 is still open.

ianoc:
The code coverage values don't look right do they? It also reports on that website that develop was tested/updated last night. But we've made no changes there.

Looking at the results on that coverage page it seems to report only 28 files are covered... none of which are from algebird-core from what i can see.

johnynek:
about develop: I think that is travis test of the push.

It looks like aggregate testing is indeed broken still. I guess we need a later sbt-coverage plugin, but it appears the coveralls support is lagging, and does not yet support sbt-coverage 1.3.3.

I guess algebird is a uniquely bad case here because all the tests are in another package.

I guess we could try codecov: https://codecov.io/guide

Have to say, this is pretty frustrating shit. We've been trying this for a year or more.

sritchie:
Closing, since we now have codecov.

PR #417, opened 2015-02-17
Add some combinators on MonoidAggregator

johnynek:
Don't merge this until we release 0.10.0. It breaks binary compatibility (adding methods to a trait). I tried to do it with an enrichment, but it hit a scala 2.10 compiler bug.

### DIFF
--- a/algebird-core/src/main/scala/com/twitter/algebird/Aggregator.scala
+++ b/algebird-core/src/main/scala/com/twitter/algebird/Aggregator.scala
@@ -333,7 +333,7 @@ trait MonoidAggregator[-A, B, +C] extends Aggregator[A, B, C] { self =>
   /**
    * Only aggregate items that match a predicate
    */
-  def filter[A1 <: A](pred: A1 => Boolean): MonoidAggregator[A1, B, C] =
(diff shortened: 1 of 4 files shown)


avibryant:
Doesn't need to be this PR, but I guess we could also have something like

`Preparer[Either[A,B]].either{case (a,b) => (a.size, b.sum)}`

johnynek:
@avibryant I don't follow your code, but I did add collect which is somewhat related:

```scala
Preparer[Either[A, B]].collect {
  case Left(a) => (a, 0)
  case Right(b) => (0, b)
}
```

avibryant:
@johnynek yeah the code I proposed would be as shorthand for,eg,

```
Preparer[Either[A, B]].split{p =>
   (p.collect{case Left(a) => a}.size,
  p.collect{case Right(b) => b}.sum)
}
```

johnynek:
@avibryant got it. That is nice actually. +1 for split3, split4 (need numbers in the names because of type erasure, I think).

johnynek:
@ianoc can we merge this before 0.10?

ianoc:
LGTM

PR #418, opened 2015-02-24
Added function to safely downsize a HyperLogLog sketch

pawasranjan:
A new function called "downsize" to safely reduce the size of a HyperLogLog sketch. The error rate goes up correspondingly due to loss of information. Based on the ideas presented in: [Set Operations On HLLs of Different Sizes](http://research.neustar.biz/2012/09/12/set-operations-on-hlls-of-different-sizes/) and its follow-up [HyperLogLog Engineering: Choosing The Right Bits](http://research.neustar.biz/2013/03/25/hyperloglog-engineering-choosing-the-right-bits/).

### DIFF
--- a/algebird-core/src/main/scala/com/twitter/algebird/HyperLogLog.scala
+++ b/algebird-core/src/main/scala/com/twitter/algebird/HyperLogLog.scala
@@ -240,6 +240,65 @@ sealed abstract class HLL extends java.io.Serializable {
    * Set each item in the given buffer to the max of this and the buffer
    */
   def updateInto(buffer: Array[Byte]): Unit
+
+  /**
+   * Common accessor for finding the new rhoW value at j in downsized HLL
+   *
+   * [[http://research.neustar.biz/2012/09/12/set-operations-on-hlls-of-different-sizes/]]
+   *
+   * [[http://research.neustar.biz/2013/03/25/hyperloglog-engineering-choosing-the-right-bits/]]
+   *
+   * @param newJ        j for which the new rhoW is needed
+   * @param reducedBits Number of bits in the reduced HLL
+   * @param getRhoW     Function that returns the current rhoW at j
+   *
+   * @return  The new rhoW at j for an HLL with reducedBits
+   */
+  protected def getNewRhoWAt(newJ: Int, reducedBits: Int, getRhoW: Int => Byte): Byte = {
+    val reducedSize = 1 << reducedBits
+
+    // bit mask to set MSBs to 1. Makes rhoW exit fast
+    val bitMask = 0xFFFFFFFF << bits
+    // for j in HLL, we need to find the max of rhoW
+    // at current j, j + stepSize, ... j + (2^reducedBits - 1) * stepSize
+    val stepSize = 1 << reducedBits
+
+    // initialize offset and max
+    var offset = 0
+    var maxRhoW = 0: Byte
+    // iterate through to find the max over current rhoW
+    while (offset < size) {
+      val modifiedRhoW = {
+        val currentJ = newJ + offset
+        val currentRhoW = getRhoW(currentJ)
+        if (currentRhoW == 0)
+          // register not set, skip
+          currentRhoW
+        else if (currentJ < reducedSize)
+          // the MSBs are guaranteed to be all zero, so we can just add the difference to rhoW
+          (currentRhoW + (bits - reducedBits)).toByte
+        else {
+          // find the number of leading zeros
+          // we use bitMask to force rhoW to stop after going through (bits - reducedBits) bits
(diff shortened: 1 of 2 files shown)


algebird-core/src/main/scala/com/twitter/algebird/HyperLogLog.scala
```diff
@@ -294,6 +353,19 @@ case class SparseHLL(bits: Int, maxRhow: Map[Int, Max[Byte]]) extends HLL {
         buffer.update(idx, buffer(idx) max (maxb.get))
     }
   }
+
+  override def downsize(reducedBits: Int): HLL = {
```
johnynek: why not add `if (reducedBits == bits) this else { ...`

algebird-core/src/main/scala/com/twitter/algebird/HyperLogLog.scala
```diff
@@ -294,6 +353,19 @@ case class SparseHLL(bits: Int, maxRhow: Map[Int, Max[Byte]]) extends HLL {
...
+    require(reducedBits > 3 && reducedBits < bits, s"Use at least 4, but less than $bits bits")
+    val reducedSize = 1 << reducedBits
+    val reducedMaxRhoW = collection.mutable.Map.empty[Int, Max[Byte]]
+    for (i <- 0 until reducedSize) {
```
johnynek: this looks expensive for the sparse case. Can't we instead iterate over the keys that are present instead of all the buckets?

algebird-core/src/main/scala/com/twitter/algebird/HyperLogLog.scala
```diff
@@ -338,6 +410,17 @@ case class DenseHLL(bits: Int, v: IndexedSeq[Byte]) extends HLL {
       idx += 1
     }
   }
+
+  override def downsize(reducedBits: Int): HLL = {
+    require(reducedBits > 3 && reducedBits < bits, s"Use at least 4, but less than $bits bits")
```
johnynek: again, let's go ahead and handle the `reducedBit == bits` case (like .take(0) is allowed, we might as well support this).

algebird-core/src/main/scala/com/twitter/algebird/HyperLogLog.scala
```diff
@@ -338,6 +410,17 @@ case class DenseHLL(bits: Int, v: IndexedSeq[Byte]) extends HLL {
...
+    val reducedSize = 1 << reducedBits
+    val reducedV = Array.fill[Byte](reducedSize)(0: Byte)
+    for (i <- 0 until reducedSize) {
+      val newRhoW = getNewRhoWAt(i, reducedBits, v(_))
+      reducedV.update(i, newRhoW.toByte)
+    }
+    DenseHLL(reducedBits, reducedV.toIndexedSeq)
```
johnynek: can you use IndexedSeqArrayByte here to avoid the copy that toIndexedSeq would do?

algebird-core/src/main/scala/com/twitter/algebird/HyperLogLog.scala
```diff
@@ -240,6 +240,65 @@ sealed abstract class HLL extends java.io.Serializable {
...
+        if (currentRhoW == 0)
+          // register not set, skip
+          currentRhoW
+        else if (currentJ < reducedSize)
+          // the MSBs are guaranteed to be all zero, so we can just add the difference to rhoW
+          (currentRhoW + (bits - reducedBits)).toByte
+        else {
+          // find the number of leading zeros
+          // we use bitMask to force rhoW to stop after going through (bits - reducedBits) bits
+          val buf = new Array[Byte](4)
```
johnynek: might as well allocate this outside the while loop so you can share across iterations for sure.

algebird-test/src/test/scala/com/twitter/algebird/HyperLogLogTest.scala
```diff
@@ -216,6 +224,16 @@ class HyperLogLogTest extends WordSpec with Matchers {
       })
     }
 
+    "correctly downsize sparse HLL" in {
+      testDownsize(10)(10, 7)
+      testDownsize(10)(14, 4)
```
johnynek: can you test if you keep the bits the same it works, also if you increase the bits you get an exception.

pawasranjan:
Changed the logic to step through the source HLL data, instead of all the possible indices in the reduced HLL. This should make Sparse HLL faster by going through only the entries present in the map, and also make Dense HLL faster with better cache utilization from sequential scan of both source and the reduced HLL array.

johnynek:
Thanks for this contribution! Very helpful.

PR #419, opened 2015-02-24
AdaptiveCache

avibryant:
This is an attempt to fix the problem of sometimes having to lower the capacity used by `SummingCache` when experiencing OOM errors during `MapsideReduce` in Scalding.

The general idea is that along with a `SummingCache`, we keep a second, "sentinel" cache that sums the evicted items, but holds onto these values with soft references. If we're ever experiencing memory pressure, the GC will clear these references and notify us. But if the sentinel cache has grown to size X *without* that happening, that's good evidence that we should be ok to grow our main SummingCache by something close to X (since, had we used that larger capacity to begin with, we empirically would have had enough memory to cache all of the values that we evicted instead).


algebird-core/src/main/scala/com/twitter/algebird/AdaptiveCache.scala
```diff
@@ -0,0 +1,115 @@
...
+ * which keeps caching and summing the evicted values.
+ * Once the SentinelCache has grown to the same size as the current cache,
+ * plus some margin, without running out of RAM, then this indicates that we
+ * have enough headroom to double the capacity.
+ */
+class AdaptiveCache[K, V: Semigroup](maxCapacity: Int, growthMargin: Double = 1.5)
+  extends StatefulSummer[Map[K, V]] {
+
+  require(maxCapacity >= 0, "Cannot have negative capacity")
+  var currentCapacity = 1
```
johnynek: can we make the member's private? Especially the vars?

algebird-core/src/main/scala/com/twitter/algebird/AdaptiveCache.scala
```diff
@@ -0,0 +1,115 @@
...
+/**
+ * This is a wrapper around SummingCache that attempts to grow the capacity
+ * by up to some maximum, as long as there's enough RAM.
+ * It determines that there's enough RAM to grow by maintaining a SentinelCache
+ * which keeps caching and summing the evicted values.
+ * Once the SentinelCache has grown to the same size as the current cache,
+ * plus some margin, without running out of RAM, then this indicates that we
+ * have enough headroom to double the capacity.
+ */
+class AdaptiveCache[K, V: Semigroup](maxCapacity: Int, growthMargin: Double = 1.5)
```
johnynek: there is a StatefulSummer law that you can use to test these things without much work. Can you wire that up?

algebird-core/src/main/scala/com/twitter/algebird/AdaptiveCache.scala
```diff
@@ -0,0 +1,115 @@
...
+*/
+
+package com.twitter.algebird
+
+/**
+ * @author Avi Bryant
+ */
+
+import collection.mutable.{ ListBuffer, HashMap }
+import ref.{ SoftReference, ReferenceQueue }
```
johnynek: let's add `java.lang.ref...` here.

algebird-core/src/main/scala/com/twitter/algebird/AdaptiveCache.scala
avibryant: oops, yep, fixed

algebird-core/src/main/scala/com/twitter/algebird/AdaptiveCache.scala
avibryant: it's actually not the java.lang ones, it's the scala.ref...

algebird-core/src/main/scala/com/twitter/algebird/AdaptiveCache.scala
```diff
@@ -0,0 +1,115 @@
...
+
+/**
+ * This is a summing cache whose goal is to grow until we run out of memory,
+ * at which point it clears itself and stops growing.
+ * Note that we can lose the values in this cache at any point;
+ * we don't put anything here we care about.
+ */
+class SentinelCache[K, V](implicit sgv: Semigroup[V]) {
+  private val queue = new ref.ReferenceQueue
+  private val map = new HashMap[K, ref.SoftReference[ListBuffer[V]]]()
```
johnynek: why is there a listbuffer? I may be missing it, but it looks like you always sum so the list always has exactly one value. Did I miss that?

algebird-core/src/main/scala/com/twitter/algebird/AdaptiveCache.scala
```diff
@@ -0,0 +1,115 @@
...
+import collection.mutable.{ ListBuffer, HashMap }
+import ref.{ SoftReference, ReferenceQueue }
+
+/**
+ * This is a summing cache whose goal is to grow until we run out of memory,
+ * at which point it clears itself and stops growing.
+ * Note that we can lose the values in this cache at any point;
+ * we don't put anything here we care about.
+ */
+class SentinelCache[K, V](implicit sgv: Semigroup[V]) {
```
johnynek: the name here bothers me a bit because from the user's standpoint the only useful thing you get is size, right? I wonder why you are summing in here at all? And aren't we summing twice?

algebird-core/src/main/scala/com/twitter/algebird/AdaptiveCache.scala
```diff
@@ -0,0 +1,115 @@
...
+
+  def stopGrowing {
+    grow = false
+    clear
+  }
+
+  def put(in: Map[K, V]) {
+    if (grow) {
+      if (queue.poll.isDefined) {
+        stopGrowing
```
reconditesea: Should we log something here to indicate the GC happens?

avibryant: Yes, I wanted to, but wasn't sure in this context what the best way to do that would be...

algebird-core/src/main/scala/com/twitter/algebird/AdaptiveCache.scala
```diff
@@ -0,0 +1,115 @@
...
+ * plus some margin, without running out of RAM, then this indicates that we
+ * have enough headroom to double the capacity.
+ */
+class AdaptiveCache[K, V: Semigroup](maxCapacity: Int, growthMargin: Double = 1.5)
+  extends StatefulSummer[Map[K, V]] {
+
+  require(maxCapacity >= 0, "Cannot have negative capacity")
+  private var currentCapacity = 1
+
+  private var summingCache = new SummingCache[K, V](currentCapacity)
```
reconditesea: Can we use SummingWithHitsCache here, which is essentially a SummingCache with a counter of key-hits. We possibly can add Hadoop counters to track how many times we will have a miss key, and how many times we will grow the cache capacity.

algebird-core/src/main/scala/com/twitter/algebird/AdaptiveCache.scala
avibryant: This is a possibly unnecessary optimization - by boxing the value in a ListBuffer I avoid having to recreate SoftReference objects all the time.

algebird-core/src/main/scala/com/twitter/algebird/AdaptiveCache.scala
avibryant: - Yes, the size is the only useful thing you get.
- I'm summing here to try to get the most accurate simulation possible of the memory used by the SummingCache for this stream of keys/values. The counterfactual here is "how big would the SummingCache have been if we had not evicted these values". We can only know that by summing.
- We're summing twice in the sense that (in a MapsideReduce context) we are doing a sum here that will happen again on the reducer. But from the perspective of the AdaptiveCache, we aren't redoing any work.

avibryant: BTW open to suggestion for other names, this is really a private class anyway (and I'm happy to move it inside AdaptiveCache).

algebird-core/src/main/scala/com/twitter/algebird/AdaptiveCache.scala
johnynek: Oh, so you just want a mutable cell? Can we do instead:

```scala
class Cell[T](init: T) { var value = init }
```
and just use that?

avibryant: Out of curiosity, why not

````scala
class Cell[T](var value: T)
````

But anyway, yes, will do.

**Comment by avibryant** [comment]
@johnynek added a test and I think addressed your other issues
@reconditesea added a `putWithStats` which is like (and uses) `putWithHits` but also includes a Boolean as to whether it grew, so that this can also be fed into a hadoop counter or whatever. (Alternatively we could feed in the amount by which the capacity grew; since for now it always starts at 1 and doubles, they're basically equivalent).

**Comment by reconditesea** [comment]
LGTM.

**Comment by avibryant** [comment]
Thanks. I'm running some stress tests with it and seeing some OOMs that I'm trying to figure out, so don't merge yet; will update.

**Comment by avibryant** [comment]
@johnynek @reconditesea ok, it turned out that using a SoftReference for every key in the map created too much memory pressure and GC overhead dealing with the refs themselves. I've much simplified it to use a single SoftReference for the entire HashMap, and am just careful to never hold a strong reference to the map while doing anything that's likely to allocate much. It now performs well in stress tests.

**Comment by avibryant** [comment]
Anything else you'd like from this pre-merge?

File: algebird-core/src/main/scala/com/twitter/algebird/AdaptiveCache.scala [hunk 8]
```diff
@@ -0,0 +1,126 @@
...
+  override def flush: Option[Map[K, V]] = {
+    val ret = summingCache.flush
+    sentinelCache.clear
+    ret
+  }
+
+  def isFlushed = summingCache.isFlushed
+
+  private var maxReportedSentinelSize = 0
+  case class Stats(hits: Int, cacheGrowth: Int, sentinelGrowth: Int)
```
reconditesea: Nit pick: call the case class CacheStats?  Nothing major, just it might be confused with the Scalding Stat and these two  are likely to be used together when recording these metrics to Hadoop counters.

avibryant:
@reconditesea re CacheStats, no problem, done.

reconditesea:
Seems there is a conflicts so cannot be automatically merged.
Merge when green.

johnynek:
Looks good to me. We can merge this when it is clean (not sure why this is not clean now, I don't know what has conflicted).

ianoc:
@avibryant *bump* fyi, just blocked on non-clean merge

DanielleSucher:
Ping! This should merge cleanly now.

ianoc:
Merged, thanks

PR #421, opened 2015-02-25
fix property tests

MansurAshraf:
Most property tests such as monoidLaw, monadLaw etc were not running properly due to combination of `org.scalatest.prop.PropertyChecks` with `org.scalacheck.Prop.forAll`. Scalatest mark test as a failure only when an exception is thrown using various matchers etc, most Algebird laws indicate failure by returning `false` inside  `org.scalacheck.Prop.forAll`. Due to this mismatch Scalatest was passing all the laws even when they failed as no exception was thrown. I have changed the tests to use `org.scalatest.prop.Checkers` instead of `org.scalatest.prop.PropertyChecks` which delegates to Scalatest for test verification keeping the old behavior of `false` for failure `true` for success and `error` for exception

### DIFF
--- a/algebird-bijection/src/test/scala/com/twitter/algebird/bijection/AlgebirdBijectionLaws.scala
+++ b/algebird-bijection/src/test/scala/com/twitter/algebird/bijection/AlgebirdBijectionLaws.scala
@@ -16,11 +16,9 @@
 
 package com.twitter.algebird.bijection
 
-import org.scalatest.{ PropSpec, Matchers }
-import org.scalatest.prop.PropertyChecks
-import org.scalacheck.{ Arbitrary, Properties }
+import com.twitter.algebird.CheckProperties
 
-class AlgebirdBijectionLaws extends PropSpec with PropertyChecks with Matchers {
+class AlgebirdBijectionLaws extends CheckProperties {
   // TODO: Fill in tests. Ideally we'd publish an algebird-testing
   // module before merging this in.
 }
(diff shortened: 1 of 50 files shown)


MansurAshraf:
@johnynek any idea why monoid laws for Array[Int] be failing?

```
[info] - Array Monoid laws *** FAILED ***
[info]       arg0 = Array()
[info]     )
```

johnynek:
This was added when the tests were broken:
https://github.com/twitter/algebird/blob/develop/algebird-core/src/main/scala/com/twitter/algebird/Monoid.scala#L127

in fact it helped us notice they are broken.

I guess there is a corner case there.

One idea is to use labels on Props to see what part of the law failed:
https://github.com/rickynils/scalacheck/blob/master/src/main/scala/org/scalacheck/Prop.scala#L138

But the problem is the return types of all these things:
https://github.com/twitter/algebird/blob/develop/algebird-test/src/main/scala/com/twitter/algebird/BaseProperties.scala

are Boolean, not Prop (which should probably be changed, I suppose).

In the mean time, can you load the repl and see what is going wrong?

My guess, since it is only one arg is that isNonZero is broken (because Array equality is reference equality). I guess isNonZero(a: Array[T]) = a.nonEmpty might fix it.

The isNonZero is a pretty busted idea. We should have taken Equiv (or our own typeclass) for that, and these cases where there are many representations of zero (Array(), Array(0), Array(0, 0)) are really problematic.

johnynek:
So glad to have these back running.

miguno:
(Just chuckled because it was Arrays again -- for better or worse. :-P)

PR #422, opened 2015-02-27
Make Preparer extend Serializable

franklinhu:
Aggregators created through the Preparer interface retain references to the original Map/FlatMap Preparer instance. For Summingbird Storm jobs, this causes issues since the aggregators need to be Java serializable.
While we could try to construct the instances such that we don't retain any unnecessary references, it seems simpler to just make Preparer extend Serializable to avoid any issues in the future.

cc @avibryant


avibryant:
LGTM

johnynek:
merge when green!

PR #424, opened 2015-03-05
Make MutableBackedMap Serializable.

tixxit:
Makes `MutableBackedMap` `Serializable` so that the result of `Semigroup.sumOption` can be serialized.


tixxit:
The failure looks to be unrelated/spurious.

johnynek:
sorry for the noisey test. Our approximate laws are pretty poorly configured. Restarted the build. Thanks for the fix (not sure why it never came up before. Are you using Kryo here? Is this with Spark?).

tixxit:
Yes, found this using Spark.

PR #425, opened 2015-03-07
Adding support for bigger HLL's for sparse HLL's

FlavSF:
Made a bigger HLL. Should be backwards compatible with existing serialized data.

### DIFF
--- a/algebird-test/src/test/scala/com/twitter/algebird/HyperLogLogTest.scala
+++ b/algebird-test/src/test/scala/com/twitter/algebird/HyperLogLogTest.scala
@@ -153,6 +153,7 @@ class HyperLogLogTest extends WordSpec with Matchers {
     }
     "Correctly serialize" in {
       (4 to 28).foreach { bits =>
+        println(bits) // Hack to get around the "No output has been received in the last 10 minutes" Travis timeout
         val mon = new HyperLogLogMonoid(bits)
         // Zero
         verifySerialization(mon.zero)
--- a/algebird-test/src/test/scala/com/twitter/algebird/HyperLogLogTest.scala
+++ b/algebird-test/src/test/scala/com/twitter/algebird/HyperLogLogTest.scala
@@ -153,16 +153,18 @@ class HyperLogLogTest extends WordSpec with Matchers {
     }
     "Correctly serialize" in {
       (4 to 28).foreach { bits =>
-        println(bits) // Hack to get around the "No output has been received in the last 10 minutes" Travis timeout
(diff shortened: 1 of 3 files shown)


johnynek:
@FlavSF why do you want such big HLLs? If we can go up to 31 bits, that means we are keeping a `2^31` length array, if I'm not mistaken. The error here would be something like `1/sqrt(2 billion) = 0.002%, but at the cost of storing 2GB arrays.

When would you want this? Why not just track an exact set? We want `2^n < m*S` where `n` is the number of bits, `m` is the number of elements, and `S` is the size of each element. For the case where we have `Long` ids, `S=8`. So that means, `n < log(m) + log(S)`.  In general, you probably want `n` to be MUCH smaller than this because if you have the full set you can do a lot more than estimate the size.

So we are talking about counting billions of uniques here before this becomes a savings, and even then, only in case where we want tiny error rates (`2^20` gives an error of about 0.1%, which is already quite low if you ask me).

FlavSF:
@johnynek I see your point and I agree, I would love to use an exact Set but I couldn't find a way to do so with Tsar. Every attempt that I made at changing the values of errPercent or exactCutoff all ended up using HLLs. Maybe I'm misusing Tsar, or maybe there's a feature missing. This PR was a hack to fix the errors I was seeing.

In the specific job that we are trying to write, a 0.1% approximation actually translates to about 200-500k MAUs, which is actually quite a big number so we were trying to get down to half that if not lower.

algebird-test/src/test/scala/com/twitter/algebird/HyperLogLogTest.scala
```diff
@@ -152,7 +152,7 @@ class HyperLogLogTest extends WordSpec with Matchers {
       }
     }
     "Correctly serialize" in {
-      (4 to 20).foreach { bits =>
+      (4 to 31).foreach { bits =>
```
johnynek: going to 31 is probably going to OOM no matter what for dense HLLs. You might want to only go to 29 or 28 or something, something just large enough to exercise this new code.

johnynek:
1) I think your change is actually okay, the more I think about it. No reason to limit the library, but I tend to think such giant HLLs are going to cause problems (as we are seeing trying to test it).

2) I think the dense versions are going to be problems if they get that large. Can we think of a way to exercise only the sparse case? I think if we just add 1 or two items that should keep it sparse. Related: if the HLL is that huge, I'm confused why we wouldn't always be sparse here.

FlavSF:
1) Yes, I'm also having some doubts about those very large HLLs, testing is clearly proving to be a pain even with an increased max heap size of 4GB.

2) I don't know enough about Sparse/Dense to answer you here.

PR #426, opened 2015-03-14
A couple of performance optimizations: HyperLogLog and BloomFilter

Bekbolatov:
Added possibility of HLL creation from pre-computed hashes to HyperLogLogMonoid.
This makes it possible to re-use pre-computed hashes and increase performance.

### DIFF
--- a/algebird-core/src/main/scala/com/twitter/algebird/BloomFilter.scala
+++ b/algebird-core/src/main/scala/com/twitter/algebird/BloomFilter.scala
@@ -112,6 +112,8 @@ sealed abstract class BF extends java.io.Serializable {
 
   def +(other: String): BF
 
+  def checkAndAdd(item: String): (BF, ApproximateBoolean)
+
   def contains(item: String): ApproximateBoolean
 
   // Estimates the cardinality of the set of elements that have been
@@ -128,6 +130,8 @@ case class BFZero(hashes: BFHash, width: Int) extends BF {
   def ++(other: BF) = other
 
   def +(other: String) = BFItem(other, hashes, width)
+
+  def checkAndAdd(other: String): (BF, ApproximateBoolean) = (this + other, ApproximateBoolean.exactFalse)
 
   def contains(item: String) = ApproximateBoolean.exactFalse
 
@@ -149,6 +153,8 @@ case class BFItem(item: String, hashes: BFHash, width: Int) extends BF {
   }
 
   def +(other: String) = this ++ BFItem(other, hashes, width)
+
+  def checkAndAdd(other: String): (BF, ApproximateBoolean) = (this + other, ApproximateBoolean.exact(other == item))
 
   def contains(x: String) = ApproximateBoolean.exact(item == x)
 
@@ -187,6 +193,11 @@ case class BFSparse(hashes: BFHash, bits: CBitSet, width: Int) extends BF {
       width)
   }
 
+  def checkAndAdd(other: String): (BF, ApproximateBoolean) = {
+    val containedApproximate = this.contains(other)
+    (this + other, containedApproximate)
+  }
+
   def contains(item: String): ApproximateBoolean = dense.contains(item)
 
   def size: Approximate[Long] = dense.size
@@ -223,6 +234,12 @@ case class BFInstance(hashes: BFHash, bits: BitSet, width: Int) extends BF {
     BFInstance(hashes,
       bits ++ bitsToActivate,
       width)
+  }
+
+  def checkAndAdd(item: String): (BF, ApproximateBoolean) = {
+    val bitsToActivate = BitSet(hashes(item): _*)
+    val contained = this.contains(item)
+    (BFInstance(hashes, bits ++ bitsToActivate, width), contained)
   }
 
   def bitSetContains(bs: BitSet, il: Int*): Boolean = {
--- a/algebird-core/src/main/scala/com/twitter/algebird/BloomFilter.scala
+++ b/algebird-core/src/main/scala/com/twitter/algebird/BloomFilter.scala
@@ -235,7 +235,12 @@ case class BFInstance(hashes: BFHash, bits: BitSet, width: Int) extends BF {
   }
 
   def +(item: String): BFInstance = {
-    val bitsToActivate = BitSet(hashes(item): _*)
+    val itemHashes = hashes(item)
+    this.+(itemHashes: _*)
+  }
+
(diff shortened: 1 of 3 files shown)


Bekbolatov:
- Possibility of HLL creation from pre-computed hashes to HyperLogLogMonoid.
- Performance optimization for BloomFilter checkAndAdd operation.

Bekbolatov:
Basically, current changes try to save time by re-using hashed values, in two useful situations. BloomFilter checkAndAdd will allow for a little error during partition merges (reduce): when the same item appears in two different partitions once. Worst case is when all partitions contain exactly same sets of items and items appears at most once in each partition.

algebird-core/src/main/scala/com/twitter/algebird/BloomFilter.scala
```diff
@@ -225,6 +236,12 @@ case class BFInstance(hashes: BFHash, bits: BitSet, width: Int) extends BF {
...
   }
 
+  def checkAndAdd(item: String): (BF, ApproximateBoolean) = {
+    val bitsToActivate = BitSet(hashes(item): _*)
+    val contained = this.contains(item)
+    (BFInstance(hashes, bits ++ bitsToActivate, width), contained)
+  }
+
   def bitSetContains(bs: BitSet, il: Int*): Boolean = {
     il.map{ i: Int => bs.contains(i) }.reduce{ _ && _ }
```
johnynek: this could be optimized by stopping at the first false:
```scala
val ilit = il.iterator
var contains = true
while(ilit.hasNext && contains) {
  contains = contains && bs.contains(ilit.next)
}
contains
```
it is mutable, but it might speed things up measurably.

algebird-core/src/main/scala/com/twitter/algebird/HyperLogLog.scala
```diff
@@ -46,6 +46,10 @@ object HyperLogLog {
   def hash(input: Array[Byte]): Array[Byte] = {
     val seed = 12345678
     val (l0, l1) = MurmurHash128(seed)(input)
+    pairLongs2Bytes(l0, l1)
+  }
+
+  def pairLongs2Bytes(l0: Long, l1: Long): Array[Byte] = {
```
johnynek: can this be private?

algebird-core/src/main/scala/com/twitter/algebird/BloomFilter.scala
```diff
@@ -150,6 +154,8 @@ case class BFItem(item: String, hashes: BFHash, width: Int) extends BF {
 
   def +(other: String) = this ++ BFItem(other, hashes, width)
```
johnynek: in ++ should we optimize the case where `other.item == item` and not allocate? I guess that's rare.

algebird-core/src/main/scala/com/twitter/algebird/BloomFilter.scala
```diff
@@ -150,6 +154,8 @@ case class BFItem(item: String, hashes: BFHash, width: Int) extends BF {
...
 
+  def checkAndAdd(other: String): (BF, ApproximateBoolean) = (this + other, ApproximateBoolean.exact(other == item))
```
johnynek: if we are checking equality here anyway, perhaps we should not use + if other == item?

algebird-core/src/main/scala/com/twitter/algebird/BloomFilter.scala
```diff
@@ -225,6 +236,12 @@ case class BFInstance(hashes: BFHash, bits: BitSet, width: Int) extends BF {
       width)
   }
 
+  def checkAndAdd(item: String): (BF, ApproximateBoolean) = {
+    val bitsToActivate = BitSet(hashes(item): _*)
```
johnynek: let's not duplicate code here. Why not just use the default implementation of + below?

algebird-core/src/main/scala/com/twitter/algebird/BloomFilter.scala
Bekbolatov: Sure, that would save an extra hashes calculation when we have a dataset with repeating element.
Rewriting to:
```
  def checkAndAdd(other: String): (BF, ApproximateBoolean) = {
    if (other == item) {
      (this, ApproximateBoolean.exactTrue)
    } else {
      (this + other, ApproximateBoolean.exactFalse)
    }
  }
```

algebird-core/src/main/scala/com/twitter/algebird/BloomFilter.scala
Bekbolatov: An extra "if <string equality>" shortcut is definitely cheaper than hashes calculations - benefit of this trade-off will depend on data consumed. Probably doesn't hurt to put the check there - but as you said, it would happens pretty rarely - data values must be same. Your call.

algebird-core/src/main/scala/com/twitter/algebird/HyperLogLog.scala
Bekbolatov: yes

algebird-core/src/main/scala/com/twitter/algebird/BloomFilter.scala
Bekbolatov: Defintiely. I considered this ealier - performance measurements showed hash calculations and bit array conversions trumped everything small like this. If we do add this, how about:     
```
il.foreach { i => if (!bs.contains(i)) return false }
    return true
```
Not sure how we feel about break-out returns on Scala, but we will avoid extra assignments.

algebird-core/src/main/scala/com/twitter/algebird/BloomFilter.scala
Bekbolatov: Good catch, actually this was not complete - to re-use hashes we will need more changes - all in latest commit.

Bekbolatov: We could also potentially not even add the new item after we figure it "contains" returns true in BFInstance, but our bit array wouldn't represent the presence of that element accurately if "contains" was only due to a collision.

This is not a problem if we just use this implementation, but it can be a problem if someone exports the bit array, or decides to fold or expand the Bloom filter (change capacity). We probably should explicitly add the element when we have more than one element there already.

johnynek:
Thanks. This will be of help!

PR #427, opened 2015-03-17
Adds a presenting benchmark and optimizes it

ianoc:
Benchmark results :

Old code:
https://gist.github.com/ianoc/980b369dd6d5354ce896

New code:
https://gist.github.com/ianoc/41afcafd7099948dbc84

Speedup: a little over 2.4x average ignoring outliers that would bias towards new code more(so if pathological cases new code still does better).

### DIFF
--- a/algebird-core/src/main/scala/com/twitter/algebird/HyperLogLog.scala
+++ b/algebird-core/src/main/scala/com/twitter/algebird/HyperLogLog.scala
@@ -25,7 +25,7 @@ case class ReadOnlyBoxedArrayByte(private val inArray: Array[Byte]) {
   // Need to do underlying array equality sensibly
   override def equals(o: Any) = {
     o match {
-      case ReadOnlyBoxedArray(that) => java.util.Arrays.equals(inArray, that)
+      case ReadOnlyBoxedArrayByte(that) => java.util.Arrays.equals(inArray, that)
       case _ => false
     }
   }
@@ -407,13 +407,20 @@ case class DenseHLL(bits: Int, v: ReadOnlyBoxedArrayByte) extends HLL {
   val (zeroCnt, z) = {
     var count: Int = 0
     var res: Double = 0
-    v.getArray.foreach { mj: Byte =>
+
+    // goto while loop to avoid closure
+    val arr: Array[Byte] = v.getArray
+    val arrSize: Int = arr.size
+    var idx: Int = 0
+    while (idx < arrSize) {
+      val mj = arr(idx)
       if (mj == 0) {
         count += 1
         res += 1.0
       } else {
         res += java.lang.Math.pow(2.0, -mj)
       }
+      idx += 1
     }
     (count, 1.0 / res)
   }
@@ -448,8 +455,13 @@ case class DenseHLL(bits: Int, v: ReadOnlyBoxedArrayByte) extends HLL {
   val toDenseHLL = this
   def updateInto(buffer: Array[Byte]): Unit = {
     assert(buffer.length == size, "Length mismatch")
-    var idx = 0
-    v.getArray.foreach { maxb =>
+
+    val arr: Array[Byte] = v.getArray
+    val arrSize: Int = arr.size
+    var idx: Int = 0
+
+    while (idx < arrSize) {
+      val maxb = arr(idx)
       buffer.update(idx, (buffer(idx)) max maxb)
       idx += 1
     }
--- a/algebird-core/src/main/scala/com/twitter/algebird/HyperLogLog.scala
+++ b/algebird-core/src/main/scala/com/twitter/algebird/HyperLogLog.scala
@@ -18,6 +18,19 @@ package com.twitter.algebird
 
 import java.nio.ByteBuffer
 
+case class ReadOnlyBoxedArrayByte(private val inArray: Array[Byte]) {
+  def apply(idx: Int) = inArray.apply(idx)
+  def size = inArray.size
+  private[algebird] def getArray = inArray
+  // Need to do underlying array equality sensibly
+  override def equals(o: Any) = {
+    if (o.isInstanceOf[ReadOnlyBoxedArrayByte]) {
+      val oAsTpe = o.asInstanceOf[ReadOnlyBoxedArrayByte]
+      java.util.Arrays.equals(inArray, oAsTpe.inArray)
+    } else false
+  }
+}
+
 /** A super lightweight (hopefully) version of BitSet */
 case class BitSetLite(in: Array[Byte]) {
   def contains(x: Int): Boolean = {
@@ -70,7 +83,8 @@ object HyperLogLog {
     buf
   }
 
-  def twopow(i: Int): Double = scala.math.pow(2.0, i)
+  @inline
+  def twopow(i: Int): Double = java.lang.Math.pow(2.0, i)
 
   /**
    * the value 'j' is equal to <w_0, w_1 ... w_(bits-1)>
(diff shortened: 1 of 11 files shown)


algebird-core/src/main/scala/com/twitter/algebird/HyperLogLog.scala
```diff
@@ -18,6 +18,19 @@ package com.twitter.algebird
 
 import java.nio.ByteBuffer
 
+case class ReadOnlyBoxedArrayByte(private val inArray: Array[Byte]) {
+  def apply(idx: Int) = inArray.apply(idx)
+  def size = inArray.size
+  private[algebird] def getArray = inArray
+  // Need to do underlying array equality sensibly
+  override def equals(o: Any) = {
```
johnynek: can we do:
```scala
o match {
  case ReadOnlyBoxedArray(that) => java.util.Arrays.equals(inArray, that)
  case _ => false
}
```

algebird-core/src/main/scala/com/twitter/algebird/HyperLogLog.scala
```diff
@@ -201,11 +214,15 @@ sealed abstract class HLL extends java.io.Serializable {
 
   def approximateSize = asApprox(initialEstimate)
 
-  def estimatedSize = approximateSize.estimate.toDouble
+  def estimatedSize: Double = initialEstimate.toDouble
```
johnynek: why .toDouble, isn't it a double?

ianoc: true, just a hold over, quite right

algebird-core/src/main/scala/com/twitter/algebird/HyperLogLog.scala
```diff
@@ -378,17 +395,28 @@ case class SparseHLL(bits: Int, maxRhow: Map[Int, Max[Byte]]) extends HLL {
...
-  lazy val zeroCnt = v.count { _ == 0 }
-
   // Named from the parameter in the paper, probably never useful to anyone
   // except HyperLogLogMonoid
-  lazy val z = 1.0 / (v.map { mj => HyperLogLog.twopow(-mj) }.sum)
+
+  val (zeroCnt, z) = {
+    var count: Int = 0
+    var res: Double = 0
+    v.getArray.foreach { mj: Byte =>
```
johnynek: while we are perfing out, why not use a while loop here?

ianoc: True, will do that

algebird-core/src/main/scala/com/twitter/algebird/HyperLogLog.scala
```diff
@@ -412,7 +449,7 @@ case class DenseHLL(bits: Int, v: IndexedSeq[Byte]) extends HLL {
   def updateInto(buffer: Array[Byte]): Unit = {
     assert(buffer.length == size, "Length mismatch")
     var idx = 0
-    v.foreach { maxb =>
+    v.getArray.foreach { maxb =>
```
johnynek: while loop?

mosesn:
y'all should consider switching to jmh now that there's decent sbt support.

ianoc:
Will take another look at it when we've the chance, since oscar got the sbt plugin in here, these work reasonably well.

algebird-core/src/main/scala/com/twitter/algebird/HyperLogLog.scala
```diff
@@ -201,11 +214,15 @@ sealed abstract class HLL extends java.io.Serializable {
 
   def approximateSize = asApprox(initialEstimate)
 
-  def estimatedSize = approximateSize.estimate.toDouble
+  def estimatedSize: Double = initialEstimate
+
+  lazy val initialEstimate: Double = {
```
johnynek: why no longer private?

algebird-core/src/main/scala/com/twitter/algebird/HyperLogLog.scala
```diff
@@ -378,17 +395,35 @@ case class SparseHLL(bits: Int, maxRhow: Map[Int, Max[Byte]]) extends HLL {
...
 
   def size = v.size
 
-  lazy val zeroCnt = v.count { _ == 0 }
-
   // Named from the parameter in the paper, probably never useful to anyone
   // except HyperLogLogMonoid
-  lazy val z = 1.0 / (v.map { mj => HyperLogLog.twopow(-mj) }.sum)
+
+  val (zeroCnt, z) = {
```
johnynek: dropped lazy here. That's going to change the performance, right?

miguno:
In https://github.com/twitter/algebird/pull/399 (CMS improvements, scheduled to be merged into Algebird v0.10.0) we have a [Bytes class](https://github.com/miguno/algebird/blob/GH-398/algebird-core/src/main/scala/com/twitter/algebird/Bytes.scala) that looks quite similar to `ReadOnlyBoxedArrayByte`.

Would it make sense to collapse these two classes?

johnynek:
+1 to reuse Bytes class. I forgot about that. What do you think @ianoc

ianoc:
No real objection, these are lighter api changes than most -- but alas since DenseHLL is public it will break the binary compat too. we can combine the two classes sure.

ianoc:
So given these are both open PR's, should we try stack this one on that one? Should we merge the other to develop now? // cc @johnynek  @miguno

miguno:
I think @johnynek wants to not introduce #399 until 0.10.0 to minimize risk of breaking changes.  I am not sure when you want to release 0.10.0, hence I can't really tell whether merging to develop now would be good timing.

The answer to the stacking question may similarly depend on the release plan -- would you be ok with postponing this PR until 0.10.0 / until #399 eventual merge date?

johnynek:
I guess 0.10 is coming up. What do you all think? /cc @isnotinvain We will need to republish scalding + summingbird as well.

ianoc:
I think we should merge and go, i guess there was a target date in Feb for 0.10 -- if its not heavily breaking the upgrades down the line shouldn't be too bad?

johnynek:
@ianoc I merged #399 want to use the same Bytes there and then we'll merge this?

ianoc:
@johnynek rebased and moved over to the bytes class. -- Also just fixes an issue crept in between when the aggregator PR was written and tests for properties changing on develop.

algebird-core/src/main/scala/com/twitter/algebird/HyperLogLog.scala
```diff
@@ -329,31 +329,35 @@ case class SparseHLL(bits: Int, maxRhow: Map[Int, Max[Byte]]) extends HLL {
...
-          case (v, (j, rhow)) =>
-            if (rhow.get > v(j)) {
-              v.updated(j, rhow.get)
-            } else {
-              v
-            }
+        val newContents: Array[Byte] = oldV.array.clone
+        val siz = newContents.size
+
+        maxRhow.foreach {
```
johnynek: should we while loop it here to be consistent? We could get the iterator, and do the hasNext, next unrolling?

Man, we really need some macros for this stuff (glanced at ScalaBlitz, does not seem to be exactly what we want here).

johnynek:
minor suggestion about using one more while loop rather than a foreach while we are optimizing.

algebird-core/src/main/scala/com/twitter/algebird/HyperLogLog.scala
ianoc: Yeah sounds good, just running tests on the change locally.

PR #430, opened 2015-04-02
Use clearer naming conventions for algebraic structures

isnotinvain:
This is a first pass at renaming some of the algebraic structures to be a little easier to understand. A lot of our users wonder just what exactly is a Monad, and how's it different from a Monoid or a Semigroup?

This PR makes the following changes with the aim of alleviating the above confusion:

1) Renames `Monad` to `ChainableCallbackCollectorBuilder` (I'm open to other suggestions though)
2) Renames `Semigroup` to `HasAdditionOperator`
3) Renames `Monoid` to `HasAdditionOperatorAndZero` -- though we could also introduce a trait for `HasZero` and not bother materializing this and just use `HasAdditionOperator with HasZero` when needed

Obviously there's a lot more to do, (Rings, Fields, Groups, and so on), but I wanted to at least get these most common ones cleaned up. 

Other proposals (not in this PR):
Renaming `Injection` to `Converter` and `Bijection` to `ReversibleConverter`
Renaming `Aggregator` to `Averager` or `AssociativeSummaryBuilder` ?

Let me know what you think!
Alex


avibryant:
Those would be good names for traits, but given the way these are used (to create new instances and as a swappable encapsulation of behavior), and the fact that they're not concrete, I think it would be more appropriate to use `AbstractHasAdditionOperatorAndZeroStrategyFactory`.

MansurAshraf:
😂😂😂😂😂😂😂😂 now Algebird is definitely enterprise grade

azymnis:
Took me a while to get it... For a minute I thought that Alex had been
possessed or something

johnynek:
awesome. Sorry I didn't look yesterday. :)

isnotinvain:
Closing because ChainableCallbackCollectorBuilder was not meant to be I suppose, and it's no longer April Fools day.

azymnis:
I recommend that we reopen this issue after we randomly generate some class names [from here](http://projects.haykranen.nl/java/).

PR #431, opened 2015-04-06
upgrade scalacheck

MansurAshraf:
Bringing it to the same version as used internally

### DIFF
--- a/algebird-test/src/test/scala/com/twitter/algebird/VectorSpaceProperties.scala
+++ b/algebird-test/src/test/scala/com/twitter/algebird/VectorSpaceProperties.scala
@@ -16,6 +16,8 @@ limitations under the License.
 
 package com.twitter.algebird
 
+import org.scalacheck.{Gen, Arbitrary}
+
 class VectorSpaceProperties extends CheckProperties {
   import com.twitter.algebird.BaseVectorSpaceProperties._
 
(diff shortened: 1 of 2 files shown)


MansurAshraf:
Following test is failing after the upgrade 

[info] - map int double scaling *** FAILED ***
[info]     Falsified after 1 successful property evaluations.
[info]       arg0 = Map(1039802616 -> 4.4384666053440264E176),
[info]       arg1 = 4.151844009363087E158,
[info]       arg2 = 7.086895641185986E-105
[info]     )

ianoc:

DanielleSucher:
Looks like `map int double scaling` now passes and instead we've moved onto a different failure:

```
[info] - should correctly downsize sparse HLL *** FAILED ***
[info]   0.09900977968481595 was not less than 0.056875 (HyperLogLogTest.scala:127)
```

ianoc:
This branch seems old @DanielleSucher, you want to take this effort over in a new PR maybe? 

-- That test btw is probabilistic so can just fail with a false positive so unless it fails always its probably not broken

DanielleSucher:
@ianoc - Sure, though I was actually thinking of going back and triaging (and seeing which could be closed of) a bunch of old branches here, too.

(Hopefully https://github.com/twitter/algebird/pull/478 will help re the probabilistic tests!)

DanielleSucher:
@ianoc Let's close this one now ^^

ianoc:
Good call, closed

Issue #429, opened 2015-04-01
QTree pruneChildrenWhere is really slow


johnynek:
We are seeing jobs using more than 60% of their CPU in that method. This is uncommon on hadoop (usually the most expensive stuff is serialization, sorting and compression).

avibryant:
Yeah, I was worried about that. I think pruning on every merge is probably crazy; it's just not clear to me what the best heuristic is to decide when it's worth pruning.

sritchie:
@johnynek can we close now that #433 is merged?

PR #433, opened 2015-04-16
Speed up QTree

johnynek:
This makes an algorithmic improvement, and uses sumOption to get pretty big wins in performance which should be enough for map/reduce applications. There are two tricks:

1) we don't need to scan the whole tree if the minCount is 1 or less, which is common, especially for smaller QTrees, which come up when you just add a few together which would be done in map-side aggregation.

2) in sumOption, we can wait to compress after we have added a number of items together. This can be controlled by the user but defaults to 25, which seems to be big enough to give a win in one of the benchmark cases, but also not so huge that I am nervous the QTree is going to get out of hand in size.

Also, I removed the `Monoid[A]` from the QTree instance, which was probably being serialized when people use Kryo, which is no good. Also, it really belongs in the Semigroup (which I think could be a Monoid if we needed, just make a zero with a zero count).

related to #385 

/cc @avibryant 

Benchmark results:
```
Old code:

[info] 0% Scenario{vm=java, trial=0, benchmark=SumOption, depthK=5, numElements=100} 984899.57 ns;
σ=57042.22 ns @ 10 trials
[info] 17% Scenario{vm=java, trial=0, benchmark=SumOption, depthK=10, numElements=100} 1083250.54
ns; σ=5840.94 ns @ 3 trials
[info] 33% Scenario{vm=java, trial=0, benchmark=SumOption, depthK=12, numElements=100} 1137581.99
ns; σ=108320.10 ns @ 10 trials
[info] 50% Scenario{vm=java, trial=0, benchmark=SumOption, depthK=5, numElements=10000} 61440033.33
ns; σ=2959694.26 ns @ 10 trials
[info] 67% Scenario{vm=java, trial=0, benchmark=SumOption, depthK=10, numElements=10000}
1150803500.00 ns; σ=10953173.74 ns @ 4 trials
[info] 83% Scenario{vm=java, trial=0, benchmark=SumOption, depthK=12, numElements=10000}
795943000.00 ns; σ=5517007.01 ns @ 3 trials
[info] 
[info] numElements depthK      us linear runtime
[info]         100      5     985 =
[info]         100     10    1083 =
[info]         100     12    1138 =
[info]       10000      5   61440 =
[info]       10000     10 1150804 ==============================
[info]       10000     12  795943 ====================
[info] 
[info] vm: java
[info] trial: 0
[info] benchmark: SumOption

New code: 
[info]  0% Scenario{vm=java, trial=0, benchmark=SumOption, depthK=5, numElements=100} 131886.03 ns;
σ=189.18 ns @ 3 trials
[info] 17% Scenario{vm=java, trial=0, benchmark=SumOption, depthK=10, numElements=100} 117297.16 ns;
σ=163.64 ns @ 3 trials
[info] 33% Scenario{vm=java, trial=0, benchmark=SumOption, depthK=12, numElements=100} 118775.84 ns;
σ=637.37 ns @ 3 trials
[info] 50% Scenario{vm=java, trial=0, benchmark=SumOption, depthK=5, numElements=10000} 15500888.37
ns; σ=428739.48 ns @ 10 trials
[info] 67% Scenario{vm=java, trial=0, benchmark=SumOption, depthK=10, numElements=10000} 39304494.17
ns; σ=1701883.67 ns @ 10 trials
[info] 83% Scenario{vm=java, trial=0, benchmark=SumOption, depthK=12, numElements=10000} 23124595.24
ns; σ=273888.75 ns @ 10 trials
[info] 
[info] numElements depthK    us linear runtime
[info]         100      5   132 =
[info]         100     10   117 =
[info]         100     12   119 =
[info]       10000      5 15501 ===========
[info]       10000     10 39304 ==============================
[info]       10000     12 23125 =================
[info] 
[info] vm: java
[info] trial: 0
[info] benchmark: SumOption

if we compress after each sum (just looking at the other optimizations):

[info] numElements depthK     us linear runtime
[info]         100      5    227 =
[info]         100     10    119 =
[info]         100     12    125 =
[info]       10000      5  37222 ===
[info]       10000     10 357207 ==============================
[info]       10000     12  61052 =====
```

### DIFF
--- a/algebird-core/src/main/scala/com/twitter/algebird/QTree.scala
+++ b/algebird-core/src/main/scala/com/twitter/algebird/QTree.scala
@@ -38,6 +38,9 @@ package com.twitter.algebird
  */
 
 object QTree {
+  /**
+   * level gives a bin size of 2^level. By default the bin size is 1/65536 (level = -16)
+   */
   def apply[A](kv: (Double, A), level: Int = -16): QTree[A] =
     QTree(math.floor(kv._1 / math.pow(2.0, level)).toLong,
       level,
@@ -55,10 +58,31 @@ object QTree {
       None)
 
   /**
-   * The common case of wanting a count and sum for the same value
+   * The common case of wanting an offset and sum for the same value
+   * This is useful if you want to query the mean inside a range later.
+   * If you truly just care about the counts/histogram, see the value method.
    */
   def apply(k: Long): QTree[Long] = apply(k -> k)
+  /**
+   * uses 1/65636 as the bin size, if you want to control that see other apply
+   * or value methods.
+   *
+   * This is useful if you want to query the mean inside a range later.
+   * If you truly just care about the counts/histogram, see the value method.
+   */
   def apply(k: Double): QTree[Double] = apply(k -> k)
+
+  /**
+   * If you are sure you only care about the approximate histogram
+   * features of QTree, you can save some space by using QTree[Unit]
+   */
+  def value(v: Long): QTree[Unit] = apply(v -> ())
+  /**
+   * If you are sure you only care about the approximate histogram
+   * features of QTree, you can save some space by using QTree[Unit]
+   * level gives a bin size of 2^level. By default this is 1/65536 (level = -16)
+   */
+  def value(v: Double, level: Int = -16): QTree[Unit] = apply(v -> (), level)
 }
 
 class QTreeSemigroup[A](k: Int)(implicit val underlyingMonoid: Monoid[A]) extends Semigroup[QTree[A]] {
@@ -120,7 +144,6 @@ case class QTree[A](
    * for this tree's root and the other tree's root, and return its
    * level (that is, the power of 2 for the interval).
    */
-
   private def commonAncestorLevel(other: QTree[A]) = {
     val minLevel = level.min(other.level)
     val leftOffset = offset << (level - minLevel)
@@ -134,7 +157,14 @@ case class QTree[A](
     ancestorLevel.max(level).max(other.level)
   }
 
-  def merge(other: QTree[A])(implicit monoid: Monoid[A]) = {
+  /**
+   * This merges with another QTree but DOES NOT compress.
+   * You should probably never use this and instead use
+   * QTreeSemigroup.plus(a, b) or .sumOption. Strongly
+   * prefer sumOption if you can, as it is much more efficient
+   * due to compressing less frequently.
+   */
+  def merge(other: QTree[A])(implicit monoid: Monoid[A]): QTree[A] = {
     val commonAncestor = commonAncestorLevel(other)
     val left = extendToLevel(commonAncestor)
     val right = other.extendToLevel(commonAncestor)
@@ -158,8 +188,15 @@ case class QTree[A](
       case (left, None) => left
     }
 
+  /**
+   * give lower and upper bounds respectively of the percentile
+   * value given. For instance, quantileBounds(0.5) would give
+   * an estimate of the median.
+   */
   def quantileBounds(p: Double): (Double, Double) = {
     val rank = math.floor(count * p).toLong
+    // get is safe below, because findRankLowerBound only returns
+    // None if rank > count, but due to construction rank <= count
     (findRankLowerBound(rank).get, findRankUpperBound(rank).get)
   }
 
@@ -190,6 +227,10 @@ case class QTree[A](
     }
   }
 
+  /**
+   * Get the bounds on the sums within a range (not percentile)
+   * This along with the rangeCountBounds can tell you the mean over a range
+   */
   def rangeSumBounds(from: Double, to: Double)(implicit monoid: Monoid[A]): (A, A) = {
     if (from <= lowerBound && to >= upperBound) {
       val s = totalSum
@@ -204,6 +245,9 @@ case class QTree[A](
     }
   }
 
+  /**
+   * Return upper and lower bounds on the counts that appear in a given range
+   */
   def rangeCountBounds(from: Double, to: Double): (Long, Long) = {
     if (from <= lowerBound && to >= upperBound) {
       val s = count
@@ -217,6 +261,12 @@ case class QTree[A](
     }
   }
 
+  /**
+   * Users should never need to call this if they are adding QTrees using the Semigroup
+   * This makes sure no element in the tree has count less than
+   * the total count / 2^k. That means after this call there
+   * are at most 2^k nodes, but usually fewer.
+   */
   def compress(k: Int)(implicit m: Monoid[A]): QTree[A] = {
     val minCount = count >> k
     if ((minCount > 1L) || (count < 1L)) {
@@ -256,11 +306,19 @@ case class QTree[A](
     case n @ None => n // make sure we pass the same ref out
   }
 
+  /**
+   * How many total nodes are there in the QTree.
+   * Not meaningful for learning statistics, but interesting
+   * to estimate serialization size.
+   */
   def size: Int = {
     val childSizes = mapChildrenWithDefault(0){ _.size }
     1 + childSizes._1 + childSizes._2
   }
 
+  /**
+   * Total sum over the entire tree.
+   */
   def totalSum(implicit monoid: Monoid[A]): A = {
     val childSums = mapChildrenWithDefault(monoid.zero){ _.totalSum }
     monoid.plus(sum, monoid.plus(childSums._1, childSums._2))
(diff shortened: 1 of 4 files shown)


avibryant:
Nice work. LGTM.

algebird-core/src/main/scala/com/twitter/algebird/QTree.scala
```diff
@@ -63,8 +61,26 @@ object QTree {
   def apply(k: Double): QTree[Double] = apply(k -> k)
 }
 
-class QTreeSemigroup[A: Monoid](k: Int) extends Semigroup[QTree[A]] {
+class QTreeSemigroup[A](k: Int)(implicit val underlyingMonoid: Monoid[A]) extends Semigroup[QTree[A]] {
```
ianoc: Any reason to move this to an implicit val rather than the type constraint?

Gabriel439:
Why is the test for `HyperLogLog` failing even though this only modifies `QTree`?

ianoc:
Probabilistic tests, occasionally fail... restarted the build

algebird-core/src/main/scala/com/twitter/algebird/QTree.scala
johnynek: So you can have access to it if you need it. For instance, a number of methods in QTree now require a `Monoid[A]`, If you get passed only a QTreeSemigroup, you could not call them, so you can by getting the monoid out.

That's the reasoning. Maybe we can do better.

One idea is to make those methods actually require the QTreeSemigroup, which does not run the risk that somehow a user will set up incoherent instances of the Monoids (more than one in play depending on call site). What do you think?

algebird-core/src/main/scala/com/twitter/algebird/QTree.scala
```diff
@@ -38,23 +38,21 @@ package com.twitter.algebird
...
     QTree(kv._1,
       0,
       1,
       kv._2,
       None,
       None)
-  }
 
   /**
    * The common case of wanting a count and sum for the same value
```
johnynek: this comment is wrong. It is offset and sum for the same value. Right?

algebird-core/src/main/scala/com/twitter/algebird/QTree.scala
```diff
@@ -258,6 +289,7 @@ case class QTree[A](
   }
 
   def interQuartileMean(implicit n: Numeric[A]): (Double, Double) = {
+    implicit val monoid: Monoid[A] = Ring.numericRing[A]
```
johnynek: there is actually an issue here. You can get a Monoid (or a Ring) from a Numeric, but it really does not have to be the same Monoid as you are using elsewhere. So, I guess what we really need is to take both a Numeric and a Monoid, and only use the Numeric to convert toDouble at the end.

algebird-core/src/main/scala/com/twitter/algebird/QTree.scala
ianoc: Ah I see what you were doing here, I hadn't tweaked that the main effect of this change is to make a public field on the Semigroup that gives access to the monoid so you can pull it easily out later. Makes sense.

algebird-core/src/main/scala/com/twitter/algebird/QTree.scala
ianoc: That seems like a pretty decent one, odds of the Numeric being incompatible in its toDouble conversion seem low

algebird-core/src/main/scala/com/twitter/algebird/QTree.scala
jnievelt: That's what implicitly is for, right?  The main benefit of having is as a parameter is that it doesn't have to be implicitly specified.

Gabriel439: I always thought the two approaches were equivalent, thanks to `implicitly` (as @jnievelt mentioned)

ianoc: It would wind up being just longer and redundant, you'd need to add another val to the class or def that was just 
val underlyingMonoid = implicitly[Monoid[A]] 

In either case it'll be picked implicitly, indeed the A: Monoid is just compiler magic for inserting it into the implicit block of the args, but with effectively an anonymous name.

algebird-core/src/main/scala/com/twitter/algebird/QTree.scala
avibryant: It's badly expressed. The point is that for a given bin with bounds (x,y), you are not just counting all k_i where x <= k_i < y, but also summing k_i. To put it another way, you are computing not just the curve of the distribution but the area under that curve (more accurately than if you did some kind of trapezoidal estimation).

Honestly this is just there because sometimes you only really care about the counts, and this is as useful a default for the sum as any.

algebird-core/src/main/scala/com/twitter/algebird/QTree.scala
johnynek: they are totally equivalent. If we want a blue bike shed I could say:

```scala
class QTreeSemigroup[A: Monoid](k: bits) extends Semigroup[QTree[A]] {
  def underlyingMonoid = implicitly[Monoid[A]]
  ...
}
```
that was just more characters. @jnievelt even in the type-bound case you can pass it explicitly if you want. So
```scala
def method[T: TypeClass1: TypeClass2](implicit tc: TypeClass3[T]) = ...
```
will be desugared to:
```scala
def method[T](implicit <somecompilergenid>: TypeClass1[T], <somecompilergenid>: TypeClass2[T], tc: TypeClass3[T]) = ...
```

algebird-core/src/main/scala/com/twitter/algebird/QTree.scala
johnynek: @ianoc just to be sure I understand: you like the solution of taking both Monoid and Numeric?

ianoc: Yep, since I think its far more likely the non-default monoid will be defined by the user in closer scope. But they'll pick up the default Numeric for the type. I doubt there'd be much issue with having the default Numeric and a custom Monoid. Currently here you'd start using the default monoid which would be bad.

algebird-core/src/main/scala/com/twitter/algebird/QTree.scala
johnynek: do you think it is worth using:
```scala
   /**
    * If you are sure you only care about the approximate histogram
    * features of QTree, you can save some space by using QTree[Unit]
    */
   def offset(o: Long): QTree[Unit] = apply(o -> ())
```
At least Unit is zero bytes serialized, and points to the same memory. We use this internally in Tsar which is only concerned with the counts.

avibryant: Yes, though I think `offset` is the wrong name (and that in fact it's totally fine to have that be `apply`)

algebird-core/src/main/scala/com/twitter/algebird/QTree.scala
```diff
@@ -38,33 +38,73 @@ package com.twitter.algebird
...
-class QTreeSemigroup[A: Monoid](k: Int) extends Semigroup[QTree[A]] {
+class QTreeSemigroup[A](k: Int)(implicit val underlyingMonoid: Monoid[A]) extends Semigroup[QTree[A]] {
+  /** Override this if you want to change how frequently sumOption calls compress */
+  def compressBatchSize: Int = 25
   def plus(left: QTree[A], right: QTree[A]) = left.merge(right).compress(k)
+  override def sumOption(items: TraversableOnce[QTree[A]]): Option[QTree[A]] = if (items.isEmpty) None
+  else {
+    // only call compressBatchSize once
+    val batchSize = compressBatchSize
+    var count = 1 // start at 1, so we only compress after batchSize items
```
jnievelt: instead highlight that we're pulling one off the iterator before starting the loop

johnynek: not sure what you mean here, @jnievelt . The count starting at 1 is so (count % batchSize == 0) after batchSize loops. If we started at 0, we would compress the first time as the code was written.

Did you want something else?

jnievelt: That's not the case though.  If batchSize = 2, that condition will hold on the first pass through the loop, not the second.

It's a little weird to think about.  Here, the first compression will be on an uncompressed combination of 25 items.  Other compressions will be on 26 (including the previous compression), until the last one.  Initializing count to 1 if it avoids doing a compression on the _first_ iteration makes sense to me.  But 25 vs. 26 for the first compression only seems inconsequential.

johnynek:
Related to #429

PR #434, opened 2015-04-21
Moments returns NaN when count is too low for higher order statistics

wlandecker:
Typically, we require at least N samples in order to compute the N'th order moment. Currently, Moments returns values even when fewer than N samples -- these values are not meaningful, and should not be relied upon for very small sample size. This PR returns NaN in these cases, instead.

### DIFF
--- a/algebird-core/src/main/scala/com/twitter/algebird/MomentsGroup.scala
+++ b/algebird-core/src/main/scala/com/twitter/algebird/MomentsGroup.scala
@@ -29,14 +29,23 @@ case class Moments(m0: Long, m1: Double, m2: Double, m3: Double, m4: Double) {
   def mean = m1
 
   // Population variance, not sample variance.
-  def variance = m2 / count
+  def variance = if (count > 1)
+    m2 / count
+  else /* don't return junk when the moment is not defined */
+    Double.NaN
 
   // Population standard deviation, not sample standard deviation.
   def stddev = math.sqrt(variance)
 
-  def skewness = math.sqrt(count) * m3 / math.pow(m2, 1.5)
+  def skewness = if (count > 2)
+    math.sqrt(count) * m3 / math.pow(m2, 1.5)
+  else /* don't return junk when the moment is not defined */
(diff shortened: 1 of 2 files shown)


algebird-core/src/main/scala/com/twitter/algebird/MomentsGroup.scala
```diff
@@ -29,14 +29,23 @@ case class Moments(m0: Long, m1: Double, m2: Double, m3: Double, m4: Double) {
   def mean = m1
 
   // Population variance, not sample variance.
-  def variance = m2 / count
+  def variance = if (count > 1)
```
johnynek: what if we do this: add `varianceUnsafe: Double` (and similar) that don't check, and then make `def variance = if (count > 1) varianceUnsafe else Double.NaN`

Is that of any value?

wlandecker: If folks have some need for the statistics in the case of a very low count, then adding `varianceUnsafe` (etc.) makes sense. (But when would these numbers be of value? By definition, the statistics aren't defined in these edge cases.)

On the other hand, if the purpose for `varianceUnsafe` is for backward compatibility, then maybe it would be better to instead dump my changes into a `varianceSafe` (and similar), and keep the current definitions as they already were. That is to say, the existing definitions are already the unsafe versions (since new algorithms might change their values when the count is very low), and the pull request is creating the safe versions.

Other reasons for making that change that I missed?

johnynek:
So we already have this:

```scala
scala> Monoid.sum(List(Moments(1))).kurtosis
res6: Double = NaN

scala> Monoid.sum(List(Moments(1))).skewness
res7: Double = NaN
scala> Monoid.sum(List.empty[Moments]).skewness
res8: Double = NaN

scala> Monoid.sum(List.empty[Moments]).kurtosis
res9: Double = NaN
```
So, I think we are already giving NaNs in a few cases, just not up to the correct sizes.

Merging.

PR #442, opened 2015-05-08
add ! to ApproximateBoolean

smarden1:
Add unary_! to ApproximateBoolean so you can !ApproximateBoolean in addition to ApproximateBoolean.not


johnynek:
Thanks!

Issue #436, opened 2015-05-05
Add QTreeAggregators


johnynek:
give a simple way to get percentiles using an Aggregator.

johnynek:
This would just be something like:

1) create a QTreeSemigroup of the right size (with Unit value).
2) andThenPresent to get out the percentiles you want.

PR #443, opened 2015-05-09
add QTreeAggregator and add approximatePercentileBounds to Aggregator

smarden1:
attempt at issue #436

### DIFF
--- a/algebird-core/src/main/scala/com/twitter/algebird/QTree.scala
+++ b/algebird-core/src/main/scala/com/twitter/algebird/QTree.scala
@@ -367,3 +367,9 @@ case class QTree[A](
     (n.toDouble(ll) / luc, n.toDouble(uu) / ulc)
   }
 }
+
+case class QTreeAggregator(percentile: Double, k: Int) extends Aggregator[Double, QTree[Double], (Double, Double)] {
+  def prepare(input: Double) = QTree(input)
+  def semigroup = new QTreeSemigroup[Double](2)
+  def present(qt: QTree[Double]) = qt.quantileBounds(percentile)
+}
--- a/algebird-core/src/main/scala/com/twitter/algebird/QTree.scala
+++ b/algebird-core/src/main/scala/com/twitter/algebird/QTree.scala
@@ -370,6 +370,6 @@ case class QTree[A](
 
 case class QTreeAggregator(percentile: Double, k: Int) extends Aggregator[Double, QTree[Double], (Double, Double)] {
   def prepare(input: Double) = QTree(input)
-  def semigroup = new QTreeSemigroup[Double](2)
+  def semigroup = new QTreeSemigroup[Double](k)
   def present(qt: QTree[Double]) = qt.quantileBounds(percentile)
 }
--- a/algebird-core/src/main/scala/com/twitter/algebird/QTree.scala
+++ b/algebird-core/src/main/scala/com/twitter/algebird/QTree.scala
@@ -382,6 +382,11 @@ object QTreeAggregator {
   val DefaultK = 9
 }
 
+/**
+ * QTree aggregator is an aggregator that can be used to find the approximate percentile bounds.
+ * The items that are iterated over to produce this approximation cannot be negative.
+ * Returns an Intersection which represents the bounded approximation.
+ */
 case class QTreeAggregator[T](percentile: Double, k: Int = QTreeAggregator.DefaultK)(implicit val num: Numeric[T])
   extends Aggregator[T, QTree[Unit], Intersection[InclusiveLower, InclusiveUpper, Double]]
   with QTreeAggregatorLike[T] {
@@ -392,6 +397,12 @@ case class QTreeAggregator[T](percentile: Double, k: Int = QTreeAggregator.Defau
   }
 }
 
+/**
+ * QTreeAggregatorLowerBound is an aggregator that is used to find an appoximate percentile.
+ * This is similar to a QTreeAggregator, but is a convenience because instead of returning an Intersection,
+ * it instead returns the lower bound of the percentile.
+ * Like a QTreeAggregator, the items that are iterated over to produce this approximation cannot be negative.
+ */
 case class QTreeAggregatorLowerBound[T](percentile: Double, k: Int = QTreeAggregator.DefaultK)(implicit val num: Numeric[T])
   extends Aggregator[T, QTree[Unit], Double]
   with QTreeAggregatorLike[T] {
--- a/algebird-core/src/main/scala/com/twitter/algebird/QTree.scala
+++ b/algebird-core/src/main/scala/com/twitter/algebird/QTree.scala
@@ -194,6 +194,8 @@ case class QTree[A](
    * an estimate of the median.
    */
   def quantileBounds(p: Double): (Double, Double) = {
+    require(p >= 0.0 && p < 1.0, "The given percentile must be of the form 0 <= p < 1.0")
+
     val rank = math.floor(count * p).toLong
     // get is safe below, because findRankLowerBound only returns
     // None if rank > count, but due to construction rank <= count
--- a/algebird-core/src/main/scala/com/twitter/algebird/QTree.scala
+++ b/algebird-core/src/main/scala/com/twitter/algebird/QTree.scala
@@ -374,7 +374,6 @@ trait QTreeAggregatorLike[T] {
   val percentile: Double
   val k: Int
   implicit val num: Numeric[T]
-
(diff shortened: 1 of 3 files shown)


algebird-core/src/main/scala/com/twitter/algebird/QTree.scala
```diff
@@ -367,3 +367,9 @@ case class QTree[A](
     (n.toDouble(ll) / luc, n.toDouble(uu) / ulc)
   }
 }
+
+case class QTreeAggregator(percentile: Double, k: Int) extends Aggregator[Double, QTree[Double], (Double, Double)] {
```
smarden1: are there good defaults for k?

avibryant: the tree will have at most `2^k` nodes. So, like with HLL, the reasonable range is somewhere from 7...12 or so. 8 or maybe 9 seems like a decent default to me.

algebird-core/src/main/scala/com/twitter/algebird/QTree.scala
```diff
@@ -367,3 +367,9 @@ case class QTree[A](
...
+  def prepare(input: Double) = QTree(input)
+  def semigroup = new QTreeSemigroup[Double](k)
+  def present(qt: QTree[Double]) = qt.quantileBounds(percentile)
```
avibryant: Should there be a version (maybe even the default version) that just picks the lower or upper bound here?

algebird-core/src/main/scala/com/twitter/algebird/QTree.scala
avibryant: How about making this generic for any `N:Numeric` instead of requiring `Double`?

algebird-core/src/main/scala/com/twitter/algebird/QTree.scala
```diff
@@ -367,3 +367,9 @@ case class QTree[A](
     (n.toDouble(ll) / luc, n.toDouble(uu) / ulc)
   }
 }
+
+case class QTreeAggregator(percentile: Double, k: Int) extends Aggregator[Double, QTree[Double], (Double, Double)] {
+  def prepare(input: Double) = QTree(input)
+  def semigroup = new QTreeSemigroup[Double](k)
```
avibryant: use `val`?

algebird-core/src/main/scala/com/twitter/algebird/QTree.scala
```diff
@@ -367,3 +369,29 @@ case class QTree[A](
     (n.toDouble(ll) / luc, n.toDouble(uu) / ulc)
   }
 }
+
+trait QTreeAggregatorLike[T] {
+  val percentile: Double
+  val k: Int
+  implicit val num: Numeric[T]
+  def prepare(input: T) = QTree(num.toDouble(input))
+  val semigroup = new QTreeSemigroup[Double](k)
```
johnynek: why don't we want `QTreeSemigroup[Unit]` here? That will be less memory and won't hurt our ability to do percentiles.

algebird-core/src/main/scala/com/twitter/algebird/QTree.scala
```diff
@@ -367,3 +369,29 @@ case class QTree[A](
...
+}
+
+object QTreeAggregator {
+  val DefaultK = 9
```
johnynek: that's pretty high, is it not?

algebird-test/src/test/scala/com/twitter/algebird/QTreeTest.scala
```diff
@@ -59,7 +59,7 @@ class QTreeTest extends WordSpec with Matchers {
     list.map{ QTree(_) }.reduce{ qtSemigroup.plus(_, _) }
   }
 
-  def trueQuantile(list: Seq[Double], q: Double) = {
+  def trueQuantile[T: Numeric](list: Seq[T], q: Double): T = {
```
johnynek: we just need `T: Ordering` here, right?

algebird-test/src/test/scala/com/twitter/algebird/QTreeTest.scala
```diff
@@ -69,6 +69,15 @@ class QTreeTest extends WordSpec with Matchers {
     list.filter{ _ >= from }.filter{ _ < to }.sum
   }
 
+  def aggregate[T](list: Seq[T], quantile: Double)(implicit num: Numeric[T]): (Double, Double) = {
+    val agg = QTreeAggregator(quantile)
+    val r = list
+      .map(agg.prepare(_))
+      .reduce(agg.semigroup.plus(_, _))
+
+    agg.present(r)
```
johnynek: this is all the same as `agg(list)` isn't it? Why the manual reimplementation?

algebird-test/src/test/scala/com/twitter/algebird/QTreeTest.scala
```diff
@@ -69,6 +69,15 @@ class QTreeTest extends WordSpec with Matchers {
     list.filter{ _ >= from }.filter{ _ < to }.sum
```
johnynek: we don't need the outer `{ }`, and this could be `trueRangeSum[T: Monoid](l: TraversableOnce[T], int: Interval[T]): T` to be more general

algebird-core/src/main/scala/com/twitter/algebird/QTree.scala
```diff
@@ -367,3 +369,29 @@ case class QTree[A](
...
+}
+
+case class QTreeAggregator[T](percentile: Double, k: Int = QTreeAggregator.DefaultK)(implicit val num: Numeric[T])
+  extends Aggregator[T, QTree[Double], (Double, Double)]
+  with QTreeAggregatorLike[T] {
+
+  def present(qt: QTree[Double]) = qt.quantileBounds(percentile)
+}
+
+case class QTreeAggregatorLowerBound[T](percentile: Double, k: Int = QTreeAggregator.DefaultK)(implicit val num: Numeric[T])
```
johnynek: scaladoc comment please.

algebird-core/src/main/scala/com/twitter/algebird/QTree.scala
```diff
@@ -367,3 +369,29 @@ case class QTree[A](
...
+  implicit val num: Numeric[T]
+  def prepare(input: T) = QTree(num.toDouble(input))
+  val semigroup = new QTreeSemigroup[Double](k)
+}
+
+object QTreeAggregator {
+  val DefaultK = 9
+}
+
+case class QTreeAggregator[T](percentile: Double, k: Int = QTreeAggregator.DefaultK)(implicit val num: Numeric[T])
```
johnynek: scala doc comment please.

algebird-core/src/main/scala/com/twitter/algebird/QTree.scala
```diff
@@ -367,3 +369,29 @@ case class QTree[A](
...
+  extends Aggregator[T, QTree[Double], (Double, Double)]
```
johnynek: can we use `Interval` here instead of `Tuple2`?

johnynek: I think this should be `QTree[Unit]`

algebird-core/src/main/scala/com/twitter/algebird/QTree.scala
smarden1: this was what avi suggested in a previous comment. Happy to change to whatever.

algebird-core/src/main/scala/com/twitter/algebird/QTree.scala
smarden1: will fix.

algebird-test/src/test/scala/com/twitter/algebird/QTreeTest.scala
smarden1: this line wasn't touched in my diff, nor was it called... I can change it if you want but it seems out of scope

algebird-test/src/test/scala/com/twitter/algebird/QTreeTest.scala
smarden1: didn't realize... will fix.

algebird-test/src/test/scala/com/twitter/algebird/QTreeTest.scala
johnynek: Ignore it if you like. I just saw it above the diff.

johnynek:
This is great! Thanks!

PR #447, opened 2015-05-26
QTree quantileBounds assert percentile <= 1.0

egonina:
Assertion fails if percentile == 1.0, should pass if 0 <= percentile <= 1.0


ianoc:
Was this a regression from an old version? new require? could we get a short test that just creates one of these with 1 so we don't regress easily again

egonina:
This was a regression, yes, several of our internal tests failed. This code is tested here https://github.com/twitter/algebird/blob/develop/algebird-test/src/test/scala/com/twitter/algebird/QTreeTest.scala#L77

egonina:
added a test to test the boundary values

PR #448, opened 2015-06-02
Pzheng/gaussian euclidean

perryzheng:
Add gaussian distribution monoid


algebird-core/src/main/scala/com/twitter/algebird/statistics/GaussianDistributionMonoid.scala
```diff
@@ -0,0 +1,28 @@
+package com.twitter.algebird.statistics
+
+import com.twitter.algebird.Monoid
+
+/**
+ * @mean - Mean
+ * @sigma - Standard deviation
+ * aka Normal distribution
+ */
+case class GaussianDistribution(mean: Double, sigma: Double)
```
johnynek: might be nice to add a method like:
```scala
def sample(r: java.util.Random): Double
```
to get a sample from this distribution.

algebird-core/src/main/scala/com/twitter/algebird/statistics/GaussianDistributionMonoid.scala
```diff
@@ -0,0 +1,28 @@
...
+  implicit def monoid: Monoid[GaussianDistribution] = new GaussianDistributionMonoid
+}
+
+/**
+ * This monoid stems from the fact that if X and Y are independent random variables
+ * that are normally distributed, then their sum is also
+ * normally distributed, with its new mean equal to the sum of two means
+ * and variance equal to the sum of two variances.
+ */
+class GaussianDistributionMonoid extends Monoid[GaussianDistribution] {
```
johnynek: this can be an `object`

algebird-core/src/main/scala/com/twitter/algebird/statistics/GaussianDistributionMonoid.scala
```diff
@@ -0,0 +1,28 @@
...
+  override def zero = new GaussianDistribution(0, 0)
+
+  override def plus(left: GaussianDistribution, right: GaussianDistribution): GaussianDistribution = {
```
johnynek: no need for `{ }` please avoid when not needed.

algebird-test/src/test/scala/com/twitter/algebird/statistics/GaussianDistributionMonoidTests.scala
```diff
@@ -0,0 +1,23 @@
...
+  implicit val gaussianGenerators = Arbitrary {
+    for (
+      mean <- Gen.choose(0, 10000);
+      sigma <- Gen.choose(0, 10000)
+    ) yield (GaussianDistribution(mean, sigma))
+  }
+
+  property("GaussianDistributionMonoid is a Monoid") {
+    monoidLawsEq[GaussianDistribution]{
+      (a, b) => a.mean == b.mean && a.sigma == b.sigma
```
johnynek: we should not need `monoidLawsEq`, I think `monoidLaws` will do.

algebird-core/src/main/scala/com/twitter/algebird/statistics/GaussianDistributionMonoid.scala
```diff
@@ -0,0 +1,28 @@
...
+    new GaussianDistribution(left.mean + right.mean, left.sigma + right.sigma)
+  }
```
johnynek: please add an optimized sumOption:

```scala
def sumOption(its: TraversableOnce[GaussionDistribution]): Option[GaussionDistribution] =
  if (its.isEmpty) None
  else {
    var mean = 0.0
    var sigma = 0.0
    val it = its.toIterator
    while (it.hasNext) {
      val g = it.next
      mean += g.mean
      sigma += g.sigma
    }
    GuassianDistribution(mean, sigma)
  }
```
The point is that summing on a large list can avoid a lot of allocation using .sum or .sumOption, which this method will enable. It uses `vars` but only locally, and still gives the user a nice, immutable functional API.

algebird-core/src/main/scala/com/twitter/algebird/statistics/GaussianDistributionMonoid.scala
```diff
@@ -0,0 +1,28 @@
...
+
+/**
+ * @mean - Mean
+ * @sigma - Standard deviation
+ * aka Normal distribution
+ */
+case class GaussianDistribution(mean: Double, sigma: Double)
+
+object GaussianDistribution {
+  implicit def monoid: Monoid[GaussianDistribution] = new GaussianDistributionMonoid
```
johnynek: this can be an `implicit val` to not allocate at every use (since this Monoid is a constant, and has no inputs).

johnynek:
Thanks for doing this! Just some minor cleanups requested.

algebird-core/src/main/scala/com/twitter/algebird/statistics/GaussianDistributionMonoid.scala
avibryant: Or possibly take it further and integrate with https://github.com/jliszka/probability-monad

johnynek: I actually have some code internal to Twitter that has this as well as some bootstrap and chernov/hoeffding-bound based randomized testing. I should release that (and perhaps make it work with the the code above rather than mine which is just a Reader monad on java.util.Random).

algebird-core/src/main/scala/com/twitter/algebird/statistics/GaussianDistributionMonoid.scala
perryzheng: done

algebird-core/src/main/scala/com/twitter/algebird/statistics/GaussianDistributionMonoid.scala
perryzheng: great!

algebird-test/src/test/scala/com/twitter/algebird/statistics/GaussianDistributionMonoidTests.scala
perryzheng: done

algebird-core/src/main/scala/com/twitter/algebird/statistics/GaussianDistributionMonoid.scala
perryzheng: done

algebird-core/src/main/scala/com/twitter/algebird/statistics/GaussianDistributionMonoid.scala
perryzheng: yep

algebird-core/src/main/scala/com/twitter/algebird/statistics/GaussianDistributionMonoid.scala
```diff
@@ -0,0 +1,45 @@
...
+ * @mean - Mean
+ * @sigma - Standard deviation
+ * aka Normal distribution
+ */
+case class GaussianDistribution(mean: Double, sigma: Double)
+
+object GaussianDistribution {
+  implicit val monoid: Monoid[GaussianDistribution] = GaussianDistributionMonoid
+
+  def sample(r: java.util.Random): Double =
```
johnynek: actually, this should be a method on the case class, and it should be scaled:
```scala
mean + r.nextGaussian() * math.sqrt(stddev)
```

algebird-core/src/main/scala/com/twitter/algebird/statistics/GaussianDistributionMonoid.scala
```diff
@@ -0,0 +1,45 @@
+package com.twitter.algebird.statistics
+
+import com.twitter.algebird.Monoid
+
+/**
+ * @mean - Mean
+ * @sigma - Standard deviation
```
johnynek: wait, this is not correct. It needs to be `sigma2` being the variance:

http://en.wikipedia.org/wiki/Sum_of_normally_distributed_random_variables

we can have a method to get the `stddev` to use the same name as: https://github.com/twitter/algebird/blob/develop/algebird-core/src/main/scala/com/twitter/algebird/MomentsGroup.scala
`def stddev = math.sqrt(sigma2)`

algebird-test/src/test/scala/com/twitter/algebird/statistics/GaussianDistributionMonoidTests.scala
```diff
@@ -0,0 +1,21 @@
+package com.twitter.algebird.statistics
+
+import com.twitter.algebird.CheckProperties
+import org.scalacheck.{ Arbitrary, Gen }
+import org.scalatest._
+
+class GaussianDistributionMonoidTests extends CheckProperties {
+  import com.twitter.algebird.BaseProperties._
+
+  implicit val gaussianMonoid = GaussianDistributionMonoid
```
johnynek: shouldn't need this. Since it is implicit in the companion object, scala should find it.

algebird-core/src/main/scala/com/twitter/algebird/statistics/GaussianDistributionMonoid.scala
perryzheng: oh gotcha!! that's right :(

johnynek:
Thank you!

PR #449, opened 2015-06-09
Implement lsh

perryzheng:


algebird-core/src/main/scala/com/twitter/algebird/lsh/hashing/EuclideanHashFamily.scala
```diff
@@ -0,0 +1,45 @@
...
+
+class EuclideanHasher(
+  hashTableId: Int,
+  hashFunctionId: Int,
+  radius: Double = 4,
+  dimension: Int)
+  extends Hasher with Serializable {
+  val rand = new Random(hashTableId * 10000 + hashFunctionId)
+
+  val offset = rand.nextInt(radius.toInt)
```
jnievelt: Seems like this should be rand.nextDouble * radius ?

algebird-core/src/main/scala/com/twitter/algebird/lsh/vector/VectorMath.scala
```diff
@@ -0,0 +1,87 @@
+package com.twitter.algebird.lsh.vector
+
+/**
+ * Vector math functions for LshVectors.
+ */
+object VectorMath {
```
jnievelt: Seems like this is too much.  We only use dot here, and that could be put somewhere else (EuclideanHashFamily helper object, etc.) if we really need to add it.

algebird-core/src/main/scala/com/twitter/algebird/lsh/hashing/HashFamily.scala
```diff
@@ -0,0 +1,14 @@
+package com.twitter.algebird.lsh.hashing
+
+import com.twitter.algebird.lsh.vector.BaseLshVector
+
+trait Hasher {
```
jnievelt: This abstraction doesn't seem right.  LSH isn't confined only to vectors of doubles (that's specific to Euclidean space).  Also Hasher as a mapping of something onto an Int isn't specific to LSH.  Likely we'd want to have `Hasher[-T] { def hash(t: T): Int }` at the algebird (or algebird.hashing) level.

algebird-core/src/main/scala/com/twitter/algebird/lsh/vector/LshVector.scala
```diff
@@ -0,0 +1,37 @@
+package com.twitter.algebird.lsh.vector
+
+import com.twitter.algebird.{ ArrayMonoid, Monoid }
+import java.util
+
+/**
+ * An LshVector must permit the operations listed below.
+ */
+trait BaseLshVector {
```
jnievelt: Why is this trait needed?  And why would it override hashCode?

algebird-core/src/main/scala/com/twitter/algebird/lsh/hashing/HashFamily.scala
```diff
@@ -0,0 +1,14 @@
...
+  def hash(vector: Array[Double]): Int
+}
+
+trait HashFamily {
+  val familyId = -1 // Should be unique to each Family.
+  def createHasher(hashTableId: Int, hashFunctionId: Int): Hasher
+  def combine(hashes: Array[Int]): Int
```
jnievelt: Why is this needed?

jnievelt:
Overall seems like a lot of code where we could probably get away with just EuclideanHasher.  I'd also consider dropping the 'hashing' in the package and just leave it 'lsh' (the h already stands for hash).

algebird-core/src/main/scala/com/twitter/algebird/lsh/hashing/HashFamily.scala
johnynek: Actually there is a scala.util.hashing.Hashing: http://www.scala-lang.org/api/2.11.4/index.html#scala.util.hashing.Hashing

Can we use that trait?

algebird-core/src/main/scala/com/twitter/algebird/lsh/vector/VectorMath.scala
johnynek: Also note we have VectorSpace: https://github.com/twitter/algebird/blob/develop/algebird-core/src/main/scala/com/twitter/algebird/VectorSpace.scala

Seems you want to add InnerProductSpace:
http://en.wikipedia.org/wiki/Inner_product_space
```scala
trait InnerProductSpace[F, C[_]] extends VectorSpace[F, C] {
  def innerProduct(left: C[F], right: C[F]): F
}
```
Then you can implement such a space for `InnerProductSpace[Double, Array]` or something `InnerProductSpace[Double, Vector]`.

algebird-core/src/main/scala/com/twitter/algebird/lsh/vector/VectorMath.scala
```diff
@@ -0,0 +1,87 @@
...
+    var sum = 0.0
+    var idx = 0
+    while (idx < vec.size) {
+      sum += vec(idx) * vec(idx)
+      idx += 1
+    }
+    math.sqrt(sum)
+  }
+
+  def normalize[U <: LshVector](vector: U): LshVector = divide(vector, magnitude(vector))
```
johnynek: this should all be in InnerProductSpace when `F =:= Double`.

algebird-core/src/main/scala/com/twitter/algebird/lsh/hashing/EuclideanHashFamily.scala
```diff
@@ -0,0 +1,45 @@
...
+  val randomProjection = Array.ofDim[Double](dimension)
+  for (i <- 0 until dimension)
+    randomProjection.update(i, rand.nextGaussian)
+
+  println("Euclidean Hasher: tableId: %d, hashFunctionId: %d, offset: %d, projection: %s",
```
johnynek: should not println.

ianoc:

PR #452, opened 2015-06-23
Add MapAggregator from 1 (key, aggregator) pair

vidma:
- Make API easier to use (more consistent), by allowing to create a MapAggregator of one "metric"

improves: https://github.com/twitter/algebird/pull/411#issuecomment-104117951

### DIFF
--- a/project/GenTupleAggregators.scala
+++ b/project/GenTupleAggregators.scala
@@ -66,8 +66,17 @@ object MapAggregator {
   }
 
   def genMapMethods(max: Int, isMonoid: Boolean = false): String = {
+    val aggType = if (isMonoid) "Monoid" else ""
+
+    // there's no Semigroup[Tuple1[T]], so just use T as intermediary type instead of Tuple1[T]
+    val aggregatorForOneItem = s"""
+       |def apply[K, A, B, C](aggDef: (K, ${aggType}Aggregator[A, B, C])): ${aggType}Aggregator[A, B, Map[K, C]] = {
+       |  val (key, agg) = aggDef
+       |  agg.andThenPresent(value => Map(key -> value))
+       |}
+    """.stripMargin
+
     (2 to max).map(i => {
-      val aggType = if (isMonoid) "Monoid" else ""
       val nums = (1 to i)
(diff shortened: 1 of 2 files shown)


project/GenTupleAggregators.scala
```diff
@@ -66,8 +66,17 @@ object MapAggregator {
   }
 
   def genMapMethods(max: Int, isMonoid: Boolean = false): String = {
+    val aggType = if (isMonoid) "Monoid" else ""
+
+    // there's no Semigroup[Tuple1[T]], so just use T as intermediary type instead of Tuple1[T]
+    val aggregatorForOneItem = s"""
+       |def apply[A, B, C](aggDef: (String, ${aggType}Aggregator[A, B, C])): ${aggType}Aggregator[A, B, Map[String, C]] = {
```
johnynek: can this be `(K, ${aggType}Aggregator[A, B, C])` so it matches the others? We should not require a String as the key, I don't think.

project/GenTupleAggregators.scala
```diff
@@ -66,8 +66,17 @@ object MapAggregator {
...
+       |  val (key, agg) = aggDef
+       |  agg.andThenPresent(value => Map(key -> value))
+       |}
+    """.stripMargin
+
     (2 to max).map(i => {
```
johnynek: can't we just change the `2` to `1` and get the same result?

johnynek: I guess that would cause needless boxing in the middle. Your original approach is better in that regard.

vidma: it didn't seem as simple as changing `2` to `1`...
with all these `%s` formatting, it's too hard infer anything about what the code-gen below does. so I'd leave it as is, at least until the original `2-max` part is refactored to use string interpolation (at least).

project/GenTupleAggregators.scala
vidma: :+1: sure!

johnynek:
looks like this is hung on coveralls, trying to fix this issue here: #458

PR #455, opened 2015-06-25
Remove unnecessary use of scala.math

sid-kap:
Change `scala.math.foo` to `math.foo`


johnynek:
generally we don't have a strong stance on this, but I'll merge this one.

PR #456, opened 2015-06-25
Don't call deprecated HyperLogLog methods in tests

sid-kap:
Replace instances of `new HyperLogLogMonoid(...).apply(...)`
with `new HyperLogLogMonoid(...).create(...)` in `HyperLogLogTest`.


sid-kap:
The only failure in the build was due to a randomized data structure.
```
[info] MinHasherSpec:
[info] MinHasher32
[info] - should measure 0.5 similarity in 1024 bytes with < 0.1 error
[info] - should measure 0.8 similarity in 1024 bytes with < 0.05 error *** FAILED ***
[info]   0.054166666666666696 was not less than 0.05 (MinHasherTest.scala:30)
[info] - should measure 1.0 similarity in 1024 bytes with < 0.01 error
```

johnynek:
this is flakey test. We need a solution to this problem.

Issue #446, opened 2015-05-21
add sumOption to GeneratedProductMacros


johnynek:
See: https://github.com/twitter/algebird/blob/develop/algebird-core/src/main/scala/com/twitter/algebird/GeneratedProductAlgebra.scala

Issues:

1) sumOption is not implemented.
2) Monoid does not extend the Semigroup (and on down), so it also does not have a fast sum.

Need to follow more closely the approach of Tuple:
https://github.com/twitter/algebird/blob/develop/algebird-core/src/main/scala/com/twitter/algebird/GeneratedAbstractAlgebra.scala

Requirements: make it binary compatible.

sid-kap:
I took care of (2) in PR #457

sritchie:
Closed by #588

PR #457, opened 2015-06-26
Update product_generators.rb

sid-kap:
Now, ProductNMonoid extends ProductNSemigroup,
     ProductNGroup extends ProductNMonoid,
     ProductNRing extends ProductNGroup.

Took advantage of inheritance by not declaring plus in semigroup, group,
and ring, not declaring zero in group and ring, etc.


sid-kap:
This change handles issue #446

johnynek:
looks good. I think our coveralls is buggy right now, so I'll merge this.

Thanks for the help!

PR #458, opened 2015-06-26
Update sbt-scoverage

johnynek:
This may be working now with more recent versions of the plugins. The report seems to work on my machine (98% coverage for algebird-core, if correct).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,5 +6,7 @@ matrix:
       script: ./sbt ++$TRAVIS_SCALA_VERSION clean test
 
     - scala: 2.11.5
-      script: ./sbt ++$TRAVIS_SCALA_VERSION clean test
-      after_success: "./sbt coveralls"
+      script:
+        - ./sbt ++$TRAVIS_SCALA_VERSION clean coverage test &&
+          ./sbt ++$TRAVIS_SCALA_VERSION coverageAggregate
+      after_success: "./sbt ++$TRAVIS_SCALA_VERSION coveralls"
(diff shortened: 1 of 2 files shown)


travisbrown:
I'm pretty sure sbt-scoverage 1.1.0 won't work nicely with Coveralls until sbt-coveralls 1.1.0 comes out (see my comment [here](https://github.com/scoverage/scalac-scoverage-plugin/issues/104#issuecomment-99160357))—we're had to stick with sbt-scoverage 1.0.4 and sbt-coveralls 1.0.0 in other projects. I guess it's worth a try here, though.

johnynek:
any idea how to disable the build staying yellow waiting for coveralls if we can't get it to run? It's really annoying.

It's sad how broken all this stuff is. I hate to dump on open source projects, but typesafe or even Twitter should pay if needed to make this work. This has yet to give us good data for any of our projects when run with Travis, as far as I know. Have you been able to get the finagle stack working with it?

`<insert envy that the analytics stack doesn't get as much of your time as finagle does>`

travisbrown:
I've got a review out right now adding Coveralls reporting to six Finagle-related projects, and that part was relatively painless—see e.g.:

https://github.com/travisbrown/util
https://github.com/travisbrown/scrooge
https://github.com/travisbrown/ostrich

We [switched Finch from Coveralls to Codecov](https://codecov.io/github/finagle/finch) as an experiment and it's definitely been vastly less fussy than Coveralls. Scoverage itself is I think fairly solid, and Codecov works directly with its reports (there's no separate sbt-codecov plugin—you just add `bash <(curl -s https://codecov.io/bash)` to `after_success`).

ianoc:

DanielleSucher:
Ping on this! Is it worth considering switching to Codecov here too?

johnynek:
@DanielleSucher I think that would be awesome.

johnynek:
let's just focus on #414

PR #462, opened 2015-07-11
Refactor MapAggregator

vidma:
- use string interpolation instead of .format()
- inline `prepare`, `present`, `semigroup`
- use more clear variable names:
  - rename i to aggrCount
  - rename nums to aggrNums
  - rename semiType to semigroup (it refers to field name, not type), add semigroupType which actually refers to type

@johnynek

### DIFF
--- a/project/GenTupleAggregators.scala
+++ b/project/GenTupleAggregators.scala
@@ -23,13 +23,25 @@ object MultiAggregator {
       genMethods(22, "def", Some("apply"), true) + "\n" + "}")
 
     val mapAggPlace = dir / "com" / "twitter" / "algebird" / "MapAggregator.scala"
-    IO.write(mapAggPlace,
-"""package com.twitter.algebird
-
-object MapAggregator {
-""" +
-      genMapMethods(22) + "\n" +
-      genMapMethods(22, true) + "\n" + "}")
+    IO.write(
+      mapAggPlace,
+      s"""
+      |package com.twitter.algebird
+      |
+      |trait MapAggregator[A, B, K, C] extends Aggregator[A, B, Map[K, C]] {
+      |  def keys: Set[K]
+      |}
+      |
+      |trait MapMonoidAggregator[A, B, K, C] extends MonoidAggregator[A, B, Map[K, C]] {
+      |  def keys: Set[K]
+      |}
+      |
+      |object MapAggregator {
+      |  ${genMapMethods(22)}
+      |  ${genMapMethods(22, isMonoid = true)}
+      |}
+      """.stripMargin
+    )
 
     Seq(tupleAggPlace, multiAggPlace, mapAggPlace)
   }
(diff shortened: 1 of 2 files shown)


vidma:
P.S. we found it quite weird that currently having an aggregator created by `MapAggregator` one can't get the keys out of it:
```scala
val aggr = MapAggregator(
  "sum" -> Aggregator.sum,
  "min" -> Aggregator.min
)
// can't do this yet
val keys = aggr.keys
```
Looks like this would need to introduce a `class MapAggregator extends Aggregator` , and then define a `keys` method (separate PR).

project/GenTupleAggregators.scala
```diff
@@ -76,29 +76,32 @@ object MapAggregator {
...
+      val semigroupType = s"Tuple${aggrCount}${semigroup.capitalize}"
+
+      val bs = aggrNums.map("B" + _).mkString(", ")
+      val tupleBs = s"Tuple${aggrCount}[$bs]"
+
+      s"""
+      |def apply[K, A, $bs, C]($inputAggs): ${aggType}Aggregator[A, $tupleBs, Map[K, C]] = {
+      |  new ${aggType}Aggregator[A, $tupleBs, Map[K, C]] {
+      |    def prepare(a: A) = (
+      |      ${aggrNums.map(i => s"agg$i._2.prepare(a)").mkString(", ")}
```
ianoc: $i._2 is ambigious... well its not, but it doesn't operate like it reads. can we use ${i} here?

project/GenTupleAggregators.scala
```diff
@@ -76,29 +76,32 @@ object MapAggregator {
...
-    def present(b: %s) = Map(%s)
-  }
-}""".format(bs, aggs, aggType, tupleBs,
-            aggType, tupleBs,
-            prepares,
-            semiType, semigroup,
-            tupleBs, present)
+      val inputAggs = aggrNums.map(i => s"agg$i: (K, ${aggType}Aggregator[A, B$i, C])").mkString(", ")
+
+      val semigroup = if (isMonoid) "monoid" else "semigroup"
```
ianoc: should maybe change this val to aggType or something more indicative. Semigroup is an odd name for something that could be monoid.

project/GenTupleAggregators.scala
vidma: :+1:

project/GenTupleAggregators.scala
vidma: a monoid extends semigroup, and this val is used as variable name, so not sure

johnynek:
+

project/GenTupleAggregators.scala
vidma: I'd prefer not to use word `aggregator` to avoid confusion with an `*Aggregator`.
so either `semigroupType/semigroupField` or a `summer`, other suggestions?

johnynek:
much cleaner code-gen! Thanks.

Issue #459, opened 2015-07-10
CMSItem disregards count in ++


johnynek:
https://github.com/twitter/algebird/blob/develop/algebird-core/src/main/scala/com/twitter/algebird/CountMinSketch.scala#L449

note that if you do `CMSZero + ("a", 100) + ("a", 1)`, I think you'll get the same as `CMSZero + ("a", 1) + ("a", 1)`

reconditesea:
@johnynek What's the definition of a CMSItem? 
If it's one single key with one element, I think there is no bug. If it's one single key with multiple counts, then I think both the frequency and innerProduct also need to be updated.

jnievelt:
That's not the definition.  It's a bit weird because `CMSMonoid#create` doesn't take a count parameter (it's implicitly 1), but `CMSZero#+` does.  Basically there's not a way to hit this without involving `CMSZero`.

While we're at it, let's fix `frequency`:

https://github.com/twitter/algebird/blob/develop/algebird-core/src/main/scala/com/twitter/algebird/CountMinSketch.scala#L454

reconditesea:
Yeah, both frequency and innerProduct are updated.

reconditesea:
I believe this is closed by #464 .

PR #464, opened 2015-07-15
Create a sparse Count-Min-Sketch.

### LINKED ISSUE #461: SparseCMS
Might be good to have a SparseCMS that keeps a map of values until the size of the map reaches `depth * width`. This could improve accuracy for cases of small data (which happens when we use CMS as a value with highly skewed keys).

reconditesea:
To close #461


algebird-core/src/main/scala/com/twitter/algebird/CountMinSketch.scala
```diff
@@ -458,6 +458,59 @@ case class CMSItem[K](item: K, override val totalCount: Long, override val param
...
+  override def +(x: K, count: Long): CMS[K] = {
+    val currentCount = exactCountTable.getOrElse(x, 0L)
+    val newTable = exactCountTable.updated(x, currentCount + count)
+    if (newTable.size < width * depth) {
+      // still sparse
+      SparseCMS(newTable, totalCount + count, params)
+    } else {
+      // Create new CMSInstace
+      var cms = CMSInstance[K](params)
+      newTable.foreach {
```
johnynek: just to be consistent, let's use foldLeft rather than foreach here (to avoid an unneeded var). I'm willing to have vars if they add performance, in this case the perf should be the same as:
```scala
newTable.foldLeft(CMSInstance[K](params))(_ + _)
```

algebird-core/src/main/scala/com/twitter/algebird/CountMinSketch.scala
```diff
@@ -458,6 +458,59 @@ case class CMSItem[K](item: K, override val totalCount: Long, override val param
...
+          cms = cms + (x, count)
+      }
+      cms
+    }
+  }
+
+  override def ++(other: CMS[K]): CMS[K] = {
+    other match {
+      case other: CMSZero[_] => this
+      case other: CMSItem[K] => this + other.item
```
johnynek: there is a bug here. Let's fix it while we are at it. This is neglecting the total in CMSItem.

algebird-core/src/main/scala/com/twitter/algebird/CountMinSketch.scala
```diff
@@ -458,6 +458,59 @@ case class CMSItem[K](item: K, override val totalCount: Long, override val param
...
+        }
+
+      case other: CMSInstance[K] => other ++ this
+    }
+  }
+
+  override def frequency(x: K): Approximate[Long] = Approximate.exact(exactCountTable.getOrElse(x, 0L))
+
+  override def innerProduct(other: CMS[K]): Approximate[Long] =
+    exactCountTable.map { case (x, count) => Approximate.exact(count) * other.frequency(x) }.reduce { _ + _ }
```
johnynek: let's do `exactCountTable.iterator.map ...` here so we are lazy until we hit the reduce. Also, what if exactCountTable is empty? I think reduce throws in that case. Might want to do .reduceOption and put a `getOrElse(Approximate.exact(0L))`

algebird-core/src/main/scala/com/twitter/algebird/CountMinSketch.scala
```diff
@@ -468,6 +521,14 @@ case class CMSInstance[K](countsTable: CMSInstance.CountsTable[K],
     other match {
       case other: CMSZero[_] => this
       case other: CMSItem[K] => this + other.item
+      case other: SparseCMS[K] =>
+        val newTotalCount = totalCount + other.totalCount
+        var cms = this
+        other.exactCountTable.foreach {
```
johnynek: `other.exactCountTable.foldLeft(this)(_ + _)`

algebird-core/src/main/scala/com/twitter/algebird/CountMinSketch.scala
```diff
@@ -468,6 +521,14 @@ case class CMSInstance[K](countsTable: CMSInstance.CountsTable[K],
     other match {
       case other: CMSZero[_] => this
       case other: CMSItem[K] => this + other.item
+      case other: SparseCMS[K] =>
+        val newTotalCount = totalCount + other.totalCount
```
johnynek: you are not using this, right?

algebird-core/src/main/scala/com/twitter/algebird/CountMinSketch.scala
```diff
@@ -458,6 +458,59 @@ case class CMSItem[K](item: K, override val totalCount: Long, override val param
...
+      case other: CMSZero[_] => this
+      case other: CMSItem[K] => this + other.item
+      case other: SparseCMS[K] =>
+        val newTable = exactCountTable ++ other.exactCountTable
+        if (newTable.size < width * depth) {
+          // still sparse
+          SparseCMS(newTable, totalCount + other.totalCount, params)
+        } else {
+          // Create new CMSInstace
+          var cms = CMSInstance[K](params)
```
johnynek: let's remove this duplication with a `toDense: CMS[K]` method which you can call here and in `+`.

algebird-core/src/main/scala/com/twitter/algebird/CountMinSketch.scala
johnynek: might need to do `newTable.foldLeft(CMSInstance[K](params)) { case (cms, (k, c)) => cms + (k, c) }` because `+` takes two args and not a tuple.

johnynek:
should we add a test and fix #459 at the same time (you added another copy of this bug in this PR)?

Also, can we add a test to specifically test this path? Like make a scalacheck generator that only generates sparse values, and verify that when we add them, either the count is exact or we have a CMSInstance?

algebird-core/src/main/scala/com/twitter/algebird/CountMinSketch.scala
reconditesea: Good call! I will put it to the companion object.

reconditesea:
@johnynek Didn't know #459 before. Will fix it in this PR along with some tests.

algebird-core/src/main/scala/com/twitter/algebird/CountMinSketch.scala
reconditesea: Fixed.

johnynek:
This is related to #390 (kind of forgot about that, sorry folks). I guess this will supercede that one.

algebird-core/src/main/scala/com/twitter/algebird/CountMinSketch.scala
```diff
@@ -441,23 +441,88 @@ case class CMSZero[K](override val params: CMSParams[K]) extends CMS[K](params)
...
+    exactCountTable.iterator.map { case (x, count) => Approximate.exact(count) * other.frequency(x) }
+      .reduceOption { _ + _ }.getOrElse(Approximate.exact(0L))
+}
+
+object SparseCMS {
+
+  /**
+   * Creates a new [[SparseCMS]] with empty exactCountTable
+   */
+  def apply[K](params: CMSParams[K]): SparseCMS[K] = {
```
johnynek: suppose we call this `.empty`. What do you think?

johnynek: actually, this follows `CMSInstance` so let's keep it as is.

algebird-core/src/main/scala/com/twitter/algebird/CountMinSketch.scala
```diff
@@ -441,23 +441,88 @@ case class CMSZero[K](override val params: CMSParams[K]) extends CMS[K](params)
  */
 case class CMSItem[K](item: K, override val totalCount: Long, override val params: CMSParams[K]) extends CMS[K](params) {
 
-  override def +(x: K, count: Long): CMS[K] = CMSInstance[K](params) + item + (x, count)
+  override def +(x: K, count: Long): CMS[K] = CMSInstance[K](params) + (item, totalCount) + (x, count)
```
johnynek: nice! Can we add a test to check this to also prevent any regressions?

reconditesea: By all means, actually I think "override def +(x: K, count: Long): CMS[K] = CMSInstance[K](params) + (item, totalCount) + (x, count)" also is not correct.
Should be using SparseCMS here "override def +(x: K, count: Long): CMS[K] = SparseCMS[K](params) + (item, totalCount)  + (x, count)"
I will change this.

johnynek:
This all looks good, but now looking back at #390, one thing I see that is different is that there was a parameter added to control how big the sparse CMS got, rather than just always doing width * depth (which I guess could use MUCH more than a CMS if the key is a String or BigInt, for instance).

Maybe we should copy that approach of adding a parameter (and perhaps have it default to something very small, like `max(width * depth / 100, 10)` or something.

reconditesea:
@johnynek Is this good to go?

reconditesea:
Seems the  coverage/coveralls test is taking forever to finish. Is there a way to turn it off or somehow?

algebird-core/src/main/scala/com/twitter/algebird/CountMinSketch.scala
```diff
@@ -441,23 +441,90 @@ case class CMSZero[K](override val params: CMSParams[K]) extends CMS[K](params)
...
 
-  override def innerProduct(other: CMS[K]): Approximate[Long] = other.frequency(item)
+  override def innerProduct(other: CMS[K]): Approximate[Long] = Approximate.exact(totalCount) * other.frequency(item)
 
 }
 
 /**
+ * A sparse Count-Min sketch structure, used for situations where the key is highly skewed.
+ */
+case class SparseCMS[K](exactCountTable: Map[K, Long], maxExactCountOpt: Option[Int] = None,
```
johnynek: let's put the maxExactCountOpt into CMSParams. Without that, you can't really control it, right? Since CMSItem will always create with the default setting, right?

reconditesea: @johnynek Good points. But that will make maxExactCountOpt available to all CMS subclasses. Is that desired?

johnynek: I agree it is only relevant to one part of the state machine, but similarly, the size is only depth/width is only relevant to when we create the hash count matrix (not to CMSItem or SparseCMS). So, I think it still fits as a parameter of the operation of the monoid.

reconditesea: Make sense. Has made the change as suggested.

On Tue, Jul 28, 2015 at 10:41 AM, P. Oscar Boykin <notifications@github.com>
wrote:

> In algebird-core/src/main/scala/com/twitter/algebird/CountMinSketch.scala
> <https://github.com/twitter/algebird/pull/464#discussion_r35676556>:
>
> >
> >  }
> >
> >  /**
> > + * A sparse Count-Min sketch structure, used for situations where the key is highly skewed.
> > + */
> > +case class SparseCMS[K](exactCountTable: Map[K, Long], maxExactCountOpt: Option[Int] = None,
>
> I agree it is only relevant to one part of the state machine, but
> similarly, the size is only depth/width is only relevant to when we create
> the hash count matrix (not to CMSItem or SparseCMS). So, I think it still
> fits as a parameter of the operation of the monoid.
>
> —
> <https://github.com/twitter/algebird/pull/464/files#r35676556>.
>



-- 
Kevin Lin | Twitter, Inc.
1355 Market St. | San Francisco, CA | 94103

Follow me: @reconditesea <https://twitter.com/reconditesea>

reconditesea:
@johnynek Is this good to go :)

ianoc:

algebird-core/src/main/scala/com/twitter/algebird/CountMinSketch.scala
```diff
@@ -441,23 +460,89 @@ case class CMSZero[K](override val params: CMSParams[K]) extends CMS[K](params)
...
+    }
+  }
+
+  override def ++(other: CMS[K]): CMS[K] = {
+    other match {
+      case other: CMSZero[_] => this
+      case other: CMSItem[K] => this + (other.item, other.totalCount)
+      case other: SparseCMS[K] =>
+        // This SparseCMS's maxExactCount is used, so ++ is not communitive
+        val newTable = exactCountTable ++ other.exactCountTable
```
johnynek: wait, we don't want `++` here right? We want `Semigroup.plus(exactCountTable, other.exactCountTable)`.

If that's correct, can you write a test that exposes this bug so we don't have a regression on this?

reconditesea: You're right. This should be a Semigroup add. I will make an iteration and also add a test.

algebird-core/src/main/scala/com/twitter/algebird/CountMinSketch.scala
```diff
@@ -89,6 +89,7 @@ package com.twitter.algebird
  *        (an interval that depends on `eps`) around the truth.
  * @param seed  A seed to initialize the random number generator used to create the pairwise independent
  *        hash functions.
+ * @param
```
DanielleSucher: Looks like this tag is missing info on the new param?

reconditesea: @DanielleSucher  You're absolutely right! I will add actual comments here as well.

DanielleSucher: :sparkling_heart:

reconditesea:
@johnynek @DanielleSucher Here you go!

algebird-core/src/main/scala/com/twitter/algebird/CountMinSketch.scala
```diff
@@ -351,24 +369,24 @@ trait CMSHeavyHitters[K] {
...
+    monoid(CMSFunctions.eps(width), CMSFunctions.delta(depth), seed, maxExactCountOpt)
 
-  def aggregator[K: CMSHasher](eps: Double, delta: Double, seed: Int): CMSAggregator[K] =
-    new CMSAggregator[K](monoid(eps, delta, seed))
+  def aggregator[K: CMSHasher](eps: Double, delta: Double, seed: Int, maxExactCountOpt: Option[Int] = None): CMSAggregator[K] =
+    new CMSAggregator[K](monoid(eps, delta, seed, maxExactCountOpt))
 
-  def aggregator[K: CMSHasher](depth: Int, width: Int, seed: Int): CMSAggregator[K] =
-    aggregator(CMSFunctions.eps(width), CMSFunctions.delta(depth), seed)
+  def aggregator[K: CMSHasher](depth: Int, width: Int, seed: Int, maxExactCountOpt: Option[Int]): CMSAggregator[K] =
```
johnynek: shouldn

algebird-core/src/main/scala/com/twitter/algebird/CountMinSketch.scala
```diff
@@ -351,24 +369,24 @@ trait CMSHeavyHitters[K] {
 
 object CMS {
 
-  def monoid[K: CMSHasher](eps: Double, delta: Double, seed: Int): CMSMonoid[K] = new CMSMonoid[K](eps, delta, seed)
+  def monoid[K: CMSHasher](eps: Double, delta: Double, seed: Int, maxExactCountOpt: Option[Int] = None): CMSMonoid[K] = new CMSMonoid[K](eps, delta, seed, maxExactCountOpt)
 
-  def monoid[K: CMSHasher](depth: Int, width: Int, seed: Int): CMSMonoid[K] =
-    monoid(CMSFunctions.eps(width), CMSFunctions.delta(depth), seed)
+  def monoid[K: CMSHasher](depth: Int, width: Int, seed: Int, maxExactCountOpt: Option[Int]): CMSMonoid[K] =
```
johnynek: have an option with maxExactCountOpt = None?

reconditesea: I tried to add maxExactCountOpt = None here as well. But scala complains about duplicate definition of this def with the above (def monid), saying both of them have the same signature. I think it's because of type-erasure. 
Not sure if there is a good way to work around?

ianoc: you could add the dummy implicit route i guess to get around it

reconditesea: Offline discussed with @ianoc . DummyImplicit won't help here, since the dup definition lies in the helper method generated for default values.

I will create multiple versions of these methods, w/ or w/o the new maxExactCountOpt respectively.

ianoc:
How does this effect our CMS benchmarks?

reconditesea:
@ianoc  Could you share some more details about the current CMS benchmarks? Not very familiar with it. But given the default maxExactCount is quite small, I believe it won't affect it too much.

ianoc:
The algebird-benchmark project has some benchmarks like
https://github.com/twitter/algebird/blob/develop/algebird-benchmark/src/main/scala/com/twitter/algebird/benchmark/CMSBenchmark.scala

It would be good to run these on develop vs this branch to ensure no regressions.


Even if your code doesn't in theory effect the higher density CMS's, if it for instances blocks any specialization then that would have a large impact everywhere.

reconditesea:
@ianoc 
On my branch: [success] Total time: 3255 s, completed Aug 10, 2015 1:36:09 PM
On develop: [success] Total time: 3269 s, completed Aug 10, 2015 2:41:02 PM

It's not a super clean comparison, because I'm doing other stuffs on my computer while the benchmarks are running. So the develop even took slightly more time than branch. But I guess it proves they are within the same performance level?

ianoc:
Thats not the benchmark timing? the benchmark should be a matrix of output timing. That looks like the output from sbt?  see the docs in the readme at the root or the sbt-jmh plugin docs

reconditesea:
Here you go.
My brach:
branch

develop:
develop

ianoc:
That suggests about 4% slow down in all the expensive jobs there right?

reconditesea:
I reran my branch's benchmark and compared it with develop. This time, some of them are better whole others are still slower than develop. I guess there are some variances in different benchmark runs.
I think the toDense() def of SparseCMS will add some additional cost, since it goes over all (k,v) items once. But that should not be too much. Any suggestions?

branch2

ianoc:
I'm not sure, 4% loss in perf while not huge, is non-trival to me. Can you profile one of consistently 4% slower ones and see where it spends the time? 

Are any of them significantly faster than develop?

reconditesea:
@ianoc I think the reason of previous slowness is that the MaxExactCount (key-count) set for SparseCMS is too low (at most 10). So in the benchmark easily a SparseCMS need to be converted to a CMSInstance, and therefore each element needs to be added to the new CMS CountTable one by one. That brings the cost. If I set MaxExactCount to 50, my branch has comparable performance as develop.
Do we think key space 50 is a good deal for a SparseCMS key size?

Branch
branch

Develop
develop

ianoc:
That seems reasonable to me, even 50 might wind up being too low. But given we have comparable performance at that level and its a configurable parameter. LGTM.

Any others issues @johnynek & @DanielleSucher ?

reconditesea:
@johnynek & @DanielleSucher, any suggestions?

reconditesea:
@ianoc  Is this good to go? Thanks!

ianoc:
Merged thanks

Issue #445, opened 2015-05-21
Add Macros for case class Algebraic structures


johnynek:
Similar to the tuple case, we should be able to:

```scala
import com.twitter.algebird.macros.caseclass
```
and be good to go.

PR #466, opened 2015-07-16
Create case class macros for algebraic structures

sid-kap:
Here's an initial sketch of some macros to generate Semigroup, Monoid, Group, Ring, and Field typeclasses of case classes. See issue #445.

To use these, you need to include
```scala
import com.twitter.algebird.macros.caseclass._
```
and you're good to go.

Potential issues with this implementation:
* This adds scalamacros.quasiquotes and scalamacros.paradise as dependencies to algebird-core. We might want to move these macros to a new module "scalding-macros" so that people who don't want the macros don't need these dependencies.
* My macros don't extend each other, i.e. if we have `case class Foo(m1: Monoid1, m2: Monoid2)` and we are generating `implicitly[Monoid[Foo]]`, the macro generates 

    ```scala
    def plus(a: Foo, b: Foo) = 
      Foo(implicitly[Monoid[Monoid1]].plus(a.m1, b.m1), 
          implicitly[Monoid[Monoid2]].plus(a.m2, b.m2))
    ```
    rather than
    ```scala
def plus(a: Foo, b: Foo) = implicitly[Semigroup[Foo]].plus(a, b)
    ```
    I don't know if it makes a difference which way we implement this, besides the fact that my implementation is a few more lines of macro code.
* We should add some tests for nested case classes. In principle, they should work.
* In theory, these macros obviate the generated tuple macros, since tuples are actually case classes. Using these macros instead of the generated tuple macros would increase compile times by a bit. However, it might be worth it if it means we can get rid of the hacky ruby script.


algebird-core/src/main/scala/com/twitter/algebird/macros/MonoidMacro.scala
```diff
@@ -0,0 +1,37 @@
...
+
+    val plusList = params.map(param => q"implicitly[_root_.com.twitter.algebird.Monoid[$param]].plus(l.$param, r.$param)")
+    val zerosList = params.map(param => q"implicitly[_root_.com.twitter.algebird.Monoid[$param]].zero")
+
+    val companion = tpe.typeSymbol.companionSymbol
+
+    val res = q"""
+    new _root_.com.twitter.algebird.Monoid[$T] {
+      def plus(l: $T, r: $T): $T = $companion.apply(..$plusList)
+      def zero: $T = $companion.apply(..$zerosList)
```
johnynek: can you also add `isNonZero`?

algebird-core/src/main/scala/com/twitter/algebird/macros/SemigroupMacro.scala
```diff
@@ -0,0 +1,35 @@
...
+    }.toList
+
+    val plusList = params.map(param => q"implicitly[_root_.com.twitter.algebird.Semigroup[$param]].plus(l.$param, r.$param)")
+
+    val companion = tpe.typeSymbol.companionSymbol
+
+    val res = q"""
+    new _root_.com.twitter.algebird.Semigroup[$T] {
+      def plus(l: $T, r: $T): $T = $companion.apply(..$plusList)
+    }
```
johnynek: for performance, it is really important to implement sumOption. See how we generated it here:

https://github.com/twitter/algebird/blob/develop/algebird-core/src/main/scala/com/twitter/algebird/GeneratedAbstractAlgebra.scala#L9

algebird-test/src/test/scala/com/twitter/algebird/macros/CaseClassMacrosTest.scala
```diff
@@ -0,0 +1,90 @@
...
+import org.scalatest._
+import com.twitter.algebird._
+import com.twitter.algebird.macros.caseclass._
+
+class CaseClassMacrosTest extends WordSpec with Matchers {
+
+  case class Foo(a: Int, b: Int)
+  case class DoubleFoo(a: Double, b: Double)
+
+  "SemigroupMacro" should {
```
johnynek: there are stronger laws you can use. See:

https://github.com/twitter/algebird/blob/develop/algebird-test/src/test/scala/com/twitter/algebird/CountMinSketchTest.scala#L27

So, we should be able to check `semigroupLaws[Foo]` and we can be sure everything is correct. You will need to implement a generator for Foo (which again, we could implement if no one else has, an Arbitrary[T] macro for case classes).

algebird-test/src/test/scala/com/twitter/algebird/macros/CaseClassMacrosTest.scala
```diff
@@ -0,0 +1,90 @@
...
+
+      assert(r.plus(a, b) == Foo(5, 5))
+      assert(r.zero == Foo(0, 0))
+      assert(r.negate(a) == Foo(-1, -2))
+      assert(r.one == Foo(1, 1))
+      assert(r.times(a, b) == Foo(4, 6))
+    }
+  }
+
+  "FieldMacro" should {
```
johnynek: this actually can't work. The pointwise version of a bunch of Fields is not a Field. You can use the `fieldLaws` test to verify that it will always fail.

Then we should remove the macro.

The issue is this: to be a field, only `0` can not have an inverse, but consider `(Double, Double)`. `(0.0, 0.0)` is the zero, but notice that `(1.0, 0.0)` and `(0.0, 1.0)` neither have inverses.

johnynek:
Nice work! Just a few minor comments:

1) use the Laws we already have for comprehensive (and easy!) testing.
2) Field won't work. You can see this based on all the exception handling you are doing, but really we want type-safety, so we want to never throw an exception if possible.
3) we need to implement sumOption, for some Monoids/Semigroups this makes a huge performance difference (lots of the sketches can amortize the immutable costs by batching updates, and scalding makes use of this).

algebird-core/src/main/scala/com/twitter/algebird/macros/MonoidMacro.scala
sid-kap: One way to make `isNonZero` more efficient is changing `def zero` to `val zero`. This will ensure that zero is only computed once. Then any call to `isNonZero(x)` will just return `x != zero`. This should be reasonably efficient.

What did you have in mind for an implementation of `isNonZero`?

algebird-core/src/main/scala/com/twitter/algebird/macros/SemigroupMacro.scala
sid-kap: Done

algebird-test/src/test/scala/com/twitter/algebird/macros/CaseClassMacrosTest.scala
sid-kap: Done

algebird-test/src/test/scala/com/twitter/algebird/macros/CaseClassMacrosTest.scala
sid-kap: Removed field macros.

algebird-core/src/main/scala/com/twitter/algebird/macros/GroupMacro.scala
```diff
@@ -0,0 +1,45 @@
...
+    }
+    """
+    c.Expr[Group[T]](res)
+  }
+
+  def negate[T](c: Context)(implicit T: c.WeakTypeTag[T]): c.Tree = {
+    import c.universe._
+
+    val companion = macros.getCompanionObject(c)
+    val negateList = getParams(c).map(param => q"implicitly[_root_.com.twitter.algebird.Group[$param]].negate(x.$param)")
```
johnynek: how is this working `Group[$param]`? Isn't `param` the name of the accessor, not the type?

algebird-core/src/main/scala/com/twitter/algebird/macros/GroupMacro.scala
```diff
@@ -0,0 +1,45 @@
...
+
+object GroupMacro {
+  def caseClassGroup[T](c: Context)(implicit T: c.WeakTypeTag[T]): c.Expr[Group[T]] = {
+    import c.universe._
+
+    macros.ensureCaseClass(c)
+
+    val res = q"""
+    new _root_.com.twitter.algebird.Group[$T] {
+      ${SemigroupMacro.plus(c)}
```
johnynek: this is a really nice solution, and the code reuse is great. There is one issue here: what if you have conflicting `Semigroup[T]` and `Group[T]`. For instance, `Semigroup[Int]` could be multiplication, so `plus` would actually be addition, but `Group[Int]` is going to have the `plus` method be addition.

This won't happen by default, but a user could shoot themselves in the foot.

We could avoid this by getting all the implicit `Group`s for the member types, and then explicitly passing those into the SemigroupMacro so it does not resolve them again. It's not immediately clear the best way to do that. Can you try something like that?

algebird-test/src/main/scala/com/twitter/algebird/macros/ArbitraryCaseClassMacro.scala
```diff
@@ -0,0 +1,53 @@
...
+    val companion = tpe.typeSymbol.companionSymbol
+
+    val res = q"""
+    for ( ..$getsList ) yield $companion.apply(..$paramsList)
+    """
+
+    c.Expr[Gen[T]](res)
+  }
+
+  def caseClassArbitrary[T](c: Context)(implicit T: c.WeakTypeTag[T]): c.Expr[Arbitrary[T]] = {
```
johnynek: this is hugely useful. Nice work. It should almost go in its own package, but we can leave it here for now.

This will be really useful at Twitter to have this (and a similar one for scrooge (and java thrift) if possible).

algebird-core/src/main/scala/com/twitter/algebird/macros/GroupMacro.scala
sid-kap: Nice catch! It should be `Group[${param.returnType}]`.

sid-kap:
I updated the macros so that, for example, every method in the Group macro uses `implicitly[Group[${param.returnType}]`, rather than using an implicit Semigroup for the plus method, etc.

However, the implicits are still potentially being resolved multiple times. We're calling `implicitly[Group[${param.returnType}]]` multiple times, rather than storing `val implicitGroup1 = implicitly[Group[${param.returnType}]]` and using `implicitGroup1` in the future.

Should I change the macros so that each implicit is only resolved once? I'm not sure if it would change any behavior, but it would probably speed up compile time by a bit.

algebird-test/src/test/scala/com/twitter/algebird/macros/CaseClassMacrosTest.scala
```diff
@@ -0,0 +1,45 @@
...
+class CaseClassMacrosTest extends PropSpec with PropertyChecks with Matchers {
+  import BaseProperties._
+
+  implicit val arbitraryFoo: Arbitrary[Foo] = arbitrary[Foo]
+  implicit val arbitraryBar: Arbitrary[Bar] = arbitrary[Bar]
+
+  case class Foo(a: Int, b: Short, c: Long)
+  case class Bar(a: Boolean, foo: Foo)
+
+  property("Foo is a Semigroup") {
```
johnynek: I'm not sure these are running. We had some issue with this in the past. I really hate scalatest. It has so many styles.

The risk is this: semigroupLaws returns a Prop. It does not do any asserts. But scalatest works on asserts. So we had tests that were failing, but we didn't notice for a long time due to this.

Rather than really fix the tests, we make a trait to do the right thing, and mixed that in. The problem is, copying that is easy to forget or hard to review singe you need to remember this issue.

I think this might not be reporting errors since I think it is expecting an assert inside the property, block.

Can we actually not use scalatest here and just use scalacheck? It is too easy to get these tests wrong.

sid-kap:
I've updated the tests to use Scalacheck only.
Does this look correct?

johnynek:
this looks right to me. Nice work.

algebird-core/src/main/scala/com/twitter/algebird/macros/SemigroupMacro.scala
```diff
@@ -0,0 +1,68 @@
...
+      case (param, instance) =>
+        q"val ${param.name.asInstanceOf[TermName]} = $instance.sumOption(items.iterator.map(_.$param))"
+    }
+
+    val result = q"$companion.apply(..$sumOptionsGetted)"
+
+    val getBlock = Block(getSumOptions, result)
+
+    q"""
+    override def sumOption(to: TraversableOnce[$T]): _root_.scala.Option[$T] = {
```
johnynek: let's add the `_root_.scala...` prefix to `TraversableOnce`.

johnynek:
merge when green.

joshualande:
@sid-kap, thanks so much for working on this! This is a feature I am really looking forward to!

PR #467, opened 2015-07-17
Move CMSHasherByteArray from scalding

sid-kap:
See https://github.com/twitter/scalding/issues/1177.
@isnotinvain

### DIFF
--- a/algebird-core/src/main/scala/com/twitter/algebird/CountMinSketch.scala
+++ b/algebird-core/src/main/scala/com/twitter/algebird/CountMinSketch.scala
@@ -1207,4 +1207,18 @@ object CMSHasherImplicits {
     override def hash(a: Int, b: Int, width: Int)(x: Bytes): Int = hashBytes(a, b, width)(x.array)
   }
 
+  /**
+   * This is based on the CMSHasherBigInt found in algebird (see docs for in depth explanation):
+   * https://github.com/twitter/algebird/blob/develop/algebird-core/src/main/scala/com/twitter/algebird/CountMinSketch.scala#L1086
+   */
+  implicit object CMSHasherByteArray extends CMSHasher[Array[Byte]] {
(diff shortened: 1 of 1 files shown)


algebird-core/src/main/scala/com/twitter/algebird/CountMinSketch.scala
```diff
@@ -1207,4 +1207,18 @@ object CMSHasherImplicits {
...
 
-}
\ No newline at end of file
+  /**
+   * This is based on the CMSHasherBigInt found in algebird (see docs for in depth explanation):
+   * https://github.com/twitter/algebird/blob/develop/algebird-core/src/main/scala/com/twitter/algebird/CountMinSketch.scala#L1086
+   */
+  implicit object CMSHasherByteArray extends CMSHasher[Array[Byte]] {
+    override def hash(a: Int, b: Int, width: Int)(x: Array[Byte]): Int = {
+      val hash: Int = scala.util.hashing.MurmurHash3.arrayHash(x, a)
```
johnynek: this should just call hashBytes directly rather than copy the code.

algebird-core/src/main/scala/com/twitter/algebird/CountMinSketch.scala
```diff
@@ -1207,4 +1207,12 @@ object CMSHasherImplicits {
     override def hash(a: Int, b: Int, width: Int)(x: Bytes): Int = hashBytes(a, b, width)(x.array)
   }
 
-}
\ No newline at end of file
+  /**
```
isnotinvain: this comment isn't needed anymore now that this is in the right home

isnotinvain:
+1 after dropping the extra comment

johnynek:
looks like we need the same fix as:

https://github.com/twitter/scalding/pull/1381

johnynek:
can we merge with develop?

sid-kap:
Merged

PR #470, opened 2015-07-23
Update sbt launcher script (sbt-extras)

sid-kap:
Same thing as https://github.com/twitter/scalding/pull/1381

Got latest version of sbt launcher script from https://github.com/paulp/sbt-extras


sid-kap:
Whoops just realized that this is a duplicate of https://github.com/twitter/algebird/pull/469

PR #471, opened 2015-07-23
Adding scoped top-N CMS monoid

### LINKED ISSUE #468: Scoped TopN CMS
Conceptually, we want behavior somewhat like `Map[K, TopNCMS[V]]`.  For each key, we separately store an estimate of the N most frequent values.  To save space we can instead use a single `TopCMS[(K, V)]`.  The meat of this would be a specialized `HeavyHittersLogic[(K, V)]` that only does evictions per key.

jnievelt:
Closes #468

A bit of refactoring out duplicate code.  Leaving in `TopPctCMSAggregator` and `TopNCMSAggregator` so this should be binary compatible.

### DIFF
--- a/algebird-core/src/main/scala/com/twitter/algebird/CountMinSketch.scala
+++ b/algebird-core/src/main/scala/com/twitter/algebird/CountMinSketch.scala
@@ -720,6 +720,47 @@ case class TopCMSInstance[K](override val cms: CMS[K], hhs: HeavyHitters[K], par
 
 }
 
+class TopCMSMonoid[K](cms: CMS[K], logic: HeavyHittersLogic[K]) extends Monoid[TopCMS[K]] {
+
+  val params: TopCMSParams[K] = TopCMSParams(logic)
+
+  val zero: TopCMS[K] = TopCMSZero[K](cms, params)
+
+  /**
+   * Combines the two sketches.
+   *
+   * The sketches must use the same hash functions.
+   */
+  def plus(left: TopCMS[K], right: TopCMS[K]): TopCMS[K] = {
+    require(left.cms.params.hashes == right.cms.params.hashes, "The sketches must use the same hash functions.")
+    left ++ right
+  }
+
+  /**
+   * Creates a sketch out of a single item.
+   */
+  def create(item: K): TopCMS[K] = TopCMSItem[K](item, cms + item, params)
+
+  /**
+   * Creates a sketch out of multiple items.
+   */
+  def create(data: Seq[K]): TopCMS[K] = {
+    data.foldLeft(zero) { case (acc, x) => plus(acc, create(x)) }
+  }
+
+}
+
+class TopCMSAggregator[K](cmsMonoid: TopCMSMonoid[K])
+  extends MonoidAggregator[K, TopCMS[K], TopCMS[K]] {
+
+  def monoid = cmsMonoid
+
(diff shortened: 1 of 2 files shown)


johnynek:
can you merge with develop to get the sbt fix?

ianoc:

algebird-core/src/main/scala/com/twitter/algebird/CountMinSketch.scala
```diff
@@ -780,7 +821,7 @@ case class TopNLogic[K](heavyHittersN: Int) extends HeavyHittersLogic[K] {
 
   require(heavyHittersN > 0, "heavyHittersN must be > 0")
 
-  override def purgeHeavyHitters(cms: CMS[K])(hitters: HeavyHitters[K]): HeavyHitters[K] = {
+  def purgeHeavyHitters(cms: CMS[K])(hitters: HeavyHitters[K]): HeavyHitters[K] = {
```
johnynek: why did you remove the override here?

algebird-core/src/main/scala/com/twitter/algebird/CountMinSketch.scala
```diff
@@ -758,7 +799,7 @@ case class TopPctLogic[K](heavyHittersPct: Double) extends HeavyHittersLogic[K]
 
   require(0 < heavyHittersPct && heavyHittersPct < 1, "heavyHittersPct must lie in (0, 1)")
 
-  override def purgeHeavyHitters(cms: CMS[K])(hitters: HeavyHitters[K]): HeavyHitters[K] = {
+  def purgeHeavyHitters(cms: CMS[K])(hitters: HeavyHitters[K]): HeavyHitters[K] = {
```
johnynek: why no override? isn't this from HeavyHittersLogic?

algebird-core/src/main/scala/com/twitter/algebird/CountMinSketch.scala
```diff
@@ -1045,14 +1017,83 @@ object TopNCMS {
 /**
  * An Aggregator for [[TopNCMS]].  Can be created using [[TopNCMS.aggregator]].
  */
-case class TopNCMSAggregator[K](cmsMonoid: TopNCMSMonoid[K])
-  extends MonoidAggregator[K, TopCMS[K], TopCMS[K]] {
+case class TopNCMSAggregator[K](cmsMonoid: TopNCMSMonoid[K]) extends TopCMSAggregator(cmsMonoid)
 
-  val monoid = cmsMonoid
+case class ScopedTopNLogic[K1, K2](heavyHittersN: Int) extends HeavyHittersLogic[(K1, K2)] {
```
johnynek: can we add a comment to this logic about what is going on? I think for each `K1`, we keep the top `heavyHittersN`.

johnynek:
just a few minor comments, then let's merge.

algebird-core/src/main/scala/com/twitter/algebird/CountMinSketch.scala
jnievelt: Eh I kinda love-hate the override keyword.  It doesn't require that the base class/trait actually implements the method (i.e., so that we know if super is interesting).  Then again, plain def doesn't require that the base class/trait declares the method either.  I'll put it back.

Issue #377, opened 2014-12-12
Possible off by 1 error for QTrees


joshualande:
+ @avibryant 

I was playing around with the QTree, and I think there may be an off-by one error in the implementation.

Consider this code snippet:
```scala
import com.twitter.algebird.{QTree,QTreeSemigroup}

val K = 10
implicit val qtreeSemigroup = new QTreeSemigroup[Double](K)

val l = (1 to 3).map { _.toDouble }
val qt = l.map { v =>QTree(v)}.reduceLeft( qtreeSemigroup.plus)
qt.quantileBounds(0.5)
```
This returns: 
```
res39: (Double, Double) = (1.0,1.0000152587890625)
```
Shouldn't the median of the list (1, 2, 3) be 2, not 1?

Out of curiosity, I tried the numbers 1 to 11, and got a median of 5 instead of 6.

And I tried the numbers 1 to 101 and got a median of 50 instead of 51.

I also tried a K of 100,000 but still get a median of 50 (for 1 to 101)

I think this is caused by an off-by-one error in the QTree implementation.

I don't think an off-by-one bug would cause any practical problems since I have lots of data, but it might be worth investigating.

Thanks!

Joshua

johnynek:
I see two tasks here: 1) convert Josh's examples into property tests. 2) fix the issue.

joshualande:
Awesome, thanks so much!

PR #472, opened 2015-07-28
Fix Qtree quantileBounds off-by-one error

sid-kap:
See issue #377


algebird-core/src/main/scala/com/twitter/algebird/QTree.scala
```diff
@@ -197,36 +197,38 @@ case class QTree[A](
...
-      None
-    else {
-      lowerChild.flatMap{ _.findRankUpperBound(rank) }.orElse {
-        val lowerCount = lowerChild.map{ _.count }.getOrElse(0L)
-        upperChild.flatMap{ _.findRankUpperBound(rank - lowerCount) }.orElse(Some(upperBound))
-      }
+    val (leftCount, rightCount) = mapChildrenWithDefault(0L)(_.count)
+    val parentCount = count - leftCount - rightCount
+
+    if (0 <= rank && rank < leftCount) {
```
jnievelt: 0 <= rank is redundant since we'll always run the require validation?

sid-kap: Yep, that's just for clarity.
I can remove it if you want.

johnynek: you can comment `// note 0 <= rank due to assert above`
or something, but let's not repeat the code.

algebird-core/src/main/scala/com/twitter/algebird/QTree.scala
```diff
@@ -197,36 +197,38 @@ case class QTree[A](
...
+
+      // lowerChild.get is safe because
+      // leftCount > 0, so lowerChild is not None.
+      lowerChild.get.findRankBounds(rank)
+
+    } else if (leftCount <= rank && rank < leftCount + parentCount) {
+      (lowerBound, upperBound)
+    } else {
+      // so leftCount + parentCount <= rank < count
```
jnievelt: IMO just the first and last comment lines here are sufficient

algebird-test/src/test/scala/com/twitter/algebird/QTreeTest.scala
```diff
@@ -68,6 +68,21 @@ class QTreeTest extends WordSpec with Matchers {
   def trueRangeSum(list: Seq[Double], from: Double, to: Double) =
     list.filter{ _ >= from }.filter{ _ < to }.sum
 
+  for (k <- Seq(3, 11, 51, 101)) {
+    s"QTree with elements (1 to $k)" should {
+      val trueMedian = (1 + k) / 2
+      s"have median $trueMedian" in {
+        implicit val sg = new QTreeSemigroup[Double](k)
```
johnynek: `QTreeSemigroup[Unit]` should work, I think, and is cheaper. People seem to be cargo-culting `QTreeSemigroup[Double]` Can we switch here?

algebird-test/src/test/scala/com/twitter/algebird/QTreeTest.scala
```diff
@@ -68,6 +68,21 @@ class QTreeTest extends WordSpec with Matchers {
...
+
+        val list = (1 to k).map(_.toDouble)
+        val qtree = sg.sumOption(list.map(QTree(_))).get
```
johnynek: `QTree.value(_)` I think will work.

algebird-core/src/main/scala/com/twitter/algebird/QTree.scala
jnievelt: I think it's more confusing actually.  If the code looks like it handles the case where rank < 0, a reasonable person may (initially) infer that it's possible for that to be true.

algebird-test/src/test/scala/com/twitter/algebird/QTreeTest.scala
sid-kap: Updated in latest commit

algebird-test/src/test/scala/com/twitter/algebird/QTreeTest.scala
sid-kap: Updated in latest commit

algebird-core/src/main/scala/com/twitter/algebird/QTree.scala
sid-kap: That's a good point. I've fixed this in the latest commit.

algebird-core/src/main/scala/com/twitter/algebird/QTree.scala
```diff
@@ -197,36 +197,32 @@ case class QTree[A](
...
-  private def findRankUpperBound(rank: Long): Option[Double] = {
-    if (rank > count)
-      None
-    else {
-      lowerChild.flatMap{ _.findRankUpperBound(rank) }.orElse {
-        val lowerCount = lowerChild.map{ _.count }.getOrElse(0L)
-        upperChild.flatMap{ _.findRankUpperBound(rank - lowerCount) }.orElse(Some(upperBound))
-      }
+  /**
+   * Precondition: if 0 <= rank < count
```
jnievelt: IMO remove the 'if' here

sid-kap: Fixed

sid-kap:
Random thought: we could easily optimize `findRankBounds` with the following changes:

* `@tailrec` the function (trivial)
* (maybe) drop the `require`, since we can prove that it is satisfied everywhere it is called anyway.

On second thought, this probably won't yield huge benefits -- it would only make a difference for deep QTrees, and the default QTree depth is only 9. 

The benefits are probably small, but given that it won't uglify the code, it might be a good idea.

ianoc:
Thats a great thing to benchmark sid, might be great, good to quantify tho

johnynek:
@sid-kap it would be interesting to quantify, but I doubt it will help. 1) the scala compiler already does tailrec when it can. Since the method is private (isn't it?), it should have already done it here. `@tailrec` just means error if you can't make it tailrec, it does not mean it won't happen without it. 2) require is probably free since if it is always true the branch predictor should notice that and assume it is true and keep on going.

sid-kap:
I ran some Caliper tests, and saw no difference between the tailrec'd version and the default version. Even at depth k=50, there was no difference. (scalac, it turns out, doesn't actually tail-rec this function by default.)

ianoc:

algebird-core/src/main/scala/com/twitter/algebird/QTree.scala
```diff
@@ -268,49 +268,32 @@ class QTree[@specialized(Int, Long, Float, Double) A] private[algebird] (
...
+  private def findRankBounds(rank: Long): (Double, Double) = {
+    require(0 <= rank && rank < count)
+
+    val (leftCount, rightCount) = mapChildrenWithDefault(0L)(_.count)
+    val parentCount = count - leftCount - rightCount
+
+    if (rank < leftCount) {
+      // Note that 0 <= rank < leftCount because of the require above.
+      // So leftCount > 0, so lowerChild is not None.
+      lowerChild.get.findRankBounds(rank)
```
sid-kap: Should we use lowerChildNullable and upperChildNullable in this method?

algebird-core/src/main/scala/com/twitter/algebird/QTree.scala
```diff
@@ -268,49 +268,32 @@ class QTree[@specialized(Int, Long, Float, Double) A] private[algebird] (
...
-        } else null
-
-        if (r2 == null) {
-          upperBound
-        } else r2
-      } else r
+  /**
+   * Precondition: 0 <= rank < count
+   */
+  private def findRankBounds(rank: Long): (Double, Double) = {
```
sid-kap: I noticed that @ianoc changed the Option[Double] that used to be here to java.lang.Double here because Option isn't specialized.
It looks like we have a similar issue in this implementation: Tuple2 is not specialized for Double.

Ideas? Or is it even worth it to optimize this method?

sid-kap: Never mind, Tuple2 is specialized for Double.

algebird-core/src/main/scala/com/twitter/algebird/QTree.scala
ianoc: If its called from the 'hot path' i.e. in plus or sumOption then yes worth doing it

sid-kap: It's only called in quantileBounds, so optimizing it is probably not necessary

ianoc: Yeah I wouldn't bother there, best to keep the less functional stuff out of the non-critical speed paths.  -- That said, it could be worth adding a benchmark on how the results are usually extracted from QTree, all the benchmarks now are for combinations. We had some big perf woes in HLL around some of them. So might be worth the experiment

sid-kap:
I added a benchmark for quantile bounds.
I didn't see any performance improvements from using nullables in findRankBounds, so I'll let that function be for now.
Can we merge this when green?

ianoc:
lgtm, merge when green

joshualande:
@sid-kap, thanks so much for working on this!

PR #475, opened 2015-07-30
Optimize QTree a bunch

ianoc:
1) Internally use nullable instances of QTree rather than Option[QTree].
2) Specialize the class
3) Pre build an array with the range values rather than calculating pow(2, level) on each instanciation

Performance numbers:

Develop::
[info] Benchmark                           (depthK)  (numElements)  Mode  Cnt         Score         Error  Units
[info] QTreeBenchmark.timePlusDouble              6          10000  avgt   15  49604065.294 ±  343946.724  ns/op
[info] QTreeBenchmark.timePlusLong                6          10000  avgt   15  54040361.060 ±  451564.394  ns/op
[info] QTreeBenchmark.timePlusUnit                6          10000  avgt   15  52363803.668 ± 1403481.464  ns/op
[info] QTreeBenchmark.timeSumOptionDouble         6          10000  avgt   15  18536550.306 ±  102450.054  ns/op
[info] QTreeBenchmark.timeSumOptionLong           6          10000  avgt   15  14265313.429 ±   80999.618  ns/op
[info] QTreeBenchmark.timeSumOptionUnit           6          10000  avgt   15  14109388.526 ±  139901.693  ns/op

Branch::

[info] Benchmark                           (depthK)  (numElements)  Mode  Cnt         Score        Error  Units
[info] QTreeBenchmark.timePlusDouble              6          10000  avgt   15  36130358.089 ± 344895.234  ns/op
[info] QTreeBenchmark.timePlusLong                6          10000  avgt   15  40619584.216 ± 213205.614  ns/op
[info] QTreeBenchmark.timePlusUnit                6          10000  avgt   15  35111406.421 ± 175324.238  ns/op
[info] QTreeBenchmark.timeSumOptionDouble         6          10000  avgt   15   8657591.363 ± 148336.269  ns/op
[info] QTreeBenchmark.timeSumOptionLong           6          10000  avgt   15   5310313.465 ±  88287.513  ns/op
[info] QTreeBenchmark.timeSumOptionUnit           6          10000  avgt   15   4735604.540 ±  45094.848  ns/op

### DIFF
--- a/algebird-core/src/main/scala/com/twitter/algebird/QTree.scala
+++ b/algebird-core/src/main/scala/com/twitter/algebird/QTree.scala
@@ -42,21 +42,39 @@ object QTree {
   /**
    * level gives a bin size of 2^level. By default the bin size is 1/65536 (level = -16)
    */
-  def apply[A](kv: (Double, A), level: Int = DefaultLevel): QTree[A] =
-    QTree(math.floor(kv._1 / math.pow(2.0, level)).toLong,
+  def apply[A](kv: (Double, A), level: Int = DefaultLevel): QTree[A] = {
+    val offset = math.floor(kv._1 / math.pow(2.0, level)).toLong
+    require(offset >= 0, "QTree can not accept negative values")
+
+    new QTree(offset,
       level,
       1,
       kv._2,
-      None,
-      None)
+      null,
+      null)
+  }
 
-  def apply[A](kv: (Long, A)): QTree[A] =
-    QTree(kv._1,
+  def apply[A](kv: (Long, A)): QTree[A] = {
+    require(kv._1 >= 0, "QTree can not accept negative values")
+
+    new QTree(kv._1,
       0,
       1,
       kv._2,
-      None,
-      None)
+      null,
+      null)
+  }
+
+  def apply[A](offset: Long,
+    level: Int,
+    count: Long,
+    sum: A, //the sum at just this node (*not* including its children)
+    lowerChild: Option[QTree[A]],
+    upperChild: Option[QTree[A]]): QTree[A] = {
+    require(offset >= 0, "QTree can not accept negative values")
+
+    new QTree(offset, level, count, sum, lowerChild.orNull, upperChild.orNull)
+  }
 
   /**
    * The common case of wanting an offset and sum for the same value
@@ -74,6 +92,12 @@ object QTree {
   def apply(k: Double): QTree[Double] = apply(k -> k)
 
   /**
+   * End user consumable unapply for QTree
+   */
+  def unapply[A](qtree: QTree[A]): Option[(Long, Int, Long, A, Option[QTree[A]], Option[QTree[A]])] =
+    Some((qtree.offset, qtree.level, qtree.count, qtree.sum, qtree.lowerChild, qtree.upperChild))
+
+  /**
    * If you are sure you only care about the approximate histogram
    * features of QTree, you can save some space by using QTree[Unit]
    */
@@ -84,11 +108,35 @@ object QTree {
    * level gives a bin size of 2^level. By default this is 1/65536 (level = -16)
    */
   def value(v: Double, level: Int = DefaultLevel): QTree[Unit] = apply(v -> (), level)
+
+  private[algebird] def mergePeers[@specialized(Int, Long, Float, Double) A](left: QTree[A], right: QTree[A])(implicit monoid: Monoid[A]): QTree[A] = {
+    assert(right.lowerBound == left.lowerBound, "lowerBound " + right.lowerBound + " != " + left.lowerBound)
+    assert(right.level == left.level, "level " + right.level + " != " + left.level)
+
+    new QTree[A](left.offset,
+      left.level, left.count + right.count,
+      monoid.plus(left.sum, right.sum),
+      mergeOptions(left.lowerChildNullable, right.lowerChildNullable),
+      mergeOptions(left.upperChildNullable, right.upperChildNullable))
+  }
+
+  private def mergeOptions[A](aNullable: QTree[A], bNullable: QTree[A])(implicit monoid: Monoid[A]): QTree[A] =
+    if (aNullable != null) {
+      if (bNullable != null) {
+        mergePeers(aNullable, bNullable)
+      } else aNullable
+    } else bNullable
+
+  private[algebird] val cachedRangeCacheSize: Int = 20
+  private[algebird] val cachedRangeLowerBound: Int = cachedRangeCacheSize * -1
+  private[algebird] val rangeLut: Array[Double] = (cachedRangeLowerBound until cachedRangeCacheSize).map { level =>
+    math.pow(2.0, level)
+  }.toArray[Double]
 }
 
 class QTreeSemigroup[A](k: Int)(implicit val underlyingMonoid: Monoid[A]) extends Semigroup[QTree[A]] {
   /** Override this if you want to change how frequently sumOption calls compress */
-  def compressBatchSize: Int = 25
+  def compressBatchSize: Int = 50
   def plus(left: QTree[A], right: QTree[A]) = left.merge(right).compress(k)
   override def sumOption(items: TraversableOnce[QTree[A]]): Option[QTree[A]] = if (items.isEmpty) None
   else {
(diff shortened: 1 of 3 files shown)


ianoc:
qtreebenchmark

algebird-core/src/main/scala/com/twitter/algebird/QTree.scala
```diff
@@ -84,9 +102,35 @@ object QTree {
...
+
+  private[algebird] val cachedRangeCacheSize: Int = 20
+  private[algebird] val cachedRangeLowerBound: Int = cachedRangeCacheSize * -1
+  private[algebird] val rangeLut: Array[Double] = (cachedRangeLowerBound until cachedRangeCacheSize).map { level =>
+    math.pow(2.0, level)
+  }.toArray[Double]
 }
 
-class QTreeSemigroup[A](k: Int)(implicit val underlyingMonoid: Monoid[A]) extends Semigroup[QTree[A]] {
+class QTreeSemigroup[@specialized(Int, Long, Float, Double, Unit) A](k: Int)(implicit val underlyingMonoid: Monoid[A]) extends Semigroup[QTree[A]] {
```
johnynek: do we need the Unit? There is only one value, so I assume boxing is fast. If we specialize on it, we get an additional copy of all the methods with type parameters, which adds to the jar size.

algebird-core/src/main/scala/com/twitter/algebird/QTree.scala
```diff
@@ -108,32 +152,45 @@ class QTreeSemigroup[A](k: Int)(implicit val underlyingMonoid: Monoid[A]) extend
...
 
-case class QTree[A](
+case class QTree[@specialized(Int, Long, Float, Double, Unit) A] private[algebird] (
   offset: Long, //the range this tree covers is offset*(2^level) ... (offset+1)*(2^level)
   level: Int,
   count: Long, //the total count for this node and all of its children
   sum: A, //the sum at just this node (*not* including its children)
-  lowerChild: Option[QTree[A]],
-  upperChild: Option[QTree[A]]) {
+  private[algebird]lowerChildNullable: QTree[A],
```
johnynek: do these need to be private, the constructor is already private.

algebird-core/src/main/scala/com/twitter/algebird/QTree.scala
```diff
@@ -108,32 +152,45 @@ class QTreeSemigroup[A](k: Int)(implicit val underlyingMonoid: Monoid[A]) extend
...
+
+  @annotation.tailrec
+  private[algebird] final def extendToLevel(n: Int)(implicit monoid: Monoid[A]): QTree[A] = {
     if (n <= level)
       this
     else {
       val nextLevel = level + 1
       val nextOffset = offset / 2
 
+      var l: QTree[A] = null
```
johnynek: can we add a comment that this was due to benchmarking and not to change it unless you have run benchmarks to show that it is faster.

algebird-core/src/main/scala/com/twitter/algebird/QTree.scala
```diff
@@ -84,9 +102,35 @@ object QTree {
...
+
+    QTree[A](left.offset,
+      left.level, left.count + right.count,
+      monoid.plus(left.sum, right.sum),
+      mergeOptions(left.lowerChildNullable, right.lowerChildNullable),
+      mergeOptions(left.upperChildNullable, right.upperChildNullable))
+  }
+
+  private def mergeOptions[@specialized(Int, Long, Float, Double, Unit) A](aNullable: QTree[A], bNullable: QTree[A])(implicit monoid: Monoid[A]): QTree[A] =
+    if (aNullable != null && bNullable != null) {
```
johnynek: we could reduce this a bit:
```scala
if (aNullable != null) {
  if (bNullable != null) {
    mergePees(aNullable, bNullable)
  }
  else aNullable
}
else bNullable
```

algebird-core/src/main/scala/com/twitter/algebird/QTree.scala
ianoc: I have a series of micro benchmarks that i'll push that show this was faster

algebird-core/src/main/scala/com/twitter/algebird/QTree.scala
ianoc: Since this is a case class this stops these being public vals' that someone can reference. nicer way for it?

algebird-core/src/main/scala/com/twitter/algebird/QTree.scala
ianoc: Probably not, will remove

algebird-core/src/main/scala/com/twitter/algebird/QTree.scala
```diff
@@ -108,32 +152,45 @@ class QTreeSemigroup[A](k: Int)(implicit val underlyingMonoid: Monoid[A]) extend
   }
 }
 
-case class QTree[A](
+case class QTree[@specialized(Int, Long, Float, Double, Unit) A] private[algebird] (
```
johnynek: it's a bummer to lose the case class matching, but don't we need to add:

```scala
  def offset
  def level
  def count
  def sum
```
or we are breaking the API?

For instance, Zak has some code to serialize QTrees, I'm not sure if that code can work without access to those values (pretty sure it can't).

I don't know how private interacts with case classes actually.

ianoc: I'm pretty sure the args here actually remain public, mima only reports the loss of some variants of this() constructors that I saw -- I'll run this through our internal stuff which should his tests before a merge

algebird-core/src/main/scala/com/twitter/algebird/QTree.scala
johnynek: Okay. I didn't know that. Related to my comments on case class with private constructors.

algebird-core/src/main/scala/com/twitter/algebird/QTree.scala
johnynek: what about the unapply method on the object? We could lose the sugar of case class, but we still get source compatibility by adding manual apply/unapply/toString/equals/hashCode

ianoc: I'm happy to do that, think it looks nicer than this use of private? we have a few apply's already, unapply we probably want to more match the old apply too. So it might wind up making some of it nicer to migrate to

avibryant:
A+++ would write unoptimized code for ianoc to make much better again.

algebird-core/src/main/scala/com/twitter/algebird/QTree.scala
```diff
@@ -155,7 +243,7 @@ case class QTree[A](
       ancestorLevel += 1
       offsetDiff >>= 1
     }
-    ancestorLevel.max(level).max(other.level)
+    ancestorLevel.max(_level).max(other.level)
```
sid-kap: Maybe use `scala.math.max` here to prevent boxing/unboxing?

johnynek: I don't think these box. I think they use AnyVal enrichments:

https://github.com/scala/scala/blob/v2.10.3/src/library/scala/runtime/RichInt.scala#L1

https://github.com/scala/scala/blob/v2.10.3/src/library/scala/runtime/RichDouble.scala

algebird-core/src/main/scala/com/twitter/algebird/QTree.scala
```diff
@@ -197,36 +268,49 @@ case class QTree[A](
...
     // get is safe below, because findRankLowerBound only returns
     // None if rank > count, but due to construction rank <= count
-    (findRankLowerBound(rank).get, findRankUpperBound(rank).get)
+    (findRankLowerBound(rank), findRankUpperBound(rank))
   }
 
-  private def findRankLowerBound(rank: Long): Option[Double] =
-    if (rank > count)
-      None
+  private def findRankLowerBound(rank: Long): java.lang.Double =
```
johnynek: I guess this is faster than `Option[Double]` because Option is not specialized.

ianoc: Yep, private so figured safe internal thing to do

ianoc:
@avibryant Any time! I never have the time to play with new algo's. But once our jobs go slow lots of motivation to improve things! :)

PR #478, opened 2015-08-04
Dealing with probabilistic tests

sid-kap:
This is a rough sketch of a way we can deal with flaky/probabilistic tests.

I defined a trait ApproximateProperty that allows us to describe approximate properties, and, as an example, implemented a working ApproximateProperty test for CMS.

In detail:

* Defined the ApproximateProperty trait to describe a relationship between
    * an `Exact` type
    * an `Approx` type which approximates the behavior of Exact
    * a behavior that both `Exact` and `Approx` implement. 
        * `Exact` should be able to take something of type `Input` and produce something of type `Result`
        * `Approx` should be able to take something of type `Input` and produce something of type `Approximate[Result]`
    * some generators/glue functions (exactGenerator, makeApproximate, inputGenerator) that determine how to create an instance of Exact, how to convert an instance of Exact to an instance of Approximate, and how to create an Input
* Defined the `ApproximateProperty.toProp` function that converts an `ApproximateProperty` to an `org.scalacheck.Prop` that runs the ApproximateProperty test many times and fails if the number of failed tests exceeds a certain threshold. (using Hoeffding's inequality)

Comments/suggestions?


algebird-test/src/main/scala/com/twitter/algebird/ApproximateProperty.scala
```diff
@@ -0,0 +1,90 @@
...
+          val exactResult = a.exactResult(exact, input)
+          val success = if (approxResult.boundsContain(exactResult)) 1 else 0
+          (success, approxResult.probWithinBounds)
+        }
+      }
+    val monoid = implicitly[Monoid[(Int, Double)]]
+    monoid.sum(list)
+  }
+
+  def toProp(a: ApproximateProperty, objectReps: Int, inputReps: Int, falsePositiveRate: Double): Prop =
```
sid-kap: A better name for `falsePositiveRate` might be "false failure rate"

algebird-test/src/main/scala/com/twitter/algebird/ApproximateProperty.scala
```diff
@@ -0,0 +1,90 @@
...
+        require(0 <= falsePositiveRate && falsePositiveRate <= 1)
+
+        val (successes, sumOfProbabilities) = successesAndProbabilities(a, objectReps, inputReps)
+        val n = objectReps * inputReps
+
+        // Computed from Hoeffding's inequality, might be inaccurate
+        // TODO Make sure this is correct
+        val diff = scala.math.sqrt(n * scala.math.log(falsePositiveRate) / -2)
+
+        val args = List(("Successes", successes),
```
sid-kap: scalacheck prints these args when the test is run, so that we can see the number of successes, etc. when we run the tests

algebird-test/src/main/scala/com/twitter/algebird/ApproximateProperty.scala
```diff
@@ -0,0 +1,90 @@
...
+          ("Expected successes", sumOfProbabilities),
+          ("Requires successes", sumOfProbabilities - diff))
+          .map {
+            case (name, value) =>
+              Prop.Arg(name, value, 0, value, Pretty.prettyAny(value), Pretty.prettyAny(value))
+          }
+
+        val success = if (successes >= (sumOfProbabilities - diff)) Prop.Proof else Prop.False
+
+        Prop.Result(success, args = args)
```
sid-kap: scalacheck prints these args when the test is run, so that we can see the number of successes, etc. when we run the tests

project/Build.scala
```diff
@@ -155,8 +155,10 @@ object AlgebirdBuild extends Build {
   )
 
   lazy val algebirdTest = module("test").settings(
+    testOptions in Test ++= Seq(Tests.Argument(TestFrameworks.ScalaCheck, "-verbosity", "4"),
+                                Tests.Argument(TestFrameworks.ScalaCheck, "-minSuccessfulTests", "1")),
```
sid-kap: TO FIX
By default, scalacheck calls `check` on a property 100 times. However, we want it to call `check` on our property only once, so that the approximate property can run all `n` tests at once and decide whether there were too many failures.

Hopefully there is a way to tell scalacheck how many times to test each property, so that it can run the approximate property tests once, and the normal property tests 100 times.

algebird-test/src/main/scala/com/twitter/algebird/ApproximateProperty.scala
```diff
@@ -0,0 +1,96 @@
...
+import org.scalacheck.util.Pretty
+
+trait ApproximateProperty {
+  type Exact
+  type Approx
+  type Input
+  type Result
+
+  def exactGenerator: Gen[Exact]
+  def inputGenerator(e: Exact): Gen[Input]
```
johnynek: what is this exact value corresponding to?

algebird-test/src/main/scala/com/twitter/algebird/ApproximateProperty.scala
```diff
@@ -0,0 +1,96 @@
...
+  def toProp(a: Iterable[ApproximateProperty], objectReps: Int, inputReps: Int, falsePositiveRate: Double): Prop = {
+    require(0 <= falsePositiveRate && falsePositiveRate <= 1)
+
+    val monoid = implicitly[Monoid[(Int, Double, List[String])]]
+    val (successes, sumOfProbabilities, _) = monoid.sum(a.map { approximateProp =>
+      successesAndProbabilities(approximateProp, objectReps, inputReps)
+    })
+    val n = objectReps * inputReps
+
+    println("Foo")
```
johnynek: what's going on here?

algebird-test/src/main/scala/com/twitter/algebird/ApproximateProperty.scala
```diff
@@ -0,0 +1,96 @@
...
+
+object ApproximateProperty {
+
+  /**
+   *  Generates a list of exactly n Ts.
+   *  Useful because `Gen.listOfN(n, gen).sample` gives us Option[List[T]],
+   *  while we often want List[T].
+   */
+  private def genListOf[T](n: Int, gen: Gen[T]): List[T] = {
+    Gen.listOfN(n, gen).sample match {
```
johnynek: I'm concerned that this is not using the Gen.Parameters, so I'm afraid that when scalacheck gives us an example input that fails a test, we can't check it at the REPL. Is there some way to plumb through the Random instance from Gen.Parameters (I think has it), so that we are not using a time-seeded Random?

algebird-test/src/main/scala/com/twitter/algebird/ApproximateProperty.scala
```diff
@@ -0,0 +1,96 @@
+package com.twitter.algebird
+
+import org.scalacheck.{ Gen, Prop }
+import org.scalacheck.util.Pretty
+
+trait ApproximateProperty {
```
johnynek: this is really nice and quite well designed. Good work.

algebird-test/src/main/scala/com/twitter/algebird/ApproximateProperty.scala
sid-kap: Exact is useful here so that we can generate appropriate test inputs. 

For example, suppose we're testing CMS[Long]. We need to generate some inputs so that we can query `exactList.frequency(x)` and `cms.frequency(x)` and compare them. 
We could simply query using random Long values. However, most of the random Long values we generate would probably not be in the list, so these test cases would not be very useful. If we know the exact value, we can choose random elements in `exactList` as our test cases.

I'm not sure if this is the best way to do this.

algebird-test/src/main/scala/com/twitter/algebird/ApproximateProperty.scala
sid-kap: Added a comment

algebird-test/src/main/scala/com/twitter/algebird/ApproximateProperty.scala
```diff
@@ -0,0 +1,96 @@
...
+  def toProp(a: ApproximateProperty, objectReps: Int, inputReps: Int, falsePositiveRate: Double): Prop =
+    new Prop {
+      def apply(params: Gen.Parameters) = {
+        require(0 <= falsePositiveRate && falsePositiveRate <= 1)
+
+        val (successes, sumOfProbabilities, exacts) = successesAndProbabilities(a, objectReps, inputReps)
+        val n = objectReps * inputReps
+
+        // Computed from Hoeffding's inequality, might be inaccurate
+        // TODO Make sure this is correct
```
johnynek: let's link to:
https://en.wikipedia.org/wiki/Hoeffding%27s_inequality

and add the result;
```
When Prob(X = 1) >= p, then

Prob(sum(X) <= (p - eps)*n) <= exp(-2*eps*eps*n)

If we want that Prob <= fprate, then it is sufficient that, fprate = exp(-2*eps*eps*n) or:
eps = math.sqrt( - math.log(fprate)/(2.0 * n) )
```
So, I think there is a slight error in your version (n is in the numerator in yours).

project/Build.scala
sid-kap: https://github.com/rickynils/scalacheck/issues/187
Looks like the Scalacheck guy might add this feature in Scalacheck 2.0.0 (in a long time)

Any ideas for a workaround?

algebird-test/src/main/scala/com/twitter/algebird/ApproximateProperty.scala
jnievelt: diff = n * fprate, so it's correct in the numerator.

My only concern here is that the formula is for the Bernoulli version of Hoeffding's.  Since we're taking separate probability values from each result, we aren't guaranteed that they'll be uniform (very possibly they won't be).  But maybe it's "close enough."

algebird-test/src/main/scala/com/twitter/algebird/ApproximateProperty.scala
sid-kap: This might not a problem since I'm not using scalacheck's error reporting mechanism. (I had to write my own hacky error reporter using scalacheck args.)
Here's an example of a failure:
https://travis-ci.org/twitter/algebird/jobs/74179669#L2565
It always prints successes, expected successes, and required successes. When a test fails, it prints some details from the first few failures.

algebird-test/src/main/scala/com/twitter/algebird/ApproximateProperty.scala
sid-kap: The general case (https://en.wikipedia.org/wiki/Hoeffding%27s_inequality#General_case) is
```
P(sum(X) - sum(E[X]) >= nt) <= exp(-2*n*t^2)
```
(here I used `sum(X) = X_1 + ... + X_n` instead of `Xbar = 1/n (X_1 + ... + X_N)`)
So we have
```
   exp(-2*n*t^2) = fprate
=> t = sqrt(-log p / (2*n) )
```

sid-kap: Whoops, I read Joe's comment again, and he's right. 
diff = n * t, so we need `diff = sqrt(-n * log(fprate) / (2 * n))`.

project/Build.scala
sid-kap: If the scalacheck maintainer accepts my PR (https://github.com/rickynils/scalacheck/pull/188), then we'll be able to do the following:
- Ensure that ApproximateProperty props don't go in the same Properties class as normal props.
- For every Properties class that contains ApproximateProperty props, override the minSuccessfulTests to 1 (the default is 100).

algebird-test/src/main/scala/com/twitter/algebird/ApproximateProperty.scala
johnynek: I agree. I overlooked the `n * t` on the outside.

I think the bound applies perfectly, with the assumption that our `Random` class is a real RNG. That is obviously not true, and in fact it is pseudo-random, but I don't see how that is escapable. We could beef up the RNG used to be stronger (but certainly slower), but that might require some changes to scalacheck, not sure.

algebird-test/src/main/scala/com/twitter/algebird/ApproximateProperty.scala
DanielleSucher: That link just gives me "This is not an active repository". Are there some creds needed to see it? Can you post a screenshot?

version.sbt
```diff
@@ -1 +1 @@
-version in ThisBuild := "0.11.0"
+version in ThisBuild := "0.11.0 -SNAPSHOT"
```
DanielleSucher: Why bump to a snapshot version here?

sid-kap: Whoops, I didn't mean to check that change in. I'll remove it

algebird-test/src/main/scala/com/twitter/algebird/ApproximateProperty.scala
ianoc: It actually looks like the travis settings somehow got broken for algebird, it seems to be disabled in travis. /cc @caniszczyk

version.sbt
ianoc: We probably should actually swap all of them to work like this. We've seen some funky errors and mixed classpath's if you have a published version in the version.sbt. Not sure if its a new sbt thing or what. But can be a different PR

algebird-test/src/main/scala/com/twitter/algebird/ApproximateProperty.scala
johnynek: but why couldn't we do something like:

```scala
// I need an exact value for my implementation:
def needExact(e: Exact): Gen[Input]
// but I could still implement this:
def inputGenerator: Gen[Input] = exactGenerator.flatMap(needExact) 
```
Am I missing something? Putting this in the trait as is seems to be constraining more than is needed. Why can't we just get by with `def inputGenerator: Gen[Input]`

algebird-test/src/main/scala/com/twitter/algebird/ApproximateProperty.scala
```diff
@@ -0,0 +1,137 @@
...
+
+  def makeApproximate(e: Exact): Approx
+
+  def exactResult(e: Exact, i: Input): Result
+  def approximateResult(a: Approx, i: Input): GeneralizedApproximate[Result]
+}
+
+trait GeneralizedApproximate[T] {
+  def withProb: Double
+  def contains(t: T): Boolean
```
johnynek: seems like what we really want is
```
trait ApproximateSet[S, -T] {
  def contains(set: S, t: T): ApproximateBoolean 
}
```
and use that as a typeclass rather than implicit conversions.
```
object ApproximateSet {
  def contains[S, T](s: S, t: T)(implicit as: ApproximateSet[S, T]): ApproximateBoolean = as.contains(s, t)
  implicit def fromApproximate[N: Numeric]: ApproximateSet[Approximate[N], N] = new ApproximateSet ...
}
```

johnynek: Actually, what you are trying to do here is add a common interface to ApproximateBoolean and `Approximate[N]` right?

Maybe just do that directly rather than through implicit conversion.

sid-kap: Yeah, I'm trying to make a common interface to ApproximateBoolean and Approximate[N].

I like the idea of having a trait ApproximateSet[T] and having 
```scala
ApproximateBoolean extends ApproximateSet[Boolean]
```
and 
```scala
Approximate[T] extends ApproximateSet[T]
```

version.sbt
sid-kap: Yeah, I keep getting a "Unresolved dependencies: algebird-test" error when I don't use "-SNAPSHOT". Is this happening to other people too?

sid-kap:
@ianoc everything in this PR uses Scalacheck foralls and Properties classes. Do we want to rewrite it in terms of Scalatest foralls and property tests?

ianoc:
Looks like this doesn't merge cleanly to develop @sid-kap can you merge in develop to your branch and resolve the conflicts?

ianoc:
Nope, lets just get this in and stable, once Oscar is happy I think we should try get it in. Can worry about that other stuff later/maybe never.

sid-kap:
Ok sounds good.
The maintainer of scalacheck finally merged in the PR that I sent. Now we just have to wait for him to do a new release. Once that happens, I can try to merge this in.

algebird-test/src/test/scala/com/twitter/algebird/CountMinSketchTest.scala
```diff
@@ -228,6 +228,97 @@ class CMSBigIntTest extends CMSTest[BigInt]
...
+  type Exact = Vector[K]
+  type Approx = CMS[K]
+
+  type Input = K
+  type Result = Long
+
+  def makeApproximate(e: Exact) = CmsProperty.makeApproximate(e)
+  def inputGenerator(e: Vector[K]): Gen[K] = Gen.oneOf(e)
+
+  def exactResult(list: Vector[K], key: K) = list.filter(_ == key).length
```
johnynek: can we call `list` `vec` instead? Also, `vec.count(_ == key)` is going to be faster here since it does not materialize a second vector.

version.sbt
```diff
@@ -1 +1 @@
-version in ThisBuild := "0.11.0"
+version in ThisBuild := "0.11.0-SNAPSHOT"
```
johnynek: should we be bumping the version?

project/Build.scala
johnynek: So, all of our tests only run with 1 sample now?

johnynek:
A couple of comments. Also note that this does not merge cleanly, so you'll need to merge in develop.

sid-kap:
Scalacheck just released the new version with the feature that I requested. So I'll try to bring in the new dependency and finish up this PR sometime next week.

sid-kap:
By the way, I don't think we should use the Chernoff-Hoeffing bound. A few weeks of studying concentration inequalities in Randomized Algorithms has taught me not to use the Hoeffding bound for a binomial with very small or very large p -- this bound doesn't depend on p, and therefore doesn't take into account the fact that the binomial concentrates much better for very small/very large p.

Using a weak bound would result in overly lax tests -- for example, suppose we are running 100 trials with 0.975 probability of success, and we want to fail at most 0.01 of the time. The true threshold for the binomial (`qbinom(0.01, 100, 0.975)`) is 93, but the Chernoff-Hoeffding bound gives us `97.5 - sqrt(-2 * log(0.01) / 100) = 82.3`. This gets worse for properties with a 0.99 success probability -- the true value is 96, and the Hoeffding bound gives us 83.8.

For fun: we could use a Bernstein-type inequality (since the binomial is sub-gamma), which would give us a cutoff of `97.5 - sqrt(-4*0.975*(1-0.975)*log(0.01)) = 90.7` for 0.975 success probability, and `99 - sqrt(-4*0.99*(1-0.99)*log(0.01)) = 94.7` for the 0.99 success probability. This bound is essentially tight because it uses the Gaussian tail bound (e^(-x^2)) instead of an exponential bound (e^(-x)). For proof, see the last page of http://www.cs.utexas.edu/~ecprice/courses/randomized/notes/lec6.pdf (Note that these notes assume that p << 1, so it uses the approximation sqrt(p(1-p)) ~ sqrt(p), which is not the case here.)

Or we could be lame and use the inverse binomial function...

sid-kap:
Do we want to try to merge this in? I've updated Scalacheck, so now it runs each test the correct number of times.

The only things I think need to be addressed are
1) Maybe make the bounds tighter, as mentioned above
2) How to enforce that people use these classes correctly

Concerning 2: I made a class called ApproximateProperties that extends Properties but has a default of running each test once instead of 100 times. All Properties classes that have approximate properties should extend from this, and should contain only approximate properties. (We need to run approximate properties only once because when the ApproximateProperty check code is called once, it generates many (say, 100) inputs and succeeds if enough of those 100 succeed.)

Is there a clean way to enforce, using the type system, that only ApproximatePropertys can be put in the ApproximateProperties class? I would not want to emulate the Scalacheck Properties class because that class is implemented using messy mutable structures (every time you call `property("foo") = `, it appends the property to a mutable list of properties). Maybe we could make a function that takes a `List<ApproximateProperty>`? This would work but it might lead to messier code (we would not be able to take advantage of the `property("foo" =` DSL).

johnynek:
@sid-kap Thanks! I do very much want to merge this, I just haven't looked back at it since the scalacheck upgrade. I'll review. Thanks for not letting it drop! Appreciated.

erikerlandson:
I've had some success comparing algorithms involving random behaviors by using the Kolmogorov-Smirnov test, where you can test whether two distributions are different (or, in the inverse case, whether they are *not* different).  If you can define (either in closed-form, or via sampling) a reference of "correct" behavior, you can then run your test some number of times and run the KS-test between test and reference.

Worth noting that with random behaviors, there is generally no way to *guarantee* that your random test will never violate, though it is possible to put things on a theoretical footing that fails less frequently.

https://github.com/apache/spark/pull/2455
http://erikerlandson.github.io/blog/2014/09/11/faster-random-samples-with-gap-sampling/

johnynek:
@ianoc what do you think? I'm happy to merge this now. I think it's a big improvement, and we let it sit too long.

ianoc:
I'm in favor of pushing ahead and getting it in, if there are futher improvements to be made we can do it later

PR #480, opened 2015-08-05
Upgrade scalacheck and scalatest

DanielleSucher:
Supersedes https://github.com/twitter/algebird/pull/431 as per @ianoc's suggestion, now with an even newer version of scalacheck and all tests passing locally (at least).


ianoc:
Excellent, thanks!

PR #481, opened 2015-08-10
Add sanity requirement for Approximate

sid-kap:
Ensure that min <= estimate <= max


algebird-core/src/main/scala/com/twitter/algebird/Approximate.scala
```diff
@@ -68,20 +68,25 @@ object ApproximateBoolean {
...
+   */
   def ~(v: N): Boolean = boundsContain(v)
-  /*
-    * Contract is:
-    * Prob(boundsContain(estimate)) >= probWithinBounds
-    */
+
+  /**
+   * Contract is:
+   * Prob(boundsContain(estimate)) >= probWithinBounds
```
johnynek: this comment looks like it should be on boundsContain. Can you move it?

sid-kap: Done

johnynek:
merge when green

PR #483, opened 2015-08-14
Cuber/roller macros

sid-kap:
Related to https://github.com/twitter/scalding/issues/1420

### DIFF
--- a/algebird-core/src/main/scala/com/twitter/algebird/MapAlgebra.scala
+++ b/algebird-core/src/main/scala/com/twitter/algebird/MapAlgebra.scala
@@ -18,6 +18,8 @@ package com.twitter.algebird
 import scala.collection.{ Map => ScMap }
 import scala.collection.mutable.{ Map => MMap }
 
+import com.twitter.algebird.macros.{ Cuber, Roller }
+
 trait MapOperations[K, V, M <: ScMap[K, V]] {
   def add(oldMap: M, kv: (K, V)): M
   def remove(oldMap: M, k: K): M
@@ -222,4 +224,31 @@ object MapAlgebra {
 
   def dot[K, V](left: Map[K, V], right: Map[K, V])(implicit mring: Ring[Map[K, V]], mon: Monoid[V]): V =
     Monoid.sum(mring.times(left, right).values)
+
+  def cube[K, V](it: TraversableOnce[(K, V)])(implicit c: Cuber[K]): Map[c.K, TraversableOnce[V]] = {
+    val sg = new Semigroup[Iterable[V]] {
+      def plus(x: Iterable[V], y: Iterable[V]) = x ++ y
+    }
+    cubeSum(it.map { case (k, v) => (k, Iterable(v)) })(c, sg)
+  }
+
+  def cubeSum[K, V](it: TraversableOnce[(K, V)])(implicit c: Cuber[K], sg: Semigroup[V]): Map[c.K, V] =
+    sumByKey(it.toIterator.flatMap { case (k, v) => c(k).map((_, v)) })
+
+  def cubeAggregate[T, K, U, V](it: TraversableOnce[T], agg: Aggregator[T, U, V])(fn: T => K)(implicit c: Cuber[K]): Map[c.K, V] =
+    sumByKey(it.toIterator.flatMap { t => c(fn(t)).map((_, agg.prepare(t))) })(agg.semigroup).mapValues(agg.present)
+
+  def rollup[K, V](it: TraversableOnce[(K, V)])(implicit r: Roller[K]): Map[r.K, TraversableOnce[V]] = {
+    val sg = new Semigroup[Iterable[V]] {
+      def plus(x: Iterable[V], y: Iterable[V]) = x ++ y
+    }
+    rollupSum(it.map { case (k, v) => (k, Iterable(v)) })(r, sg)
+  }
+
+  def rollupSum[K, V](it: TraversableOnce[(K, V)])(implicit r: Roller[K], sg: Semigroup[V]): Map[r.K, V] =
+    sumByKey(it.toIterator.flatMap { case (k, v) => r(k).map((_, v)) })
+
+  def rollupAggregate[T, K, U, V](it: TraversableOnce[T], agg: Aggregator[T, U, V])(fn: T => K)(implicit r: Roller[K]): Map[r.K, V] =
+    sumByKey(it.toIterator.flatMap { t => r(fn(t)).map((_, agg.prepare(t))) })(agg.semigroup).mapValues(agg.present)
+
 }
--- a/algebird-core/src/main/scala/com/twitter/algebird/MapAlgebra.scala
+++ b/algebird-core/src/main/scala/com/twitter/algebird/MapAlgebra.scala
@@ -226,29 +226,47 @@ object MapAlgebra {
     Monoid.sum(mring.times(left, right).values)
 
   def cube[K, V](it: TraversableOnce[(K, V)])(implicit c: Cuber[K]): Map[c.K, TraversableOnce[V]] = {
-    val sg = new Semigroup[Iterable[V]] {
-      def plus(x: Iterable[V], y: Iterable[V]) = x ++ y
+    val map: collection.mutable.Map[c.K, List[V]] = collection.mutable.Map[c.K, List[V]]()
+    it.toIterator.foreach {
+      case (k, v) =>
+        c(k).foreach { ik =>
+          map.get(ik) match {
+            case Some(vs) => map += ik -> (v :: vs)
+            case None => map += ik -> List(v)
+          }
+        }
     }
-    cubeSum(it.map { case (k, v) => (k, Iterable(v)) })(c, sg)
+    map.foreach { case (k, v) => map(k) = v.reverse }
+    new MutableBackedMap(map)
   }
 
   def cubeSum[K, V](it: TraversableOnce[(K, V)])(implicit c: Cuber[K], sg: Semigroup[V]): Map[c.K, V] =
     sumByKey(it.toIterator.flatMap { case (k, v) => c(k).map((_, v)) })
 
   def cubeAggregate[T, K, U, V](it: TraversableOnce[T], agg: Aggregator[T, U, V])(fn: T => K)(implicit c: Cuber[K]): Map[c.K, V] =
-    sumByKey(it.toIterator.flatMap { t => c(fn(t)).map((_, agg.prepare(t))) })(agg.semigroup).mapValues(agg.present)
+    sumByKey(it.toIterator.flatMap { t => c(fn(t)).map((_, agg.prepare(t))) })(agg.semigroup)
+      .map { case (k, v) => (k, agg.present(v)) }
 
   def rollup[K, V](it: TraversableOnce[(K, V)])(implicit r: Roller[K]): Map[r.K, TraversableOnce[V]] = {
-    val sg = new Semigroup[Iterable[V]] {
-      def plus(x: Iterable[V], y: Iterable[V]) = x ++ y
+    val map: collection.mutable.Map[r.K, List[V]] = collection.mutable.Map[r.K, List[V]]()
+    it.toIterator.foreach {
+      case (k, v) =>
+        r(k).foreach { ik =>
+          map.get(ik) match {
+            case Some(vs) => map += ik -> (v :: vs)
+            case None => map += ik -> List(v)
+          }
+        }
     }
-    rollupSum(it.map { case (k, v) => (k, Iterable(v)) })(r, sg)
+    map.foreach { case (k, v) => map(k) = v.reverse }
+    new MutableBackedMap(map)
   }
 
   def rollupSum[K, V](it: TraversableOnce[(K, V)])(implicit r: Roller[K], sg: Semigroup[V]): Map[r.K, V] =
     sumByKey(it.toIterator.flatMap { case (k, v) => r(k).map((_, v)) })
 
   def rollupAggregate[T, K, U, V](it: TraversableOnce[T], agg: Aggregator[T, U, V])(fn: T => K)(implicit r: Roller[K]): Map[r.K, V] =
-    sumByKey(it.toIterator.flatMap { t => r(fn(t)).map((_, agg.prepare(t))) })(agg.semigroup).mapValues(agg.present)
+    sumByKey(it.toIterator.flatMap { t => r(fn(t)).map((_, agg.prepare(t))) })(agg.semigroup)
+      .map { case (k, v) => (k, agg.present(v)) }
 
 }
--- a/algebird-core/src/main/scala/com/twitter/algebird/MapAlgebra.scala
+++ b/algebird-core/src/main/scala/com/twitter/algebird/MapAlgebra.scala
@@ -225,7 +225,7 @@ object MapAlgebra {
   def dot[K, V](left: Map[K, V], right: Map[K, V])(implicit mring: Ring[Map[K, V]], mon: Monoid[V]): V =
     Monoid.sum(mring.times(left, right).values)
 
-  def cube[K, V](it: TraversableOnce[(K, V)])(implicit c: Cuber[K]): Map[c.K, TraversableOnce[V]] = {
+  def cube[K, V](it: TraversableOnce[(K, V)])(implicit c: Cuber[K]): Map[c.K, List[V]] = {
     val map: collection.mutable.Map[c.K, List[V]] = collection.mutable.Map[c.K, List[V]]()
     it.toIterator.foreach {
       case (k, v) =>
@@ -247,7 +247,7 @@ object MapAlgebra {
     sumByKey(it.toIterator.flatMap { t => c(fn(t)).map((_, agg.prepare(t))) })(agg.semigroup)
       .map { case (k, v) => (k, agg.present(v)) }
 
-  def rollup[K, V](it: TraversableOnce[(K, V)])(implicit r: Roller[K]): Map[r.K, TraversableOnce[V]] = {
+  def rollup[K, V](it: TraversableOnce[(K, V)])(implicit r: Roller[K]): Map[r.K, List[V]] = {
     val map: collection.mutable.Map[r.K, List[V]] = collection.mutable.Map[r.K, List[V]]()
     it.toIterator.foreach {
       case (k, v) =>
--- a/algebird-core/src/main/scala/com/twitter/algebird/macros/Cuber.scala
+++ b/algebird-core/src/main/scala/com/twitter/algebird/macros/Cuber.scala
@@ -72,23 +72,35 @@ object Cuber {
     ensureCaseClass(c)
 
     val params = getParams(c)
-    if (params.length > 22)
+    val arity = params.length
+    if (arity > 22)
       c.abort(c.enclosingPosition, s"Cannot create Cuber for $T because it has more than 22 parameters.")
-    if (params.length == 0)
+    if (arity == 0)
       c.abort(c.enclosingPosition, s"Cannot create Cuber for $T because it has no parameters.")
 
-    val tupleName = newTypeName(s"Tuple${params.length}")
+    val tupleName = newTypeName(s"Tuple${arity}")
     val types = params.map { param => tq"_root_.scala.Option[${param.returnType}]" }
 
-    val fors = params.map { param =>
-      fq"""${param.name.asInstanceOf[c.TermName]} <- _root_.scala.Seq(_root_.scala.Some(in.${param}), _root_.scala.None)"""
+    val somes = params.zip(Stream.from(1)).map {
+      case (param, index) =>
+        val name = newTermName(s"some$index")
+        q"val $name = _root_.scala.Some(in.$param)"
     }
-    val names = params.map { param => param.name.asInstanceOf[c.TermName] }
+
+    val options = (1 to arity).map { index =>
+      val some = newTermName(s"some$index")
+      q"if (((1 << ${index - 1}) & i) == 0) { _root_.scala.None } else { $some }"
+    }
 
     val cuber = q"""
     new _root_.com.twitter.algebird.macros.Cuber[${T}] {
       type K = $tupleName[..$types]
-      def apply(in: ${T}): _root_.scala.Seq[K] = for (..$fors) yield new K(..$names)
+      def apply(in: ${T}): _root_.scala.Seq[K] = {
+        ..$somes
+        (0 until (1 << $arity)).map { i =>
+          new K(..$options)
+        }
+      }
     }
     """
     c.Expr[Cuber[T]](cuber)
--- a/algebird-core/src/main/scala/com/twitter/algebird/macros/Cuber.scala
+++ b/algebird-core/src/main/scala/com/twitter/algebird/macros/Cuber.scala
@@ -32,9 +32,9 @@ trait Cuber[I] {
 }
 
 /**
- * For a tuple N produces a result with (N + 1) tuples each of arity N
+ * Given a TupleN, produces a sequence of (N + 1) tuples each of arity N
  * such that, for all k from 0 to N, there is a tuple with k Somes
- * followed by (N - k) nones.
+ * followed by (N - k) Nones.
  *
  * This is useful for comparing some metric across multiple layers of
  * some hierarchy.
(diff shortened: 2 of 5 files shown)


algebird-core/src/main/scala/com/twitter/algebird/macros/Cuber.scala
```diff
@@ -0,0 +1,81 @@
+package com.twitter.algebird.macros
+
+import scala.language.experimental.{ macros => sMacros }
+import scala.reflect.macros.Context
+import scala.reflect.runtime.universe._
+
+// import com.twitter.algebird._
```
ianoc: kill comment

algebird-core/src/main/scala/com/twitter/algebird/macros/Cuber.scala
```diff
@@ -0,0 +1,81 @@
...
+
+trait Cuber[I] {
```
ianoc: comments on cuber and roller traits would be great

algebird-core/src/main/scala/com/twitter/algebird/macros/Cuber.scala
```diff
@@ -0,0 +1,81 @@
...
+  def apply(in: I): TraversableOnce[K]
+}
+
+trait Roller[I] {
+  type K
+  def apply(in: I): TraversableOnce[K]
+}
+
+object Cuber {
+  def cuber[T]: Cuber[T] = macro cuberImpl[T]
```
ianoc: implicit i think you want here

algebird-core/src/main/scala/com/twitter/algebird/macros/Cuber.scala
```diff
@@ -0,0 +1,81 @@
...
+      type K = $tupleName[..$types]
+      def apply(in: ${T}): _root_.scala.Seq[K] = for (..$fors) yield new K(..$names)
+    }
+    """
+    c.Expr[Cuber[T]](cuber)
+  }
+}
+
+object Roller {
+  def roller[T]: Roller[T] = macro rollerImpl[T]
```
ianoc: implicit here too

algebird-test/src/test/scala/com/twitter/algebird/macros/CuberMacroTest.scala
```diff
@@ -0,0 +1,67 @@
+package com.twitter.algebird.macros
+
+import com.twitter.algebird._
+import com.twitter.algebird.macros._
+import com.twitter.algebird.macros.ArbitraryCaseClassMacro.arbitrary
+
+import org.scalacheck.{ Properties, Arbitrary }
+import org.scalacheck.Prop.forAll
+
+object CuberRollerProperties extends Properties("Cuber and roller macro") {
```
ianoc: can we use scalatest here instead of scalacheck so it runs with the other tests? It will get counted towards the metrics then of how many tests we've ran overall

algebird-test/src/test/scala/com/twitter/algebird/macros/CuberMacroTest.scala
```diff
@@ -0,0 +1,67 @@
...
+  import BaseProperties._
+
+  implicit val arbitraryFoo: Arbitrary[Foo] = arbitrary[Foo]
+  implicit val arbitraryBar: Arbitrary[Bar] = arbitrary[Bar]
+  implicit val arbitraryBaz: Arbitrary[Baz] = arbitrary[Baz]
+
+  case class Foo(a: Int, b: Option[String], c: Long)
```
ianoc: can we add a test case here of a nested tuple or case class to show it not recursing

ianoc:
Whats the definition of a cube vs a rollup?

sid-kap:
Cube produces all 2^n subsets of the data.  Rollup is useful for hierarchical data, and produces only n+1 subsets, where each field is included only if all the previous (more general) fields are included.

@joshualande has a great blog post on this: http://joshualande.com/cube-rollup-pig-data-science/

ianoc:
Ok cool, well sounds like we should add some decent comments to the code and links to that post

sid-kap:
We should probably add enhancements to Map and List so that people can do
```scala 
val map1: Map[(Option[Int], Option[Gender]), Person] = 
  people.cubeBy { p => (p.age, p.gender) } 
val map2: Map[(Option[Int], Option[Gender]), Person] = 
  people.rollupBy { p => (p.age, p.gender) } 
```

algebird-core/src/main/scala/com/twitter/algebird/macros/Cuber.scala
```diff
@@ -0,0 +1,139 @@
...
+
+    val fors = params.map { param =>
+      fq"""${param.name.asInstanceOf[c.TermName]} <- _root_.scala.Seq(_root_.scala.Some(in.${param}), _root_.scala.None)"""
+    }
+    val names = params.map { param => param.name.asInstanceOf[c.TermName] }
+
+    val cuber = q"""
+    new _root_.com.twitter.algebird.macros.Cuber[${T}] {
+      type K = $tupleName[..$types]
+      def apply(in: ${T}): _root_.scala.Seq[K] = for (..$fors) yield new K(..$names)
```
ianoc: can we just statically unroll this in the macro ?

ianoc: we can then cache our Some allocations, right now we would re-allocate a lot

algebird-core/src/main/scala/com/twitter/algebird/macros/Cuber.scala
```diff
@@ -0,0 +1,139 @@
...
+ *
+ * Then we could do
+ * > import com.twitter.algebird.macros.Roller.roller
+ * > val data: List[Data]
+ * > val averageTemps: Map[(Option[String], Option[String], Option[String]), Double] =
+ * > data.flatMap { d => roller((d.continent, d.country, d.city)).map((_, d)) }
+ * >   .groupBy(_._1)
+ * >   .mapValues { xs => val temps = xs.map(_.temperature); temps.sum / temps.length }
+ */
+trait Roller[I] {
```
johnynek: how is this different from Cuber?

algebird-core/src/main/scala/com/twitter/algebird/macros/Cuber.scala
```diff
@@ -0,0 +1,139 @@
...
+
+    ensureCaseClass(c)
+
+    val params = getParams(c)
+    if (params.length > 22)
+      c.abort(c.enclosingPosition, s"Cannot create Cuber for $T because it has more than 22 parameters.")
+    if (params.length == 0)
+      c.abort(c.enclosingPosition, s"Cannot create Cuber for $T because it has no parameters.")
+
+    val tupleName = newTypeName(s"Tuple${params.length}")
```
johnynek: don't we need `_root_.scala.` here?

algebird-core/src/main/scala/com/twitter/algebird/macros/Cuber.scala
johnynek: +1 to unrolling in the macro. The nested flatMaps really deep will probably optimize poorly.

algebird-test/src/test/scala/com/twitter/algebird/macros/CuberMacroTest.scala
```diff
@@ -0,0 +1,67 @@
...
+  case class Bar(a: Boolean, foo: Foo)
+  case class Baz(a: Int)
+
+  property("Cuber works for Foo") = forAll { f: Foo =>
+    Cuber.cuber(f).toSet == Set(
```
johnynek: can we also verify the length is correct? toSet could be removing duplicates.

algebird-core/src/main/scala/com/twitter/algebird/macros/Cuber.scala
johnynek: okay, I get it. This is doing a prefix only, not the full data cube. Given that the type is isomorphic, it is easy to get confused. Can we make a shorter explanation at the top of the comment? Something like:
```scala
/**
 * For a tuple N produces a result with (N + 1) elements each of arity N such that there is a suffix of k Nones, for all k
 * from 0 to N.
```

algebird-core/src/main/scala/com/twitter/algebird/macros/Cuber.scala
johnynek: can we add a method like:

```scala
  def cube[K, V](t: TraversableOnce[(K, V)])(implicit c: Cuber[T], sg: Semigroup[V]): Map[c.K, V] =
    MapAlgebra.sumByKey(t.toIterator.flatMap { case (k, v) => c(k).map((_, v)) })
```
and similarly for rollup?

(and tests? kthxbai)

algebird-test/src/test/scala/com/twitter/algebird/macros/CuberMacroTest.scala
sid-kap: I thought @johnynek preferred scalacheck over scalatest?

algebird-test/src/test/scala/com/twitter/algebird/macros/CuberMacroTest.scala
sid-kap: Bar is a nested case class -- it contains a Foo

algebird-test/src/test/scala/com/twitter/algebird/macros/CuberMacroTest.scala
ianoc: He's not a huge fan of scalatest its true, but most of our tests are already scalatest, and it provides metrics on how many run. A big pain point is when tests don't get run, so it would provide us that data so we can graph/track in future.

algebird-test/src/test/scala/com/twitter/algebird/macros/CuberMacroTest.scala
sid-kap: Fixed

algebird-core/src/main/scala/com/twitter/algebird/macros/Cuber.scala
sid-kap: Fixed

algebird-core/src/main/scala/com/twitter/algebird/macros/Cuber.scala
```diff
@@ -0,0 +1,159 @@
...
+
+    val somes = params.zip(Stream.from(1)).map {
+      case (param, index) =>
+        val name = newTermName(s"some$index")
+        q"val $name = _root_.scala.Some(in.$param)"
+    }
+
+    val options = (1 to arity).map { index =>
+      val some = newTermName(s"some$index")
+      q"if (((1 << ${index - 1}) & i) == 0) _root_.scala.None else $some"
```
johnynek: you can evaluate `((1 << ${index - 1}) & i) == 0` at compile time. What not:

```scala
 (1 to arity).map { index =>
  val some = newTermName(s"some$index")
  if (((1 << ${index - 1}) & i) == 0) q"_root_.scala.None" else q"$some"
}
```

johnynek: ahh, I see. It has `i` in it, which is not known at this time.

johnynek:
Looks good to me (I'll defer to Ian if he really wants scalatest, I get the desire for a count of tests, but we've had several issues with property checks not running with scalatest, which worries me. It's harder to have that happen using scalacheck's approach, which is type checked, as opposed to scalatest using functions to unit which silently ignore functions that have no asserts).

As to the API, I'd still like to see:
```scala
object MapAlgebra {
  def cube/rollup[K, V](it: TraversableOnce[(K, V)])(implicit c: Cuber[K], sg: Semigroup[V]): Map[c.K, V]
  // maybe we could also do:
  def cubeBy[T, K,U,V](it: TraversableOnce[T], agg: Aggregator[T,U,V])(fn: T => K)(implicit c: Cuber[K]): Map[c.K, V]
  // using scala's groupBy usually performs very poorly. We'd probably need to use MapAlgebra.sumByKey which internally uses a mutable map.
}
```

algebird-core/src/main/scala/com/twitter/algebird/macros/Cuber.scala
sid-kap: The cube function in Pig and SQL doesn't seem to have aggregation built in -- it only generates the tuples, and then you manually do a groupby/aggregation/etc afterwards.

Maybe we should reduce the number of things that the `cube` function does to follow that style -- either 
1) only produce the tuples (don't group or aggregate) i.e. return a `TraversableOnce[c.K, TraversableOnce[V]]`
2) produce the tuples and group them (don't aggregate), i.e. return a `Map[c.K, TraversableOnce[V]]`

Then the user can decide what they want to do with the resulting sequence or map.

(This also seems more in line with how we were planning to implement this in Scalding -- TypedPipe#cube would return a TypedPipe[(c.K, V)].)

The method that you mentioned above is definitely useful -- we could call it `cubeAggregate` or something like that to indicate that it's doing more than just cubing the records.

johnynek: the difference is that in scala, there is no optimizer that can avoid the allocation if you generate `Map[c.K, List[V]]` If you generate a TraversableOnce or Iterator, it is not so safe (you can't reuse it, and the type system does not know about ownership, consuming either does not have quite the same risk).

Also, scalding is different because we don't materialize the whole `TypedPipe/Grouped` into memory at once, so we can probably just have a `cube` method.

That said, are there good examples of things you want to do that are not semigroups in these cases (even commutative semigroups)? I'm not a big fan of giving people more general APIs than are commonly needed, when we have more constrained ones that have nice laws.

sid-kap:
(I posted a comment in the diff above but it's messy to have this discussion in two places so might as well just move it here.)

The cube function in Pig and SQL doesn't seem to have aggregation built in -- it only generates the tuples, and then you manually do a groupby/aggregation/etc afterwards. Maybe we should emulate that here, by keeping the aggregation separate from the `cube` method.

This would be more in line with how we were planning to implement this in Scalding -- if we want `TypedPipe[(K,V)]#cube` to return a `TypedPipe[(c.K, V)]`, then it would make sense for `MapAlgebra.cube(t: TraversableOnce[(K,V)])` to return either a `TraversableOnce[(c.K, V)]` or a `Map[c.K, V]`.

I don't have a huge problem with semigroup aggregation here, but I think it would be nice to have consistent names and types between the List/Map version of cube and the TypedPipe cube.

(Maybe we could find a different name for the cube function you mentioned above? Maybe `cubeAggregate` or something?)

johnynek:
How about `MapAlgebra.cubeSum` and `MapAlgebra.cubeAggregate` that use
semigroup and Aggregator respectively to return a `Map`.

On Mon, Aug 17, 2015 at 2:16 PM, Sidharth Kapur <notifications@github.com>
wrote:

> (I posted a comment in the diff above but it's messy to have this
> discussion in two places so might as well just move it here.)
>
> The cube function in Pig and SQL doesn't seem to have aggregation built in
> -- it only generates the tuples, and then you manually do a
> groupby/aggregation/etc afterwards. Maybe we should emulate that here, by
> keeping the aggregation separate from the cube method.
>
> This would be more in line with how we were planning to implement this in
> Scalding -- if we want TypedPipe[(K,V)]#cube to return a TypedPipe[(c.K,
> V)], then it would make sense for MapAlgebra.cube(t:
> TraversableOnce[(K,V)]) to return either a TraversableOnce[(c.K, V)] or a Map[c.K,
> V].
>
> I don't have a huge problem with semigroup aggregation here, but I think
> it would be nice to have consistent names and types between the List/Map
> version of cube and the TypedPipe cube.
>
> (Maybe we could find a different name for the cube function you mentioned
> above? Maybe cubeAggregate or something?)
>
> —
> <https://github.com/twitter/algebird/pull/483#issuecomment-132002212>.
>



-- 
Oscar Boykin :: @posco :: http://twitter.com/posco

sid-kap:
Sure, that sounds good.

algebird-core/src/main/scala/com/twitter/algebird/MapAlgebra.scala
```diff
@@ -222,4 +224,31 @@ object MapAlgebra {
 
   def dot[K, V](left: Map[K, V], right: Map[K, V])(implicit mring: Ring[Map[K, V]], mon: Monoid[V]): V =
     Monoid.sum(mring.times(left, right).values)
+
+  def cube[K, V](it: TraversableOnce[(K, V)])(implicit c: Cuber[K]): Map[c.K, TraversableOnce[V]] = {
+    val sg = new Semigroup[Iterable[V]] {
+      def plus(x: Iterable[V], y: Iterable[V]) = x ++ y
+    }
+    cubeSum(it.map { case (k, v) => (k, Iterable(v)) })(c, sg)
```
sid-kap: The trade-off here is to allocate lots more objects (the Semigroup, lots of Iterable singletons and Maps) in order to take advantage of sumByKey (rather than using Scala's groupBy). Is this a good idea?

algebird-test/src/test/scala/com/twitter/algebird/macros/CuberMacroTest.scala
```diff
@@ -0,0 +1,81 @@
...
+        (Some(b.a), Some(b.foo))))
+    }
+  }
+
+  property("Roller works for Baz") {
+    forAll { b: Baz =>
+      assert(Roller.roller(b) == List(
+        Tuple1(None),
+        Tuple1(Some(b.a))))
+    }
```
johnynek: can we test cubeSum and cubeAggregate at least?

algebird-core/src/main/scala/com/twitter/algebird/MapAlgebra.scala
johnynek: I don't think this approach is going to work well (the default Iterable is List, I think, so you're going to get repeated ++ on List which gives an O(N^2) algorithm).

You're better off doing something like:
```scala
val map: collection.mutable.Map[c.K, List[V]] = collection.mutable.Map[c.K, List[V]]()
it.toIterator.foreach { case (k, v) =>
  c(k).foreach { ik =>
    map.get(ik) match {
     case Some(vs) => map += ik -> (v :: vs)
     case None => map += ik -> List(v)
    }
  }
  // now reverse the lists and return an immutable map.
  // you might just wrap this one with the immutable map that wraps a mutable one (so we
  // don't pay the immutable construction cost unless we need to, i.e. the user calls + or .toMap).
}
```

algebird-core/src/main/scala/com/twitter/algebird/MapAlgebra.scala
```diff
@@ -222,4 +224,31 @@ object MapAlgebra {
...
+      def plus(x: Iterable[V], y: Iterable[V]) = x ++ y
+    }
+    rollupSum(it.map { case (k, v) => (k, Iterable(v)) })(r, sg)
+  }
+
+  def rollupSum[K, V](it: TraversableOnce[(K, V)])(implicit r: Roller[K], sg: Semigroup[V]): Map[r.K, V] =
+    sumByKey(it.toIterator.flatMap { case (k, v) => r(k).map((_, v)) })
+
+  def rollupAggregate[T, K, U, V](it: TraversableOnce[T], agg: Aggregator[T, U, V])(fn: T => K)(implicit r: Roller[K]): Map[r.K, V] =
+    sumByKey(it.toIterator.flatMap { t => r(fn(t)).map((_, agg.prepare(t))) })(agg.semigroup).mapValues(agg.present)
```
johnynek: mapValues is lazy in scala, and we tend not to use it (because it keeps things in scope we don't expect, and we recompute the values over and over).

johnynek: so, prefer `.map { case (k, v) => (k, agg.present(v)) }` here.

algebird-core/src/main/scala/com/twitter/algebird/MapAlgebra.scala
sid-kap: I can't seem to find an immutable wrapper to a mutable map in the standard library. So our options here are 

* write a wrapper class
* return an immutable map by calling `.toMap` on the map
* return the mutable map, but specify the return type as `scala.collection.Map` instead of `scala.collection.mutable.Map`. In this case, the user doesn't know that it's mutable, so it's effectively immutable. (They can only access the mutator methods by casting.) If they really want a `immutable.Map`, they can call `.toMap` on it.

ianoc: We have one around in algebird I believe.

BTW option 3 is bad because the contract there lets the underlying map move about on you, you can't presume its stable/immutable. Just that the handle you have is. Subtle but important difference

sid-kap: Ah, found the mutable backed map.

algebird-test/src/test/scala/com/twitter/algebird/macros/CuberMacroTest.scala
sid-kap: Added tests

algebird-core/src/main/scala/com/twitter/algebird/MapAlgebra.scala
sid-kap: Fixed

algebird-core/src/main/scala/com/twitter/algebird/MapAlgebra.scala
```diff
@@ -222,4 +224,49 @@ object MapAlgebra {
 
   def dot[K, V](left: Map[K, V], right: Map[K, V])(implicit mring: Ring[Map[K, V]], mon: Monoid[V]): V =
     Monoid.sum(mring.times(left, right).values)
+
+  def cube[K, V](it: TraversableOnce[(K, V)])(implicit c: Cuber[K]): Map[c.K, TraversableOnce[V]] = {
```
johnynek: let's make the value return type `List[V]` or `Iterable[V]`.

algebird-core/src/main/scala/com/twitter/algebird/MapAlgebra.scala
```diff
@@ -222,4 +224,49 @@ object MapAlgebra {
...
+  }
+
+  def cubeSum[K, V](it: TraversableOnce[(K, V)])(implicit c: Cuber[K], sg: Semigroup[V]): Map[c.K, V] =
+    sumByKey(it.toIterator.flatMap { case (k, v) => c(k).map((_, v)) })
+
+  def cubeAggregate[T, K, U, V](it: TraversableOnce[T], agg: Aggregator[T, U, V])(fn: T => K)(implicit c: Cuber[K]): Map[c.K, V] =
+    sumByKey(it.toIterator.flatMap { t => c(fn(t)).map((_, agg.prepare(t))) })(agg.semigroup)
+      .map { case (k, v) => (k, agg.present(v)) }
+
+  def rollup[K, V](it: TraversableOnce[(K, V)])(implicit r: Roller[K]): Map[r.K, TraversableOnce[V]] = {
```
johnynek: same make the value `List[V]` or `Iterable[V]`.

algebird-test/src/test/scala/com/twitter/algebird/macros/CuberMacroTest.scala
```diff
@@ -0,0 +1,282 @@
...
+import org.scalatest.{ PropSpec, Matchers }
+import org.scalatest.prop.PropertyChecks
+
+class CuberRollerProperties extends PropSpec with PropertyChecks with Matchers {
+  import BaseProperties._
+
+  implicit val arbitraryFoo: Arbitrary[Foo] = arbitrary[Foo]
+  implicit val arbitraryBar: Arbitrary[Bar] = arbitrary[Bar]
+  implicit val arbitraryBaz: Arbitrary[Baz] = arbitrary[Baz]
+  implicit def arbitraryCaseClassWithTypeParams[T: Arbitrary]: Arbitrary[CaseClassWithTypeParams[T]] =
```
johnynek: nice improvement.

johnynek:
:+1: @ianoc any concerns?

ianoc:
Nope, this looks great to me

algebird-core/src/main/scala/com/twitter/algebird/MapAlgebra.scala
sid-kap: fixed

algebird-core/src/main/scala/com/twitter/algebird/MapAlgebra.scala
sid-kap: fixed

PR #484, opened 2015-08-18
add immutable version of sorted(Reverse)Take

nevillelyh:


ianoc:
Test(s)?

algebird-core/src/main/scala/com/twitter/algebird/TopKMonoid.scala
```diff
@@ -87,3 +87,9 @@ class TopKMonoid[T](k: Int)(implicit ord: Ordering[T]) extends Monoid[TopK[T]] {
     TopK(size, reversed, max)
   }
 }
+
+class TopKToListAggregator[A](max: Int)(implicit ord: Ordering[A]) extends MonoidAggregator[A, TopK[A], List[A]] {
+  override def monoid: Monoid[TopK[A]] = new TopKMonoid[A](max)(ord)
```
johnynek: can we make this a `val`? the def will allocate on each access.

algebird-core/src/main/scala/com/twitter/algebird/Aggregator.scala
```diff
@@ -103,6 +103,16 @@ object Aggregator extends java.io.Serializable {
   def sortedReverseTake[T: Ordering](count: Int): MonoidAggregator[T, PriorityQueue[T], Seq[T]] =
     new mutable.PriorityQueueToListAggregator[T](count)(implicitly[Ordering[T]].reverse)
   /**
+   * Immutable version of sortedTake
```
johnynek: can we add a comment as to when you might want to use this?

A few years ago I benchmarked the PriorityQueue version to be much faster (but that was before we had the benchmark subproject sadly). Given that users generally don't care what the middle type is, when would I want this?  I assume when I am disassembling the Aggregator and running each part myself (as happens in scalding or spark), and for some reason I don't want mutable intermediate objects (I can't think of a reason we care about that if it is a lot faster though).

algebird-core/src/main/scala/com/twitter/algebird/TopKMonoid.scala
nevillelyh: done

algebird-core/src/main/scala/com/twitter/algebird/Aggregator.scala
nevillelyh: I've seen frameworks that check immutability of reduce functions, for example [https://github.com/GoogleCloudPlatform/DataflowJavaSDK/commit/ca2d22e198212c38e76b8b91ae8698dc2d877666](Dataflow). This would work around that. Also for the sake of completeness?
It's not very intuitive to write something like:

```scala
Aggregator
    .prepareMonoid[T, TopK[T]](i => new TopK(1, List(i), Some(i)))(new TopKMonoid[T](max)(ord))
    .andThenPresent(_.items.toSeq)
```

nevillelyh:
@ianoc I see that both `TopKMonoid` and `PriorityQueueMonoid` are covered but there's nothing for `Aggregator._`. Is this necessary?

ianoc:
If none of the other aggregators are covered thats fine. I was just hoping it would be a 1 - 2 line addition. Oh well

ianoc:
LGTM, thanks!

PR #485, opened 2015-08-20
Fix tests that were not actually running

sid-kap:
These tests were not running due to improper mixing of Scalacheck and Scalatest props.


johnynek:
This shit constantly reminds me how easy it is to misuse scalatest and how dangerous test migrations are.

PR #486, opened 2015-08-28
Use .flatten instead of filter and then map

ntnj:


benpence:
+1

ianoc:
Have you checked there is no performance loss here vs the two? Flatten possibly allocates more than doing a .filter.map

mosesn:
Might also be interesting to compare against .flatMap(identity)

ntnj:
I didn't check for performance difference between the two.
So I decided to dive into scala library's collection code(!) to get some ideas. For a common case of `items: Seq[Option[Int]]`,
`flatten` is defined in https://github.com/scala/scala/blob/2.11.x/src/library/scala/collection/generic/GenericTraversableTemplate.scala#L169.
`filter` is defined in https://github.com/scala/scala/blob/2.11.x/src/library/scala/collection/TraversableLike.scala#L256
`map` is defined in https://github.com/scala/scala/blob/2.11.x/src/library/scala/collection/TraversableLike.scala#L238. `flatMap` just below it. (It's same as `flatten` if the function is `identity`).

`flatten` calls `.seq` on each item, which for `Option` calls `.toList` (via an implicit). Then there is a call to `++=` on the builder, which calls `.foreach` on the result `List`, which then calls `.iterator` on the list. So for each `Option`, `.toList.iterator.foreach` is called. (which could have been just `.foreach`, if `Option` directly extended `TraversableOnce`, instead of an implicit conversion to it)
So, there is an allocation of `List` (`List()` or `::`) and `Iterator` for each `Option` in `flatten`.
For the `filter` and then `map` case, there is an extra `Builder` allocation, (which is `ListBuffer` for a `Seq`), and it iterates twice.

So, no definite conclusion, but I guess that if most of the Option's are nonEmpty, `flatten` might be faster due to avoiding iterating over the list twice. If most are empty, `flatten` will create an iterator for each empty `List`. (There may be a short circuit somewhere to just return constant `Iterator.empty` for `List()` but I couldn't find it)

(The methods and call chains mentioned here were found using IntelliJ's go to definition for each type. If I missed something, whole thing might be invalid. Scala's collection code is quite complex to go through.)

ianoc:
.filter.map doesn't iterate twice on a traversable once. Map/flat
map/filter/etc are all lazy on traversable once instances. Indeed they must
be as you can only traverse it once and there is no contract that it could
be buffered into memory.

mosesn:
@ianoc I don't think that's right.  NB, Seq(3) is-a TraversableOnce[Int].

algebird-core/src/main/scala/com/twitter/algebird/Monoid.scala
```diff
@@ -71,7 +71,7 @@ class OptionMonoid[T](implicit semi: Semigroup[T]) extends Monoid[Option[T]] {
   }
   override def sumOption(items: TraversableOnce[Option[T]]): Option[Option[T]] =
     if (items.isEmpty) None
-    else Some(semi.sumOption(items.filter(_.isDefined).map { _.get }))
+    else Some(semi.sumOption(items.flatten))
```
johnynek: there are several choices here. For algebird internals, we really try to have benchmarks (see the benchmark subproject) to make the choice. Some alternatives:
```scala
// we can try with or without .toIterator, which may force some laziness we might want
items.toIterator.flatMap(identity)
items.toIterator.collect { case Some(t) => t }
```

ianoc:
Its lazy alright, example:


scala> val l = List(1,2,3)

l: List[Int] = List(1, 2, 3)

scala> val x: TraversableOnce[Int] = l

x: TraversableOnce[Int] = List(1, 2, 3)

scala> x.map{ x => sys.error(x.toString); x}

res0: scala.collection.TraversableOnce[Int] = non-empty iterator

scala> res0.foreach(println)

java.lang.RuntimeException: 1

at scala.sys.package$.error(package.scala:27)

at $anonfun$1.apply$mcII$sp(<console>:10)

mosesn:
wat?  how does that even work?  is that because of the TraversableOnce cbf?

ianoc:
@mosesn I believe its because flatMap and map aren't defined on TraversableOnce itself, so they can't be overridden. They are supplied by an implicit class of MonadOps, which does the math on iterators.

sritchie:
We should add a benchmark here and see which one helps, then merge or close.

PR #493, opened 2015-10-04
Add Accurate Count-Distinct aggregator (RoaringBitmap)

vidma:
Internally, this uses `RoaringBitmap`, a compressed alternative to `BitSet` (it's rather fast, and used commonly in projects as `spark`, `druid` etc).

This is very simple, but IMHO algebird is still missing this :)

I'll add tests if you'll be willing to merge it (our tests are currently Spark/DataFrame dependent)

@johnynek

### DIFF
--- a/algebird-core/src/main/scala/com/twitter/algebird/RoaringBitmapAggregator.scala
+++ b/algebird-core/src/main/scala/com/twitter/algebird/RoaringBitmapAggregator.scala
@@ -1,0 +1,41 @@
+/*
+Copyright 2015 Vinted.com
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package com.twitter.algebird
+
+import org.roaringbitmap._
+
+object ExactCountDistinct {
+  implicit val monoid = new RoaringBitampMonoid()
+
+  /**
+   * Accurate Count-Distinct for integer values (<= 2,147,483,646)
+   * Internally, this uses `RoaringBitmap`, a compressed alternative to `BitSet`.
+   */
+  def accurateDistinctAggregator: MonoidAggregator[Int, RoaringBitmap, Int] = {
+    Aggregator.prepareMonoid { value: Int =>
+      RoaringBitmap.bitmapOf(value)
+    }.andThenPresent(_.getCardinality)
+  }
+}
+
+class RoaringBitampSemigroup extends Semigroup[RoaringBitmap] {
+  def plus(l: RoaringBitmap, r: RoaringBitmap): RoaringBitmap = RoaringBitmap.or(l, r)
+}
+
+class RoaringBitampMonoid extends RoaringBitampSemigroup with Monoid[RoaringBitmap] {
+  def zero = new RoaringBitmap()
+}
--- a/algebird-core/src/main/scala/com/twitter/algebird/RoaringBitmapAggregator.scala
+++ b/algebird-core/src/main/scala/com/twitter/algebird/RoaringBitmapAggregator.scala
@@ -17,9 +17,12 @@ limitations under the License.
 package com.twitter.algebird
 
 import org.roaringbitmap._
+import org.roaringbitmap.FastAggregation
+
+import scala.collection.JavaConversions.asJavaIterator
 
 object ExactCountDistinct {
-  implicit val monoid = new RoaringBitampMonoid()
+  implicit val monoid = new RoaringBitMapMonoid()
 
   /**
    * Accurate Count-Distinct for integer values (<= 2,147,483,646)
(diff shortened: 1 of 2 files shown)


ianoc:
I think our main concern with this will be the addition of the dependency.  Looks like its pretty pure of a dep, with all the packages it uses seem to be only used in tests. Small enough to be in core you think @johnynek or we should have it in an algebird-X ?

johnynek:
Yes the dependency is the trick here. This would add a new dependency for everyone, even those that don't use this feature.

A second concern is that this would be the second bitset dependency (we took another one, which may have been a mistake):
https://github.com/twitter/algebird/blob/develop/algebird-core/src/main/scala/com/twitter/algebird/BloomFilter.scala#L22

we definitely should not have two compressed bitset dependencies. They claim this performs better than Ewah. If that's true, we should consider just moving to this.

In the mean time, it might be better to make the aggregator without adding an additional dependency just by using the current dependency, and then we can evaluate if we should change to only use the RoaringBitmap implementation. That would be my opinon.

Also, note, there is a faster aggregation for RoaringBitmap according to the docs:
https://github.com/lemire/RoaringBitmap/blob/master/src/main/java/org/roaringbitmap/FastAggregation.java

which could be used as the implementation for sumOption on the semigroup, which should speed things up when this is used with spark or scalding.

One more comment: if we did merge this, the monoid should probably be in the mutable namespace since this datastructure is mutable (even though we are not mutating it here).

algebird-core/src/main/scala/com/twitter/algebird/RoaringBitmapAggregator.scala
```diff
@@ -0,0 +1,41 @@
...
+   * Internally, this uses `RoaringBitmap`, a compressed alternative to `BitSet`.
+   */
+  def accurateDistinctAggregator: MonoidAggregator[Int, RoaringBitmap, Int] = {
+    Aggregator.prepareMonoid { value: Int =>
+      RoaringBitmap.bitmapOf(value)
+    }.andThenPresent(_.getCardinality)
+  }
+}
+
+class RoaringBitampSemigroup extends Semigroup[RoaringBitmap] {
```
avibryant: `Bitamp` should be `Bitmap` (also appears elsewhere)

avibryant:
Agreed that we should pick a compressed bitset to use. It seems worth noting that the author of the EWAH implementation we use (Lemire) is also the author of the RoaringBitmap paper and implementation, which gives his claims about which one is faster a lot of weight.

vidma:
good point on `sumOption`, finally Algebird's spark support might do some good  for us (I hope).

> Agreed that we should pick a compressed bitset to use. 

Sounds good. then, I'll look into `sumOption` and add some tests.

vidma:
btw, @johnynek ,  regarding `sumOption`  ...

looking again at https://github.com/twitter/algebird/pull/397/files , I see your comment

> I don't see a way to use sumOption in sumByKey or aggregateByKey with reimplementing or skipping map-side combining. I wish spark had something like scalding's sumByLocalKeys.

I wish too! any new ideas how to add this `sumByLocalKeys` to Spark?

johnynek:
Guess I forgot about that. No new ideas on how to solve it.

On Monday, October 5, 2015, vidma <notifications@github.com> wrote:

> btw, @johnynek <https://github.com/johnynek> , regarding sumOption ...
>
> looking again at https://github.com/twitter/algebird/pull/397/files , I
> see your comment
>
> I don't see a way to use sumOption in sumByKey or aggregateByKey with
> reimplementing or skipping map-side combining. I wish spark had something
> like scalding's sumByLocalKeys.
>
> I wish too! any new ideas how to add this sumByLocalKeys to Spark?
>
> —
> <https://github.com/twitter/algebird/pull/493#issuecomment-145667014>.
>


-- 
Oscar Boykin :: @posco :: http://twitter.com/posco

vidma:
continuing  `sumByLocalKeys`  offtopic, I'd say the only useful place is a combiner after a Shuffle, if using a sort-based shuffle (locality), but not so simple i guess...

P.S. as I understand even Scalding calls `sumOption` only after shuffle [SummingCache uses `.plus`].

johnynek:
yes, for now, we only use sumOption after shuffle. We could do this on the
map-side as well, but the problem there is always about how many items to
cache in memory to possibly improve summing. A jit-like approach would be
great (do some measurements on a few items, then use those to do the rest.
That or store history and check a history service to see how many items
should be kept for each key before calling sumOption).

On Mon, Oct 5, 2015 at 12:57 PM, vidma <notifications@github.com> wrote:

> continuing sumByLocalKeys offtopic, I'd say the only useful place is a
> combiner after a Shuffle, if using a sort-based shuffle (locality), but not
> so simple i guess...
>
> P.S. as I understand even Scalding calls sumOption only after shuffle
> [SummingCache uses .plus].
>
> —
> <https://github.com/twitter/algebird/pull/493#issuecomment-145692343>.
>



-- 
Oscar Boykin :: @posco :: http://twitter.com/posco

ianoc:
Scalding only uses it reduce side natively right now, though Summingbird actually will use sumOption mapside optionally. 

I've used it map side with two general strategies before:
1) If we can quickly compute the topN (some code used in SB is in algebird-util here) then we can do something with those
2) If we know the overall space cardinality is small, or are happy to rely on lots of LRU evictions then putting lists in for the values works pretty well.

In scalding using sumByLocalKeys it looks something like:    .map(specialType).sumByLocalKeys.map(_.present).group.forceToReducers.sum

PR #494, opened 2015-10-08
Add ScalableBloomFilter implementation

ghost:
Hi! 

I wrote this implementation of a _scalable_ bloom filter a while ago, based on this [paper](http://gsd.di.uminho.pt/members/cbm/ps/dbloom.pdf), that uses the BloomFilterMonoid as its underlying implementation.  We've been using this SBF at Spotify for about two years in our music discovery pipelines.  I would like to contribute it upstream as I think it's useful for many situations.  

One of the major issues with typical bloom filter implementations is that you need to estimate ahead of time how many items will be added to the filter and can't adjust this estimate when it becomes necessary.  An example is using a bloom filter for tracking user listening histories, which both grow over time and vary greatly in size between users.

This implementation can save space by starting with the same initial filters (with the same false probability rates) and then grow only the ones necessary.  The way they "grow" is by adding new filters to the underlying collection of filters, each one sized by a _growth rate_ larger than the previous.  There's also a _tightening ratio_ that can be used to keep the overall false probability rate near the one initially specified.

I'm open to any suggestions and hope to be able to contribute this!

### DIFF
--- a/algebird-core/src/main/scala/com/twitter/algebird/ScalableBloomFilter.scala
+++ b/algebird-core/src/main/scala/com/twitter/algebird/ScalableBloomFilter.scala
@@ -1,0 +1,94 @@
+package com.twitter.algebird
+
+import com.googlecode.javaewah.{ EWAHCompressedBitmap => CBitSet }
+import scala.collection.mutable.{ BitSet => MBitSet }
+
+object ScalableBloomFilter {
+
+  /**
+   * The default parameter values for this implementation are based on the findings in "Scalable Bloom Filters",
+   * Almeida, Baquero, et al.: http://gsd.di.uminho.pt/members/cbm/ps/dbloom.pdf
+   */
+  def apply(fpProb: Double,
+    initialCapacity: Int,
+    growthRate: Int = 2,
+    tighteningRatio: Double = 0.9) = {
+    new ScalableBloomFilter(fpProb, initialCapacity, growthRate, tighteningRatio,
+      List(BloomFilter(initialCapacity, fpProb).create()))
+  }
+}
+
+/**
+ * Scalable Bloom Filter - a scalable collection of [[BloomFilterMonoid]] that approximately
+ * maintains its false probability rate as it grows.
+ *
+ * @param fpProb - desired false probability rate
+ * @param headCapacity - expected number of insertions in the initial filter after which a new filter is added
+ * @param growthRate - factor by which each additional bloom filter is larger than the previous
+ * @param tighteningRatio - how much "tighter" the fpProb must be for each successive filter added
+ * @param filters - the underlying [[BF]] collection
+ */
+case class ScalableBloomFilter(fpProb: Double,
+  headCapacity: Int,
+  growthRate: Int,
+  tighteningRatio: Double,
+  filters: List[BF]) extends java.io.Serializable {
+
+  /**
+   * Add a single item to the SBF.  If this increases the estimated number of items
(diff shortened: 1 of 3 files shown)


jnievelt:
I'll try to look this over when I have time.  But I'm not 100% confident algebird is the right place for this.  Perhaps other contributors can comment, but I'm wondering if this structure is a monoid?  If so, it would be great to include one with the PR.

Otherwise, I think you're referencing `BloomFilterMonoid` here when it should be `BF` (perhaps unfortunate that it's not called BloomFilter).

edwardnewett:
Hi @jnievelt, thanks I'd appreciate your feedback when you have the time to look it over.  

Also my mistake- the code uses BF not BloomFilterMonoid.  The SBF does not have a Monoid implementation currently.  Making it a Monoid would be tricky since each SBF is composed of a collection of BF of different sizes with different hashes & widths.

jnievelt:
Actually it looks not to be a monoid.  Even if two SBF had BF lists that matched exactly in count and dimensions, you can't generally combine them without violating the fill ratio restrictions.  At best we could concatenate the lists, but we lose out on the compactness in that case (at least).

You should be able to wrap an Aggregator around it pretty easily though.

edwardnewett:
Would it be a design problem to contribute SBF without having a monoid implementation?

algebird-core/src/main/scala/com/twitter/algebird/ScalableBloomFilter.scala
```diff
@@ -0,0 +1,94 @@
...
+  /**
+   * The default parameter values for this implementation are based on the findings in "Scalable Bloom Filters",
+   * Almeida, Baquero, et al.: http://gsd.di.uminho.pt/members/cbm/ps/dbloom.pdf
+   */
+  def apply(fpProb: Double,
+    initialCapacity: Int,
+    growthRate: Int = 2,
+    tighteningRatio: Double = 0.9) = {
+    new ScalableBloomFilter(fpProb, initialCapacity, growthRate, tighteningRatio,
+      List(BloomFilter(initialCapacity, fpProb).create()))
```
jnievelt: fpProb is 'P' from the paper, while we need to specify P0 here.  By my reading we should be using:

BloomFilter(initialCapacity, fpProb * (1 - tighteningRatio))

algebird-core/src/main/scala/com/twitter/algebird/ScalableBloomFilter.scala
```diff
@@ -0,0 +1,94 @@
...
+  filters: List[BF]) extends java.io.Serializable {
+
+  /**
+   * Add a single item to the SBF.  If this increases the estimated number of items
+   * beyond the current filter capacity, a new filter will be created.
+   */
+  def +(item: String) = {
+    val head = filters.head + item
+
+    if (head.size.estimate >= headCapacity) {
```
jnievelt: I'm not sure about this check.  It may be more expensive than we need and I'm not even sure it would be correct.

From the paper, it seems that we want a (perhaps protected) method on BF which counts the number of set bits, and we want something like:
```
val initialDistFromOneHalf = math.abs(0.5 - (head.setBits / totalBits))
val newDistFromOneHalf = math.abs(0.5 - (newHead.setBits / totalBits))
if (newDistFromOneHalf > initialDistFromOneHalf) {
  // create new head BF
} else {
  // incorporate into head
}
```

If we wanted to follow the paper even more closely, we'd implement a block bloom filter.  I imagine the math is pretty close either way.

PR #495, opened 2015-10-18
An implementation of t-digest numeric distribution sketching

erikerlandson:
As described in the following paper:
Computing Extremely Accurate Quantiles Using t-Digests
Ted Dunning and Otmar Ertl
https://github.com/tdunning/t-digest/blob/master/docs/t-digest-paper/histo.pdf

### DIFF
--- a/algebird-core/src/main/scala/com/twitter/algebird/TDigest.scala
+++ b/algebird-core/src/main/scala/com/twitter/algebird/TDigest.scala
@@ -56,7 +56,7 @@ case class TDigest(
     if (s.nclusters <= R) s
     else {
       // too many clusters: attempt to compress it by re-clustering
-      val ds = scala.util.Random.shuffle(s.clusters.toVector)
+      val ds = TDigest.shuffle(s.clusters.toVector)
       ds.foldLeft(TDigest.empty(delta))((d, e) => d.update(e))
     }
   }
@@ -100,7 +100,7 @@ case class TDigest(
 
       // assign new mass (wn) among the clusters
       var cmNew = Vector.empty[(Double, Double)]
-      scala.util.Random.shuffle(s).foreach { clust =>
+      TDigest.shuffle(s).foreach { clust =>
         if (wn <= 0.0) {
           // if we have already distributed all the mass, remaining clusters unchanged
           cmNew = cmNew :+ ((clust.centroid, clust.mass))
@@ -153,6 +153,9 @@ case class TDigest(
 
 /** Factory functions for TDigest */
 object TDigest {
+  import scala.collection.SeqLike
+  import scala.collection.generic.CanBuildFrom
+
   /**
    * Default value for a t-digest delta parameter.  The number of clusters varies, roughly, as
    * about (50/delta), when data are presented in random order
@@ -194,8 +197,7 @@ object TDigest {
     delta: Double = deltaDefault)(implicit num: Numeric[N]): TDigest = {
     require(delta > 0.0, s"delta was not > 0")
     val td = data.foldLeft(empty(delta))((c, e) => c + ((e, 1)))
-    val ds = scala.util.Random.shuffle(td.clusters.toVector)
-    ds.foldLeft(empty(delta))((c, e) => c + e)
+    TDigest.shuffle(td.clusters.toVector).foldLeft(empty(delta))((c, e) => c + e)
   }
 
   /**
@@ -217,6 +219,16 @@ object TDigest {
   /** Try to add a bit of efficiency to wasteful construction of t-digest objects for each input */
   private[algebird] def prepare[N](x: N, delta: Double = deltaDefault)(implicit num: Numeric[N]) =
     TDigest(delta, 1, TDigestMap.empty + ((num.toDouble(x), 1.0)))
+
+  // Shuffle a sequence in a referentially-transparent way: pseudo-randomly, but with a random
+  // seed that is a function of the sequence argument.
+  private[algebird] def shuffle[E, S[X] <: SeqLike[X, S[X]]](
+    seq: S[E])(implicit cbf: CanBuildFrom[S[E], E, S[E]]) =
+    if (seq.length <= 1) seq
+    else {
+      val seed = scala.util.hashing.MurmurHash3.productHash((seq(0), seq(1), seq.length))
+      (new scala.util.Random(seed)).shuffle(seq)
+    }
 }
 
 /** The Semigroup type class for TDigest objects */
@@ -243,8 +255,8 @@ object TDigestSemigroup {
       val d = rtd.clusters.asInstanceOf[tdmap.tree.INodeTD].data
       ltd + ((d.key, d.value))
     } else {
-      val ds = scala.util.Random.shuffle(ltd.clusters.toVector ++ rtd.clusters.toVector)
-      ds.foldLeft(TDigest.empty(delta))((d, e) => d + e)
+      TDigest.shuffle(ltd.clusters.toVector ++ rtd.clusters.toVector)
+        .foldLeft(TDigest.empty(delta))((d, e) => d + e)
     }
   }
 }
--- a/algebird-core/src/main/scala/com/twitter/algebird/TDigest.scala
+++ b/algebird-core/src/main/scala/com/twitter/algebird/TDigest.scala
@@ -1,0 +1,263 @@
+/*
+Copyright 2015 Twitter, Inc.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package com.twitter.algebird
+
+import tdmap.TDigestMap
+
+/**
+ * A t-digest sketch of sampled numeric data, as described in:
+ * Computing Extremely Accurate Quantiles Using t-Digests,
+ * Ted Dunning and Otmar Ertl,
+ * https://github.com/tdunning/t-digest/blob/master/docs/t-digest-paper/histo.pdf
+ */
+case class TDigest(
+  delta: Double,
+  nclusters: Int,
+  clusters: TDigestMap) extends Serializable {
+
+  // re-cluster when number of clusters exceeds this threshold
+  private val R = (TDigest.K / delta).toInt
+
+  private case class Cluster(centroid: Double, mass: Double, massUB: Double)
+
+  /**
+   * Returns a new t-digest with value x included in its sketch; td + x is equivalent to
+   * td + (x, 1).
+   * @param x The numeric data value to include in the sketch
+   * @return the updated sketch
+   */
+  def +[N](x: N)(implicit num: Numeric[N]): TDigest = this.+((x, 1))
+
+  /**
+   * Returns a new t-digest with new pair (x, w) included in its sketch.
+   * @param xw A pair (x, w) where x is the numeric value and w is its weight
+   * @return the updated sketch
+   * @note This implements 'algorithm 1' from:
+   * Computing Extremely Accurate Quantiles Using t-Digests,
+   * Ted Dunning and Otmar Ertl,
+   * https://github.com/tdunning/t-digest/blob/master/docs/t-digest-paper/histo.pdf
+   */
+  def +[N1, N2](xw: (N1, N2))(implicit num1: Numeric[N1], num2: Numeric[N2]): TDigest = {
+    val s = this.update(xw)
+    if (s.nclusters <= R) s
+    else {
+      // too many clusters: attempt to compress it by re-clustering
+      val ds = scala.util.Random.shuffle(s.clusters.toVector)
+      ds.foldLeft(TDigest.empty(delta))((d, e) => d.update(e))
+    }
+  }
+
+  /**
+   * Add this digest to another
+   * @param that The right-hand t-digest operand
+   * @return the sum of left and right digests, as defined by TDigestSemigroup
+   * @see TDigestSemigroup
+   */
+  def ++(that: TDigest): TDigest = TDigestSemigroup.plus(this, that, this.delta)
+
+  // This is most of 'algorithm 1', except for re-clustering which is factored out to avoid
+  // recursive calls during a reclustering phase
+  private def update[N1, N2](xw: (N1, N2))(implicit num1: Numeric[N1], num2: Numeric[N2]) = {
+    val xn = num1.toDouble(xw._1)
+    var wn = num2.toDouble(xw._2)
+    require(wn > 0.0, "data weight must be > 0")
+
+    // Get the current cluster nearest to incoming (xn)
+    // Note: 'near' will have length 0,1, or 2:
+    // length 0 => current cluster map was empty (no data yet)
+    // length 1 => exactly one cluster was closest to (xn)
+    // length 2 => (xn) was mid-point between two clusters (both are returned, in key order)
+    val near = clusters.nearest(xn)
+
+    if (near.isEmpty) {
+      // our map is empty, so insert this pair as the first cluster
+      TDigest(delta, nclusters + 1, clusters + ((xn, wn)))
+    } else {
+      // compute upper bounds for cluster masses, from their quantile estimates
+      var massPS = clusters.prefixSum(near.head._1, open = true)
+      val massTotal = clusters.sum
+      val s = near.map {
+        case (c, m) =>
+          val q = (massPS + m / 2.0) / massTotal
+          val ub = massTotal * delta * q * (1.0 - q)
+          massPS += m
+          Cluster(c, m, ub)
+      }
+
+      // assign new mass (wn) among the clusters
+      var cmNew = Vector.empty[(Double, Double)]
+      scala.util.Random.shuffle(s).foreach { clust =>
+        if (wn <= 0.0) {
+          // if we have already distributed all the mass, remaining clusters unchanged
+          cmNew = cmNew :+ ((clust.centroid, clust.mass))
+        } else if (xn == clust.centroid) {
+          // if xn lies exactly on the centroid, add all mass in regardless of bound
+          cmNew = cmNew :+ ((clust.centroid, clust.mass + wn))
+          wn = 0.0
+        } else if (clust.mass < clust.massUB) {
+          // cluster can accept more mass, respecting its upper bound
+          val dm = math.min(wn, clust.massUB - clust.mass)
+          val mass = clust.mass + dm
+          val dc = dm * (xn - clust.centroid) / mass
+          wn -= dm
+          cmNew = cmNew :+ ((clust.centroid + dc, mass))
+        } else {
+          // cluster is at its upper bound for mass, it remains unchanged
+          cmNew = cmNew :+ ((clust.centroid, clust.mass))
+        }
+      }
+
+      // any remaining mass becomes a new cluster
+      if (wn > 0.0) cmNew = cmNew :+ ((xn, wn))
+
+      // remove original clusters and replace with the new ones
+      val clustDel = near.iterator.map(_._1).foldLeft(clusters)((c, e) => c - e)
+      val clustNew = cmNew.foldLeft(clustDel)((c, p) => c.increment(p._1, p._2))
+      val nc = nclusters - s.length + cmNew.length
+
+      // return the updated t-digest
+      TDigest(delta, nc, clustNew)
+    }
+  }
+
+  /**
+   * Compute a cumulative probability (CDF) for a numeric value, from the estimated probability
+   * distribution represented by this t-digest sketch.
+   * @param x a numeric value
+   * @return the cumulative probability that a random sample from the distribution is <= x
+   */
+  def cdf[N](x: N)(implicit num: Numeric[N]): Double = clusters.cdf(x)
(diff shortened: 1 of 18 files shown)


erikerlandson:
For reference, my blog post that describes my reasoning behind binary tree algorithms backing the t-digest:
http://erikerlandson.github.io/blog/2015/09/26/a-library-of-binary-tree-algorithms-as-mixable-scala-traits/

erikerlandson:
CI is showing a failure, but the log output makes it seem like tests succeeded.  I'm not replicating any failures on my local machine.

algebird-core/src/main/scala/com/twitter/algebird/TDigest.scala
```diff
@@ -0,0 +1,257 @@
...
+  }
+
+  /**
+   * Combine this t-digest with another to form an aggregated sketch.
+   * @param that The t-digest to aggregate with
+   * @return a new digest that is formed from aggregating the arguments
+   * @note This operation, combined with the empty digest, satisfy a Monoid law, with the caveat
+   * that it involves random shuffling of clusters, and so the result is not deterministic.
+   */
+  def ++(that: TDigest) = TDigestMonoid.plus(this, that, this.delta)
```
johnynek: return type on public methods, please.

algebird-core/src/main/scala/com/twitter/algebird/TDigest.scala
```diff
@@ -0,0 +1,257 @@
...
+    }
+  }
+
+  /**
+   * Compute a cumulative probability (CDF) for a numeric value, from the estimated probability
+   * distribution represented by this t-digest sketch.
+   * @param x a numeric value
+   * @return the cumulative probability that a random sample from the distribution is <= x
+   */
+  def cdf[N](x: N)(implicit num: Numeric[N]) = clusters.cdf(x)
```
johnynek: return type on public methods please.

algebird-core/src/main/scala/com/twitter/algebird/TDigest.scala
```diff
@@ -0,0 +1,257 @@
...
+
+  /**
+   * Compute the inverse cumulative probability (inverse-CDF) for a quantile value, from the
+   * estimated probability distribution represented by this t-digest sketch.
+   * @param q a quantile value.  The value of q is expected to be on interval [0, 1]
+   * @return the value x such that cdf(x) = q
+   */
+  def cdfInverse[N](q: N)(implicit num: Numeric[N]) = clusters.cdfInverse(q)
```
johnynek: return type, please.

algebird-core/src/main/scala/com/twitter/algebird/TDigest.scala
```diff
@@ -0,0 +1,257 @@
...
+  import scala.language.reflectiveCalls
+  import com.twitter.algebird.Monoid
+
+  /**
+   * Default value for a t-digest delta parameter.  The number of clusters varies, roughly, as
+   * about (50/delta), when data are presented in random order
+   * (it may grow larger if data are not presented randomly).  The default corresponds to
+   * an expected number of clusters of about 100.
+   */
+  val deltaDefault = (50.0 / 100.0) // delta * E[clusters] ~ 50
```
johnynek: type on public val, please.

algebird-core/src/main/scala/com/twitter/algebird/TDigest.scala
```diff
@@ -0,0 +1,257 @@
...
+
+  /**
+   * The t-digest algorithm will re-cluster itself whenever its number of clusters exceeds
+   * (K/delta).  This value is set such that the threshold is about 10x the heuristically
+   * expected number of clusters for the user-specified delta value.  Generally the number of
+   * clusters will only trigger the corresponding re-clustering threshold when data are being
+   * presented in a non-random order.
+   */
+  val K = 10.0 * 50.0
```
johnynek: type on public val, please.

algebird-core/src/main/scala/com/twitter/algebird/TDigest.scala
```diff
@@ -0,0 +1,257 @@
...
+  def aggregator[N](delta: Double = deltaDefault)(implicit num: Numeric[N]) =
+    new TDigestAggregator[N](delta)
+
+  /** Try to add a bit of efficiency to wasteful construction of t-digest objects for each input */
+  private[algebird] def prepare[N](x: N, delta: Double = deltaDefault)(implicit num: Numeric[N]) =
+    TDigest(delta, 1, TDigestMap.empty + ((num.toDouble(x), 1.0)))
+}
+
+/** The Monoid type class for TDigest objects */
+class TDigestMonoid(delta: Double = TDigest.deltaDefault) extends Monoid[TDigest] {
```
johnynek: I don't like that both TDigest and the monoid know about delta. Can we make it so only one needs to?

erikerlandson: I think you're right, the monoid has no need to know.   That does imply the need for a policy on what it means when adding two t-digests with differing delta values.  My proposal would be just adopting the left-hand-side delta, which is simple and also allows a user to choose which delta to use via picking which argument appears on the left.  I don't like them as much, but other plausible options would be:
* minimum  (i.e. pick lowest sketch resolution)
* maximum (highest resolution)
* average
* throw exception (only allow adding digests with identical delta values)

erikerlandson:
Comment on t-digest monoids:   I'm pretty sure I can address the issue of non-deterministic behavior of monoid `plus`, which may be worth doing, but I think there is a deeper failure of strict associativity, since (td1 ++ td2) ++ td3  is unlikely to yield the same final set of clusters as td1 ++ (td2 ++ td3), even if the re-clustering presentation order becomes repeatable.  I'd content that this is OK (the behavior is "statistically monoidal", to whatever extent that's a thing), but worth calling out, since I'm declaring that it has a monoid type class.

algebird-core/src/main/scala/com/twitter/algebird/TDigest.scala
```diff
@@ -0,0 +1,257 @@
...
+   * estimated probability distribution represented by this t-digest sketch.
+   * @param q a quantile value.  The value of q is expected to be on interval [0, 1]
+   * @return the value x such that cdf(x) = q
+   */
+  def cdfInverse[N](q: N)(implicit num: Numeric[N]) = clusters.cdfInverse(q)
+}
+
+/** Factory functions for TDigest */
+object TDigest {
+  import scala.language.reflectiveCalls
```
johnynek: can we remove this?

algebird-core/src/main/scala/com/twitter/algebird/TDigest.scala
```diff
@@ -0,0 +1,257 @@
...
+  import com.twitter.algebird.Monoid
```
johnynek: this is not needed now, right? This is in the space right?

algebird-core/src/main/scala/com/twitter/algebird/maps/increment.scala
```diff
@@ -0,0 +1,149 @@
...
+   * val map1 = IncrementMap.key[String].value[Int]
+   * // Use a custom ordering
+   * val ord: Ordering[String] = ...
+   * val map2 = IncrementMap.key(ord).value[Int]
+   * // A custom value monoid, defines what 'increment' means
+   * val mon: Monoid[Int] = ...
+   * val map2 = IncrementMap.key[String].value(mon)
+   * }}}
+   */
+  def key[K](implicit ord: Ordering[K]) = new AnyRef {
```
johnynek: you are using several types here, why not use a named type here? `UnsetValueIncrementMap` or something.

avibryant:
@erikerlandson agreed on "statistically monoidal" - it seems fine, and is consistent with `QTree` and with the top-k stuff in `CMS`. We should come up with some formal expression of this but it's definitely something we're comfortable with.

erikerlandson:
@johnynek looks like some unrelated HLL unit test failed, but I can't re-run it

algebird-core/src/main/scala/com/twitter/algebird/TDigest.scala
```diff
@@ -0,0 +1,263 @@
...
+   * Computing Extremely Accurate Quantiles Using t-Digests,
+   * Ted Dunning and Otmar Ertl,
+   * https://github.com/tdunning/t-digest/blob/master/docs/t-digest-paper/histo.pdf
+   */
+  def +[N1, N2](xw: (N1, N2))(implicit num1: Numeric[N1], num2: Numeric[N2]): TDigest = {
+    val s = this.update(xw)
+    if (s.nclusters <= R) s
+    else {
+      // too many clusters: attempt to compress it by re-clustering
+      val ds = scala.util.Random.shuffle(s.clusters.toVector)
```
johnynek: this is non-deterministic, but I wonder if we track a seed which we update on merges if we can make it deterministic and yet keep the property that in expectation it is randomized.

algebird-core/src/main/scala/com/twitter/algebird/TDigest.scala
```diff
@@ -0,0 +1,263 @@
...
+   * @return A t-digest sketch of the input data
+   * @note Smaller values of delta yield sketches with more clusters, and higher resolution
+   * @note The expected number of clusters will vary (roughly) as (50/delta)
+   */
+  def sketch[N](
+    data: TraversableOnce[N],
+    delta: Double = deltaDefault)(implicit num: Numeric[N]): TDigest = {
+    require(delta > 0.0, s"delta was not > 0")
+    val td = data.foldLeft(empty(delta))((c, e) => c + ((e, 1)))
+    val ds = scala.util.Random.shuffle(td.clusters.toVector)
```
johnynek: here is another randomization. Is this needed? Why not have the user shuffle if they want that behavior and just do the fold here?

algebird-core/src/main/scala/com/twitter/algebird/maps/redblack.scala
```diff
@@ -0,0 +1,219 @@
...
+*/
+
+package com.twitter.algebird.maps.redblack
+
+import math.Ordering
+
+object tree {
+  /** The color (red or black) of a node in a Red/Black tree */
+  sealed trait Color extends Serializable
+  case object R extends Color
```
johnynek: can we call these Red and Black?

algebird-core/src/main/scala/com/twitter/algebird/maps/redblack.scala
```diff
@@ -0,0 +1,219 @@
...
+  case object B extends Color
+
+  /** Defines the data payload of a tree node */
+  trait Data[K] extends Serializable {
+    /** The axiomatic unit of data for R/B trees is a key */
+    val key: K
```
johnynek: `def key: K` is what we normally use here. You can override def with a val, but an abstract val has some bad properties (like opening up the val-init-order bug as a possibility).

algebird-core/src/main/scala/com/twitter/algebird/maps/redblack.scala
```diff
@@ -0,0 +1,219 @@
...
+import math.Ordering
+
+object tree {
+  /** The color (red or black) of a node in a Red/Black tree */
+  sealed trait Color extends Serializable
+  case object R extends Color
+  case object B extends Color
+
+  /** Defines the data payload of a tree node */
+  trait Data[K] extends Serializable {
```
johnynek: this can be covariant: `Data[+K]`, which may be useful.

algebird-core/src/main/scala/com/twitter/algebird/maps/redblack.scala
```diff
@@ -0,0 +1,219 @@
...
+    /** The axiomatic unit of data for R/B trees is a key */
+    val key: K
+
+    override def hashCode = key.hashCode
+    override def equals(that: Any) = that match {
+      case data: Data[K] => this.key.equals(data.key)
```
johnynek: I think this gives a warning because the `K` can't be checked. I think you need: `case data: Data[_] => key.equals(data.key)`

algebird-core/src/main/scala/com/twitter/algebird/maps/redblack.scala
```diff
@@ -0,0 +1,219 @@
...
+  }
+
+  /**
+   * Base class of a Red/Black tree node
+   * @tparam K The key type
+   */
+  trait Node[K] extends Serializable {
+
+    /** The ordering that is applied to key values */
+    val keyOrdering: Ordering[K]
```
johnynek: I'd prefer not to have abstract vals.

algebird-core/src/main/scala/com/twitter/algebird/maps/redblack.scala
```diff
@@ -0,0 +1,219 @@
...
+  trait INode[K] extends Node[K] {
+    /** The Red/Black color of this node */
+    val color: Color
+    /** Including, but not limited to, the key */
+    val data: Data[K]
+    /** The left sub-tree */
+    val lsub: Node[K]
+    /** The right sub-tree */
+    val rsub: Node[K]
+
```
johnynek: same abstract val concern.

algebird-core/src/main/scala/com/twitter/algebird/maps/redblack.scala
```diff
@@ -0,0 +1,219 @@
...
+        else if (keyOrdering.gt(d.key, data.key)) rNode(data, lsub, rsub.ins(d))
+        else rNode(d, lsub, rsub)
+      } else {
+        if (keyOrdering.lt(d.key, data.key)) balance(bNode(data, lsub.ins(d), rsub))
+        else if (keyOrdering.gt(d.key, data.key)) balance(bNode(data, lsub, rsub.ins(d)))
+        else bNode(d, lsub, rsub)
+      }
+  }
+
+  private object RNode {
```
johnynek: can we call these RedNode/BlackNode?

algebird-core/src/main/scala/com/twitter/algebird/TDigest.scala
erikerlandson: Input data is not pre-shuffled.   What you see there is a final shuffle which is for a final clustering compression.

algebird-core/src/main/scala/com/twitter/algebird/maps/redblack.scala
johnynek: Actually, why not make:
```scala
case class RedNode[K](data: Data[K], lsub: INode[K], rsub: INode[K]) extends INode[K]
case class ...
```
or even:
```scala
case class INode[+C <: Color, K](color: C, lsub: INode[Color, K], rsub: INode[Color, K])
```
This would be more idiomatic.

algebird-core/src/main/scala/com/twitter/algebird/TDigest.scala
erikerlandson: Yes, definitely.  I was thinking of just using a Random object seeded with some variation of `hashCode` on the data, something along the lines of:

```scala
def deterministicShuffle[T](data: Seq[T]) =
  if (data.isEmpty) data else (new Random(data.head.hashCode)).shuffle(data)
```

johnynek:
This is exciting, and I really appreciate the work on this. I want to get this merged, but also I hope you agree with us that careful code review is important.

One thing I'm wondering: could we break this into two PRs? One to add the Red/black trees with tests, and the second to leverage the red/black trees to implement T-digest (that builds on the first clearly)?

What I'm not seeing yet in the review is why scala.collection.immutable.SortedMap: http://www.scala-lang.org/api/current/index.html#scala.collection.immutable.SortedMap won't work here? Could you comment in the code? Sorry if I missed it.

Also, Red/black trees are O(log_2 N), but the hash-trie based standard maps are O(log_32 N). Can you comment what features we need of the red/black that the faster hash-trie can't do for us?

algebird-core/src/main/scala/com/twitter/algebird/maps/redblack.scala
erikerlandson: Because they get inherited from, and scala doesn't seem to like inheriting from case classes.

erikerlandson:
@johnynek, I agree, it's a big PR with a lot going on, and I think that reviews may as well be rigorous otherwise they aren't as useful

erikerlandson:
@johnynek regarding inheriting from `SortedMap`, I am pretty sure I can update the tree classes to satisfy the `SortedMap` interface trait, although it involves implementing several methods that I didn't immediately have great answers for, so I punted a bit.  I had figured that might be addressed in a future PR.  However, if you prefer I can try and take that bull by the horns now.

Regarding the particular use of binary trees, one nontrivial reason for that choice was that binary trees support the log-time maintenance and queries for prefix sums, and my "cover" constructs, and a couple other basic algorithms.   So there is more going on than just mapping objects.  Although that might be possible in a non-binary-tree data structure, I think the logic would be a lot harder at best, and maybe not possible at all.

Using red-black trees was just a way to guarantee balanced trees with a well-understood data structure that I had a hope of getting to work along with all the other functionality I was layering on top :)

algebird-core/src/main/scala/com/twitter/algebird/maps/redblack.scala
erikerlandson: I think updating these to be `def` instead of `val` will be relatively straightforward

erikerlandson:
Maybe an even better way to explain it is that a lot of the logic I need to support requires that clusters are maintained in numeric / location order.  So any hash-based container will not get me what I need.  A tree data structure maintains the ordering by (numeric) key that I need, along with the desirable log-time operations.

beni05:
What’s in a great support request?

Provide us with as much essential information as possible.
Let us know the name of the user or repository you’re having trouble with.
Include any screenshots or links that are related to your problem.??

erikerlandson:
@johnynek regarding separate Red/Black node classes, in an earlier iteration I was doing that, but the code wasn't as dry.  To make a long story short, it was resulting in two parallel copies of internal node logic (which is where most of the logic is).  So I felt like designing around a single internal node class, with color as one field, was the cleaner solution.

erikerlandson:
@johnynek splitting into a "tree PR" and a "t-digest PR" is OK with me, would you like me to pull the trigger on that?

algebird-core/src/main/scala/com/twitter/algebird/maps/redblack.scala
erikerlandson: Covariance involves some open questions for me.  For example, if we want to inherit from SortedMap[K, +V], then I have to support a type-widening insert:
```scala
def +[V2 >: V](kv: (K, V2)): Map[K, V2]
```
Which I think I can do, but it isn't painless either.

johnynek:
@erikerlandson Yes. Please.

erikerlandson:
The supporting tree/map library has been factored out to #496, and I will now be keeping this branch rebased off of topic branch `feature/treemaps`

beni05:
Users must have admin or owner rights within the receiving organization before they can transfer a repository that they individually own. If the user does not already have this level of access, a temporary admin team can be created with only the user. The user sending the repository is the only one who can perform the transfer.Just as when transferring a repository between two organizations,

erikerlandson:
@johnynek @ianoc @avibryant is there still interest in #495 and #496 ?

kainoa21:
Would very much like to see this PR merged, I am currently using the [reference implementation](https://github.com/tdunning/t-digest) but would benefit from an Algebird based implementation.

erikerlandson:
@kainoa21 coincidentally I started re-visiting this last weekend.  There was a request to un-factor the various tree functions to reduce the code bulk on the back end, which is still on my to-do list.

algebird-core/src/main/scala/com/twitter/algebird/TDigest.scala
```diff
@@ -0,0 +1,275 @@
...
+      }
+
+      // assign new mass (wn) among the clusters
+      var cmNew = Vector.empty[(Double, Double)]
+      TDigest.shuffle(s).foreach { clust =>
+        if (wn <= 0.0) {
+          // if we have already distributed all the mass, remaining clusters unchanged
+          cmNew = cmNew :+ ((clust.centroid, clust.mass))
+        } else if (xn == clust.centroid) {
+          // if xn lies exactly on the centroid, add all mass in regardless of bound
```
jnievelt: What's the justification for this?  It doesn't seem to come from the paper.

In fact, there is explicit acknowledgment that two clusters could have the same centroid:
> For centroids with identical means, order of creation is used as a tie-breaker to allow an unambiguous ordering

erikerlandson: Allows centroid to serve as unique key, which simplifies things.   Multiple clusters with the same centroid adds nothing useful to the model.

johnynek:
Sorry, we did not communicate clearly on this.

This is indeed an exciting algorithm, however, if we are going to become the maintainers of it, here it needs to be significantly smaller in size. An implementation that could reuse existing scala collection classes may be small enough to accept.

Honestly, the best approach is probably for @erikerlandson to provide a low dependency subproject that hosts the t-digest, and then we could have `algebird-tdigest` sub-project which adds the relevant algebird integration (monoids, aggregators mostly, I think).

johnynek:
I'd like to add, these cases are a challenge. We want to welcome contributions, but then inclusion can seem like an endorsement which has burned us a few times in the past when implementations bit-rot or are poorly performing (not saying that will happen here, I just mean that we need to really have the bandwidth to understand and maintain the code to add it).

willb:
@johnynek What about an `algebird-contrib` repository (or organization) for cases just like this, where you have community-maintained code that was designed to work with Algebird but without an Algebird imprimatur?

erikerlandson:
This is now available as a package here:
https://github.com/isarn/isarn-sketches

So I'm going to close this out

PR #498, opened 2015-11-01
Add Applicative.sequenceGen

johnynek:
This is a version of `Applicative.sequence` that works for more input and output types.


ianoc:
lgtm, merge when green

PR #504, opened 2015-12-15
Using dropRight instead of tail for shortening BytesSpec array

### LINKED ISSUE #503: `BytesSpec` fails `Ordering` property test
We caught this property-based test failure for the `BytesSpec` `Ordering` test and I wanted to record the failing test case here:

```
[algebird] [info] BytesSpec:
[algebird] [info] - requires non-null arrays
[algebird] [info] - has a sane equals function
[algebird] [info] - has a sane hashCode function
[algebird] [info] - provides an Ordering typeclass *** FAILED ***
[algebird] [info]   TestFailedException was thrown during property evaluation.
[algebird] [info]     Message: Bytes(63,63,63,63,63,63,63,63,63,63,63,63,63,63,119,63,63,63,63,63,63,63,63,63,63,63,63,63,63,63,63,63,63,63,63,63,63,63,63,63,63,63,63,63,63,63,63,63,63,63,63,63,63,63,63,63,63,63,63,63,63,63,63,63,63,63,63,63,63,63,63,63,63,63,63,63,63,63,63,63,63,63,63,63,63,63,63,63,63,63,63,63,63,63,63,63,63,63,63,63,63,63,63,63,63,63,63,63,63,63,63,63,63,63,63,63,63,63,63,63,63,63,63,63,63,63,63,63,63,63,63,63,63,63,63,63,63,63,63,63,63) was not greater than Bytes(63,63,63,63,63,63,63,63,63,63,63,63,63,119,63,63,63,63,63,63,63,63,63,63,63,63,63,63,63,63,63,63,63,63,63,63,63,63,63,63,63,63,63,63,63,63,63,63,63,63,63,63,63,63,63,63,63,63,63,63,63,63,63,63,63,63,63,63,63,63,63,63,63,63,63,63,63,63,63,63,63,63,63,63,63,63,63,63,63,63,63,63,63,63,63,63,63,63,63,63,63,63,63,63,63,63,63,63,63,63,63,63,63,63,63,63,63,63,63,63,63,63,63,63,63,63,63,63,63,63,63,63,63,63,63,63,63,63,63,63)
[algebird] [info]     Location: (BytesSpec.scala:120)
[algebird] [info]     Occurred when passed generated values (
[algebird] [info]       wordLength = 141 // 7 shrinks
[algebird] [info]     )
```

jnievelt:
A truncated array is only guaranteed to compare as less if the truncation happens at the end.  Also we should be able to compare an empty Array so I'm not excluding the case where length == 1.

I've verified this by testing with a "\0" prefix on the random string (which makes the original code fail consistently).

Fixes #503

### DIFF
--- a/algebird-test/src/test/scala/com/twitter/algebird/BytesSpec.scala
+++ b/algebird-test/src/test/scala/com/twitter/algebird/BytesSpec.scala
@@ -115,11 +115,9 @@ class BytesSpec extends WordSpec with Matchers with GeneratorDrivenPropertyCheck
       suffixed should be > bytes1
 
       // 4. a Bytes instances and a "shorter" Bytes instance.
-      if (bytes1.array.length > 1) {
-        val shortened = Bytes(bytes1.array.tail)
-        bytes1 should be > shortened
-        shortened should be < bytes1
-      }
+      val shortened = Bytes(bytes1.array.dropRight(1))
+      bytes1 should be > shortened
+      shortened should be < bytes1
     }
   }
 


Gabriel439:
Actually, the generator itself is still excluding empty strings:

```scala
forAll((Gen.choose(1, 4096), "wordLength")) { (wordLength: Int) =>
```

While you are here, could you change the generator to also test empty strings?  I expect this will also require a change to your test because `"".dropRight(1)` is not less than `""`

jnievelt:
Oops, lost this in a recreate.  I'll reissue.

PR #506, opened 2016-02-02
Update the build

johnynek:
use .sbt and add some cleanups and pjugins.


ianoc:
quick read of various sections. But think good enough. LGTM

PR #511, opened 2016-02-17
Identity class and Monad instance

NathanHowell:
Having `Identity` around is useful for those of us writing `Applicative` and `Monad` heavy code.


johnynek:
Thanks!

jnievelt:
Curious why Monad.option doesn't suffice here.  Seems like it would only handle `null` differently?

NathanHowell:
@jnievelt `Monad.some` is equivalent, `Monad.option` isn't due to null handling and returning an `Option` that needs to be pattern matched. `Identity` is the common name in Haskell (and scalaz) and perhaps more obvious in intent.

PR #514, opened 2016-02-26
Fix flaky `MinHasherSpec` test.  Fixes #500

### LINKED ISSUE #500: MinHasher32 test is flakey
```
[info] MinHasher32
[info] - should measure 0.5 similarity in 1024 bytes with < 0.1 error
[info] - should measure 0.8 similarity in 1024 bytes with < 0.05 error *** FAILED ***
[info]   0.05833333333333324 was not less than 0.05 (MinHasherTest.scala:30)
[info] - should measure 1.0 similarity in 1024 bytes with < 0.01 error
```

This is quite common. Need to weaken the bound or figure out if there is really some bug (doubt it).

Gabriel439:
One `MinHasher32` being tested was using 247 hash functions internally, which
means that the expected absolute error bound:

    1 / sqrt {# of hash functions} = 1 / sqrt 247  ~  0.064

The same test required an upper bound on the error of `0.05`, which was below
the expected error bound, meaning that this test would give flaky results due to
occasionally exceeding the required error bound.  This change increases the
upper bound on the error to 0.1 which should give enough headroom over the
expected error bound to reduce the number of random test failures.

### DIFF
--- a/algebird-test/src/test/scala/com/twitter/algebird/MinHasherTest.scala
+++ b/algebird-test/src/test/scala/com/twitter/algebird/MinHasherTest.scala
@@ -55,7 +55,7 @@ class MinHasherSpec extends WordSpec with Matchers {
     "measure 0.5 similarity in 1024 bytes with < 0.1 error" in {
       test(new MinHasher32(0.5, 1024), 0.5, 0.1)
     }
-    "measure 0.8 similarity in 1024 bytes with < 0.05 error" in {
+    "measure 0.8 similarity in 1024 bytes with < 0.1 error" in {
       test(new MinHasher32(0.8, 1024), 0.8, 0.1)
     }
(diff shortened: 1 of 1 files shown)


algebird-test/src/test/scala/com/twitter/algebird/MinHasherTest.scala
```diff
@@ -56,7 +56,7 @@ class MinHasherSpec extends WordSpec with Matchers {
       test(new MinHasher32(0.5, 1024), 0.5, 0.1)
     }
     "measure 0.8 similarity in 1024 bytes with < 0.05 error" in {
```
piyushnarang: with < 0.1 error?

Gabriel439: Whoops!  Fixed

piyushnarang:
:+1:

PR #518, opened 2016-04-09
Fix the input type of Operators.toRichTraversable

joshualande:
I could be missing something obvious, but I think Operators.toRichTraversable should take a TraversableOnce instead of a Traversable. This will allow it to be less restrictive in the input types that it allows.

Currently:
```
scala> import com.twitter.algebird.Operators.toRichTraversable

scala> val x: Traversable[Int] = Traversable(1, 2, 3)
x: Traversable[Int] = List(1, 2, 3)
scala> x.monoidSum
res11: Int = 6

scala> val y: TraversableOnce[Int] = Traversable(1, 2, 3)
y: TraversableOnce[Int] = List(1, 2, 3)
scala> y.monoidSum
<console>:17: error: value monoidSum is not a member of TraversableOnce[Int]
              y.monoidSum
                ^
```

With my fix:
```
scala> val y: TraversableOnce[Int] = Traversable(1, 2, 3)
y: TraversableOnce[Int] = List(1, 2, 3)

scala> y.monoidSum
res0: Int = 6
```

### DIFF
--- a/algebird-core/src/main/scala/com/twitter/algebird/Operators.scala
+++ b/algebird-core/src/main/scala/com/twitter/algebird/Operators.scala
@@ -20,7 +20,7 @@ object Operators {
   implicit def toMinus[T: Group](t: T) = new MinusOp(t)
   implicit def toTimes[T: Ring](t: T) = new TimesOp(t)
   implicit def toDiv[T: Field](t: T) = new DivOp(t)
-  implicit def toRichTraversable[T](t: Traversable[T]) = new RichTraversable(t)
+  implicit def toRichTraversable[T](t: TraversableOnce[T]) = new RichTraversable(t)
 }
 
 class PlusOp[T: Semigroup](t: T) {


joshualande:
Hmm, it looks my branch didn't build, but I can't figure out what the issue was (https://travis-ci.org/twitter/algebird/jobs/121875937). Does anybody know why it failed?

isnotinvain:
from the log looks like it got stuck:
```
No output has been received in the last 10 minutes, this potentially indicates a stalled build or something wrong with the build itself.
```

If you hit the restart button it may just work.

isnotinvain:
Is `TraversableOnce` the right type here? Reading the scala docs, `TraversableOnce` is the common type between `Iterator` and `Traverseable` so that sounds right -- does that imply that `Traversables` themselves can only be iterated a single time? I can't seem to find that contract.

joshualande:
Hi Alex!

Apparently I would need write access to Algebird to restart the build: http://stackoverflow.com/questions/17606874/trigger-a-travis-ci-rebuild-without-pushing-a-commit. Is there any chance that you or another owner could restart the build?

Concerning your comment about types, the RichTraversable type takes a TraversableOnce so allowing toRichTraversable to take a TraversableOnce will allow a larger range of objects to be passed into the function.

Does that make sense!

Joshua

isnotinvain:
Yeah that makes sense. I'll restart the build for you.

joshualande:
@isnotinvain thanks so much!!!

joshualande:
@isnotinvain looks like all tests pass now!!! Thanks for your help!

joshualande:
@johnynek thanks so much for the help!!!

johnynek:
@joshualande thank you!

PR #519, opened 2016-04-18
make SSOne/SSMany constructors private and add apply method with count

koertkuipers:
These constructors should have been private from the beginning. The documentation already says other constructors should be used instead.
Also a new constructor/apply is added to set the exact count of an item instead of default value of 1

### DIFF
--- a/algebird-core/src/main/scala/com/twitter/algebird/SpaceSaver.scala
+++ b/algebird-core/src/main/scala/com/twitter/algebird/SpaceSaver.scala
@@ -9,6 +9,12 @@ object SpaceSaver {
    */
   def apply[T](capacity: Int, item: T): SpaceSaver[T] = SSOne(capacity, item)
 
+  /**
+   * Construct SpaceSaver with given capacity containing a single item with provided exact count.
+   * This is the public api to create a new SpaceSaver.
+   */
+  def apply[T](capacity: Int, item: T, count: Long): SpaceSaver[T] = SSMany(capacity, Map(item -> (count, 0L)))
+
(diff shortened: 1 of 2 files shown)


algebird-core/src/main/scala/com/twitter/algebird/SpaceSaver.scala
```diff
@@ -101,14 +107,14 @@ object SSMany {
...
-  private def apply[T](capacity: Int, counters: Map[T, (Long, Long)]): SSMany[T] =
+  private[algebird] def apply[T](capacity: Int, counters: Map[T, (Long, Long)]): SSMany[T] =
     SSMany(capacity, counters, bucketsFromCounters(counters))
 
   private[algebird] def apply[T](one: SSOne[T]): SSMany[T] =
     SSMany(one.capacity, Map(one.item -> (1L, 0L)), SortedMap(1L -> Set(one.item)))
 }
 
-case class SSMany[T](capacity: Int, counters: Map[T, (Long, Long)], buckets: SortedMap[Long, Set[T]]) extends SpaceSaver[T] {
+case class SSMany[T] private (capacity: Int, counters: Map[T, (Long, Long)], buckets: SortedMap[Long, Set[T]]) extends SpaceSaver[T] {
```
johnynek: should this be `private[algebird]` why are the others if not?

koertkuipers: the  2 apply methods in companion object are now both called from other classes so they both need to be private[algebird]. the one on line 110 had to be opened up because of new SpaceSaver.apply method on line 16.

the main constructor is not used outside the class. it can also very easily be used incorrectly (e.g. create buckets that are inconsistent with counters), so i rather keep it as private as possible.

PR #530, opened 2016-06-14
Add Batched[A] type for efficient lazy addition

non:
This is the free semigroup -- it represents combining one or more
elements using an associative operation. The values are stored in a
tree structure, which can be traversed. If a semigroup for the
underlying values becomes available, the structure can be summed to a
single value.

Batching represents a space/time trade-off (we use more space to use
less time). In some cases these structures may become too big. To
address this, there are constructors for `Monoid[Batched[A]]` and
`Semigroup[Batched[A]]` instances that will periodically compact
themselves (summing the tree to a single value) to keep the overall
size below a specific parameter.

Batching is appropriate anytime an eager concatenation or combination
will produce expensive intermediate values that will be thrown away, and a more
efficient aggregate summation is possible. In those cases, building up
a batch and then summing it should result in much more efficient
operations.

This commit adds the type, with documentation. It also adds some basic
law-checks and tests.

### DIFF
--- a/algebird-core/src/main/scala/com/twitter/algebird/Batched.scala
+++ b/algebird-core/src/main/scala/com/twitter/algebird/Batched.scala
@@ -126,7 +126,7 @@ object Batched {
    * produce different lists (for instance, if one of the batches has
    * more zeros in it than another one).
    */
-  def equiv[A](implicit s: Semigroup[A]): Equiv[Batched[A]] =
+  implicit def equiv[A](implicit s: Semigroup[A]): Equiv[Batched[A]] =
     new Equiv[Batched[A]] {
       def equiv(x: Batched[A], y: Batched[A]): Boolean = x.sum(s) == y.sum(s)
     }
@@ -285,7 +285,7 @@ object Batched {
    */
   private[algebird] class ReverseItemsIterator[A](root: Batched[A]) extends ItemsIterator[A](root) {
     def descend(v: Batched[A]): A = {
-      @tailrec def descend0(v: Batched[A]): A =
+      @inline @tailrec def descend0(v: Batched[A]): A =
         v match {
           case Items(lhs, rhs) =>
             stack = lhs :: stack
--- a/algebird-core/src/main/scala/com/twitter/algebird/Batched.scala
+++ b/algebird-core/src/main/scala/com/twitter/algebird/Batched.scala
@@ -126,9 +126,10 @@ object Batched {
    * produce different lists (for instance, if one of the batches has
    * more zeros in it than another one).
    */
-  implicit def equiv[A](implicit s: Semigroup[A]): Equiv[Batched[A]] =
+  implicit def equiv[A](implicit e: Equiv[A], s: Semigroup[A]): Equiv[Batched[A]] =
     new Equiv[Batched[A]] {
-      def equiv(x: Batched[A], y: Batched[A]): Boolean = x.sum(s) == y.sum(s)
+      def equiv(x: Batched[A], y: Batched[A]): Boolean =
+        e.equiv(x.sum(s), y.sum(s))
     }
 
   /**
@@ -305,7 +306,7 @@ object Batched {
  * than `batchSize` values in it. When more values are added, the
  * tree is compacted using `s`.
  */
-class BatchedSemigroup[T](batchSize: Int)(implicit sg: Semigroup[T]) extends Semigroup[Batched[T]] {
+class BatchedSemigroup[T: Semigroup](batchSize: Int) extends Semigroup[Batched[T]] {
 
   require(batchSize > 0, s"Batch size must be > 0, found: $batchSize")
 
@@ -320,12 +321,11 @@ class BatchedSemigroup[T](batchSize: Int)(implicit sg: Semigroup[T]) extends Sem
  * than `batchSize` values in it. When more values are added, the
  * tree is compacted using `m`.
  */
-class BatchedMonoid[T](batchSize: Int)(implicit monoid: Monoid[T])
-  extends BatchedSemigroup(batchSize)(monoid) with Monoid[Batched[T]] {
-  val zero: Batched[T] = Batched(monoid.zero)
+class BatchedMonoid[T: Monoid](batchSize: Int) extends BatchedSemigroup[T](batchSize) with Monoid[Batched[T]] {
+  val zero: Batched[T] = Batched(Monoid.zero)
 
   // if we knew that (a+b=0) only for (a=0, b=0), we could instead do:
   //   new Batched.ItemsIterator(b).exists(monoid.isNonZero)
   override def isNonZero(b: Batched[T]): Boolean =
-    monoid.isNonZero(b.sum(monoid))
+    Monoid.isNonZero(b.sum)
 }
--- a/algebird-core/src/main/scala/com/twitter/algebird/Batched.scala
+++ b/algebird-core/src/main/scala/com/twitter/algebird/Batched.scala
@@ -1,0 +1,303 @@
+package com.twitter.algebird
+
+import scala.annotation.tailrec
+
+/**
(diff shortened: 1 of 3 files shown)


algebird-core/src/main/scala/com/twitter/algebird/Batched.scala
```diff
@@ -0,0 +1,303 @@
...
+ * batched values defers the underlying semigroup action, instead
+ * storing all values in memory (in a tree structure). If an
+ * underlying semigroup is available, `Batched.semigroup` and
+ * `Batch.monoid` can be configured to periodically sum the tree to
+ * keep the overall size below `batchSize`.
+ *
+ * `Batched[T]` values are guaranteed not to be empty -- that is, they
+ * will contain at least one `T` value.
+ */
+sealed abstract class Batched[T] {
```
johnynek: can we extend `Serializable`?

algebird-core/src/main/scala/com/twitter/algebird/Batched.scala
```diff
@@ -0,0 +1,303 @@
...
+   * y, x, and then w.
+   */
+  def reverseIterator: Iterator[T] =
+    this match {
+      case Batched.Item(t) => Iterator.single(t)
+      case b => new Batched.ReverseItemsIterator(b)
+    }
+
+  /**
+   * Report the size of the underlying tree structure.
```
johnynek: can we comment this is `O(1)` cost, not `O(N)`.

algebird-core/src/main/scala/com/twitter/algebird/Batched.scala
non: Sure.

algebird-core/src/main/scala/com/twitter/algebird/Batched.scala
```diff
@@ -0,0 +1,303 @@
...
+   * them to flatten the tree.
+   *
+   * It represents the identity using `Batched(m.zero)`, but does not
+   * have a unique representation of zero. For example, `zero combine
+   * zero` has a different tree structure to `zero`, but they both
+   * represent the same value (as seen by `.sum`).
+   */
+  def monoid[A](implicit m: Monoid[A]): Monoid[Batched[A]] =
+    new Monoid[Batched[A]] {
+      val zero: Batched[A] = Batched(m.zero)
```
johnynek: I don't think this is lawful. `plus(zero, a).size != a.size`, right?

algebird-core/src/main/scala/com/twitter/algebird/Batched.scala
```diff
@@ -0,0 +1,303 @@
...
+    }
+
+  /**
+   * Compacting semigroup for batched values.
+   *
+   * This semigroup ensures that the batch's tree structure has fewer
+   * than `batchSize` values in it. When more values are added, the
+   * tree is compacted using `s`.
+   */
+  def semigroup[A](batchSize: Int)(implicit s: Semigroup[A]): Semigroup[Batched[A]] =
```
johnynek: will this name being the same as above cause implicit resolution issues with there is `import Batched._`? I know there is some name insanity (that is a scala bug that would be AWESOME to fix).

algebird-core/src/main/scala/com/twitter/algebird/Batched.scala
```diff
@@ -0,0 +1,303 @@
...
+   * Compacting monoid for batched values.
+   *
+   * This monoid ensures that the batch's tree structure has fewer
+   * than `batchSize` values in it. When more values are added, the
+   * tree is compacted using `m`.
+   */
+  def monoid[A](batchSize: Int)(implicit m: Monoid[A]): Monoid[Batched[A]] =
+    new BatchedMonoid[A](batchSize, m)
+
+  def aggregator[A, B, C](batchSize: Int, agg: Aggregator[A, B, C]): Aggregator[A, Batched[B], C] = new Aggregator[A, Batched[B], C] {
```
johnynek: can we comment what this one does? It is batching up an Aggregator so that sumOption can be used on the middle type with a given batchsize. This helps when you have a relatively slow `plus` but a fast `sumOption` which possibly uses some temporary mutable state.

algebird-core/src/main/scala/com/twitter/algebird/Batched.scala
johnynek: okay, I see what is going on... we have two different `Eq[Batched[T]]` we have `a.sum == b.sum` and `a.toList == b.toList`.

This worries me a bit. I wonder if we should use two different type name here (`Concat[T]` and `Batched[T]` for the `.toList` and `.sum` cases?) It is easy to get confused which case we are in.

johnynek: what is really weird is that the default semigroup for `Batched` is `toList` one, but the default monoid is `.sum` one. So, that's really confusing.

algebird-core/src/main/scala/com/twitter/algebird/Batched.scala
```diff
@@ -105,6 +116,22 @@ object Batched {
...
+   *
+   * Batches are equivalent if they sum to the same value. Since the
+   * free semigroup is associative, it's not correct to take tree
+   * structure into account when determining equality.
+   *
+   * One thing to note here is that two equivalent batches might
+   * produce different lists (for instance, if one of the batches has
+   * more zeros in it than another one).
+   */
+  def equiv[A](implicit s: Semigroup[A]): Equiv[Batched[A]] =
```
johnynek: can we make this implicit?

algebird-core/src/main/scala/com/twitter/algebird/Batched.scala
non: Done.

algebird-core/src/main/scala/com/twitter/algebird/Batched.scala
```diff
@@ -0,0 +1,331 @@
...
+      descend0(v)
+    }
+  }
+
+  /**
+   * Right-to-left iterator through a batch's tree.
+   */
+  private[algebird] class ReverseItemsIterator[A](root: Batched[A]) extends ItemsIterator[A](root) {
+    def descend(v: Batched[A]): A = {
+      @tailrec def descend0(v: Batched[A]): A =
```
johnynek: why no `@inline` on this one like the above?

johnynek:
👍

johnynek:
looks like a 2.10 compile failure: https://travis-ci.org/twitter/algebird/jobs/137652585#L1517

algebird-core/src/main/scala/com/twitter/algebird/Batched.scala
```diff
@@ -0,0 +1,331 @@
...
+   * free semigroup is associative, it's not correct to take tree
+   * structure into account when determining equality.
+   *
+   * One thing to note here is that two equivalent batches might
+   * produce different lists (for instance, if one of the batches has
+   * more zeros in it than another one).
+   */
+  implicit def equiv[A](implicit s: Semigroup[A]): Equiv[Batched[A]] =
+    new Equiv[Batched[A]] {
+      def equiv(x: Batched[A], y: Batched[A]): Boolean = x.sum(s) == y.sum(s)
```
johnynek: actually his needs `Equiv[A]` too. We should not be using universal equals inside here.

non: Whoops! Heh...

johnynek:
👍  when green

non:
Looks like a stalled build or something. 🚒

johnynek:
Thanks @non! this is great! `Batched` along with `sumOption` pretty much allow fairly good optimization of almost any semigroup (assuming the allocation costs of Batched `<<` cost of semigroup.plus).

PR #531, opened 2016-06-15
Add a default k value for Aggregator.approximatePercentile

joshualande:
This is pretty minor, but QTreeAggregator and QTreeAggregatorLowerBound have default k values, so I think approximatePercentile and approximatePercentileBounds should also get the default k value. This might help out users who are new to QTrees :)

A small bit of testing:

```scala
scala> import com.twitter.algebird.Aggregator
import com.twitter.algebird.Aggregator

scala> val agg = Aggregator.approximatePercentile[Int](0.5)
agg: com.twitter.algebird.QTreeAggregatorLowerBound[Int] = QTreeAggregatorLowerBound(0.5,9)

scala> agg(Seq(1,2,3,4,5))
res4: Double = 3.0
```


johnynek:
seems good to me.

joshualande:
Thanks! Greatly appreciated =)

PR #532, opened 2016-06-16
Add a MapAlgebra.group function

joshualande:
In some personal code I wanted a 'MapAlgebra.group' analog to the 'TypedPipe.group' function:

```scala
def group[K, V](pairs: TraversableOnce[(K, V)]): Map[K, List[V]]
```
I think my implementation seems reasonably efficient, but I would be open to feedback if there was a better solution.

While I was at it, I added some additional documentation to MapAlgebra.sumByKey to mention that it filters out any keys where the values are 0. 

I also slightly updated the MapAlgebra.sumByKey unit test to (a) explicitly test that sumByKey drops values that are 0 and (b) test both the implicit and explicit version of the function.

I hope these changes all seem reasonable!

### DIFF
--- a/algebird-core/src/main/scala/com/twitter/algebird/MapAlgebra.scala
+++ b/algebird-core/src/main/scala/com/twitter/algebird/MapAlgebra.scala
@@ -16,7 +16,7 @@ limitations under the License.
 package com.twitter.algebird
 
 import scala.collection.{ Map => ScMap }
-import scala.collection.mutable.{ Map => MMap }
+import scala.collection.mutable.{ Builder, Map => MMap }
 
 import com.twitter.algebird.macros.{ Cuber, Roller }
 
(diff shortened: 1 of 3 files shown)


algebird-core/src/main/scala/com/twitter/algebird/Operators.scala
```diff
@@ -43,6 +43,9 @@ class RichTraversable[T](t: TraversableOnce[T]) {
   def sumByKey[K, V](implicit ev: <:<[T, (K, V)], sg: Semigroup[V]): Map[K, V] =
     MapAlgebra.sumByKey(t.map { _.asInstanceOf[(K, V)] })
```
johnynek: these is no reason for the map here. Just cast t:
```scala
MapAlgebra.sumByKey(t.asInstanceOf[TraversableOnce(K, V)])
```
Not sure why that map is there. Same below. This is safe because due to the `ev` we know `T <: (K, V)` and `TraversableOnce` is covariant (so it is safe to upcast the type parameter).

We really need a function like:
```scala
def upcast[T[+_], A, B](t: T[A])(implicit ev: A <:< B): T[B] =
  // it is safe to upcast covariant types:
  t.asInstanceOf[T[B]]
```
```
scala> val t: List[Any] = upcast(List[Int](1, 2))
t: List[Any] = List(1, 2)
scala> val t: List[Int] = upcast(List[Any](1, 2))

<console>:12: error: type mismatch;
 found   : List[Any]
 required: List[Int]
       val t: List[Int] = upcast(List[Any](1, 2))
```

algebird-core/src/main/scala/com/twitter/algebird/MapAlgebra.scala
```diff
@@ -187,10 +187,28 @@ object MapAlgebra {
...
   def sumByKey[K, V: Semigroup](pairs: TraversableOnce[(K, V)]): Map[K, V] =
     Monoid.sum(pairs map { Map(_) })
 
+  /**
+   * For each key, creates a list of all values. This function is equivalent to:
+   *
+   *   pairs.groupBy(_._1).mapValues(_.map(_._2))
+   */
+  def group[K, V](pairs: TraversableOnce[(K, V)]): Map[K, List[V]] =
+    sumByKey(pairs.map { case (k, v) => k -> List(v) })
```
johnynek: actually, a few monoids (`List[T]` and `String`) have O(N^2) `plus` operations (repeatedly concatenating is not efficient).

This is why we added `Batched[T]`. I think for a library function like this, we probably need a special purpose method.
Something like:
```scala
    if (items.isEmpty) Map.empty
    else {
      val mutable = MMap[K, V]()
      items.foreach { case (k, v) =>
          case (k, v) =>
            val oldVOpt = mutable.get(k)
            // sorry for the micro optimization here: avoiding a closure
            val bldr = if (oldVOpt.isEmpty) {
                  val b = List.newBuilder[V]
                  mutable.update(k, b)
                  b
               } else oldVOpt.get
            bldr += v
        }
        mutable.map { case (k, bldr) => (k, bldr.result) } : Map[K, List[V]]
      }

```

joshualande:
@johnynek as aways, amazing suggestion! I learn something new every time! I remember learning that insertion in the front of a linked list is O(1) and at the end is O(n), but it wasn't clear to me that we were doing the O(n) thing. Using the mutable list seems much better.

I had to make a few minor cosmetic changes to your code to get it to work including:
* changing the type of the map to MMap[K, Builder[V, List[V]]
* removing a duplicate case (k, v)
* explicitly converting the map from mutable to immutable via a toMap at the end.

I hope this works for you.

Thanks again for your help!

algebird-core/src/main/scala/com/twitter/algebird/MapAlgebra.scala
```diff
@@ -187,10 +187,43 @@ object MapAlgebra {
...
+          val oldVOpt = mutable.get(k)
+          // sorry for the micro optimization here: avoiding a closure
+          val bldr = if (oldVOpt.isEmpty) {
+            val b = List.newBuilder[V]
+            mutable.update(k, b)
+            b
+          } else oldVOpt.get
+          bldr += v
+      }
+      mutable.map { case (k, bldr) => (k, bldr.result) }.toMap
```
johnynek: one minor thing, probably should be: `mutable.iterator.map (...).toMap` so we don't have to build a structure twice.

johnynek:
looks good other than a minor comment (looks like CI got stuck, but otherwise it looks correct).

joshualande:
@johnynek I added the .iterator trick. I will keep that in mind for next time!!!

johnynek:
Thank you sir!

joshualande:
@johnynek, this is great. Thanks for your help!

PR #533, opened 2016-06-16
Optimize count-min sketch

non:
This PR updates the existing `CMS[A]` type to be more efficient, especially when summing many `CMS[A]` structures. The changes are expected to be binary-compatible. The only difference is that more efficient `CMSHasher` instances are available without an import (the import is still available for the legacy instances, which may be needed for old serialized data).

I've also updated the benchmarks to help demonstrate the changes. Here are the benchmarks from before any of the changes:

```
Benchmark                         (eps)  (size)    Cnt    Score     Error  Units
CMSBenchmark.sumLargeBigIntCms      0.1    1000     5   49.939 ±   1.152  ops/s
CMSBenchmark.sumLargeBigIntCms    0.005    1000     5   48.309 ±   2.453  ops/s
CMSBenchmark.sumLargeStringCms      0.1    1000     5   50.682 ±  14.959  ops/s
CMSBenchmark.sumLargeStringCms    0.005    1000     5   51.256 ±   3.108  ops/s
CMSBenchmark.sumSmallBigIntCms      0.1    1000     5  556.247 ±  34.430  ops/s
CMSBenchmark.sumSmallBigIntCms    0.005    1000     5  377.775 ±  34.519  ops/s
CMSBenchmark.sumSmallLongCms        0.1    1000     5  594.712 ±  26.725  ops/s
CMSBenchmark.sumSmallLongCms      0.005    1000     5  449.726 ± 110.672  ops/s
```

and here are the benchmarks after:

```
Benchmark                         (eps)  (size)    Cnt     Score      Error  Units
CMSBenchmark.sumLargeBigIntCms      0.1    1000     5   626.961 ±   32.379  ops/s
CMSBenchmark.sumLargeBigIntCms    0.005    1000     5   573.082 ±  186.829  ops/s
CMSBenchmark.sumLargeStringCms      0.1    1000     5   173.149 ±   64.034  ops/s
CMSBenchmark.sumLargeStringCms    0.005    1000     5   146.868 ±  136.613  ops/s
CMSBenchmark.sumSmallBigIntCms      0.1    1000     5  1369.887 ±  188.736  ops/s
CMSBenchmark.sumSmallBigIntCms    0.005    1000     5  1144.827 ±  238.539  ops/s
CMSBenchmark.sumSmallLongCms        0.1    1000     5  7998.298 ± 6520.702  ops/s
CMSBenchmark.sumSmallLongCms      0.005    1000     5  4708.305 ± 1749.729  ops/s
```

There should also be much less garbage generated in most cases.

### DIFF
--- a/algebird-benchmark/src/main/scala/com/twitter/algebird/benchmark/TopCMSBenchmark.scala
+++ b/algebird-benchmark/src/main/scala/com/twitter/algebird/benchmark/TopCMSBenchmark.scala
@@ -1,0 +1,91 @@
+package com.twitter.algebird.benchmark
+
+import java.util.concurrent.TimeUnit
+import org.openjdk.jmh.annotations._
+import com.twitter.algebird.{ TopPctCMS, CMSHasherImplicits, TopPctCMSMonoid }
+
+/**
+ * Benchmarks the Count-Min sketch implementation in Algebird.
+ *
+ * We benchmark different `K` types as well as different input data streams.
+ */
+
+object TopCMSBenchmark {
+
+  @State(Scope.Benchmark)
+  class CMSState {
+
+    val Seed = 1
+    val JavaCharSizeInBits = 2 * 8
+
+    @Param(Array("0.1", "0.005"))
+    var eps: Double = 0.0
+
+    @Param(Array("0.0000001" /* 1E-8 */ ))
+    var delta: Double = 0.0
+
+    @Param(Array("0.2"))
+    var heavyHittersPct: Double = 0.0
+
+    @Param(Array("1000"))
+    var ops: Int = 0 // Number of operations per benchmark repetition (cf. `reps`)
+
+    @Param(Array("2048"))
+    var maxBits: Int = 0
+
+    var random: scala.util.Random = _
+    var cmsLongMonoid: TopPctCMSMonoid[Long] = _
+    var cmsBigIntMonoid: TopPctCMSMonoid[BigInt] = _
+    var cmsStringMonoid: TopPctCMSMonoid[String] = _
+    var inputsBigInt: Vector[BigInt] = _
+    var inputsString: Vector[String] = _
+
+    @Setup(Level.Trial)
+    def setup(): Unit = {
+      // Required import of implicit values (e.g. for BigInt- or Long-backed CMS instances)
+      import CMSHasherImplicits._
+
+      cmsLongMonoid = TopPctCMS.monoid[Long](eps, delta, Seed, heavyHittersPct)
+      cmsBigIntMonoid = TopPctCMS.monoid[BigInt](eps, delta, Seed, heavyHittersPct)
+      cmsStringMonoid = TopPctCMS.monoid[String](eps, delta, Seed, heavyHittersPct)
+
+      random = new scala.util.Random
+
+      inputsString = (0 to ops).map { i => random.nextString(maxBits / JavaCharSizeInBits) }.toVector
+      Console.out.println(s"Created ${inputsString.size} input records for String")
+      inputsBigInt = inputsString.map { s => BigInt(s.getBytes) }.toVector
+      Console.out.println(s"Created ${inputsBigInt.size} input records for BigInt")
+    }
+  }
+}
+
+class TopCMSBenchmark {
+  import TopCMSBenchmark._
+  // Case A (K=Long): We count the first hundred integers, i.e. [1, 100]
+  @Benchmark
+  def timePlusOfFirstHundredIntegersWithLongCms(st: CMSState) = {
+    val m = st.cmsLongMonoid
+    m.sumOption((1 to st.ops).iterator.map(n => m.create(n)))
+  }
+
+  // Case B.1 (K=BigInt): We count the first hundred integers, i.e. [1, 100]
+  @Benchmark
+  def timePlusOfFirstHundredIntegersWithBigIntCms(st: CMSState) = {
+    val m = st.cmsBigIntMonoid
+    m.sumOption((1 to st.ops).iterator.map(n => m.create(BigInt(n))))
+  }
+
+  // Case B.2 (K=BigInt): We count numbers drawn randomly from a 2^maxBits address space
+  @Benchmark
+  def timePlusOfRandom2048BitNumbersWithBigIntCms(st: CMSState) = {
+    val m = st.cmsBigIntMonoid
+    m.sumOption(st.inputsBigInt.iterator.map(m.create(_)))
+  }
+
+  // Case C (K=String): We count strings drawn randomly from a 2^maxBits address space
+  @Benchmark
+  def timePlusOfRandom2048BitNumbersWithStringCms(st: CMSState) = {
+    val m = st.cmsStringMonoid
+    m.sumOption(st.inputsString.iterator.map(m.create(_)))
+  }
+}
--- a/algebird-benchmark/src/main/scala/com/twitter/algebird/benchmark/TopCMSBenchmark.scala
+++ b/algebird-benchmark/src/main/scala/com/twitter/algebird/benchmark/TopCMSBenchmark.scala
@@ -1,8 +1,9 @@
-package com.twitter.algebird.benchmark
+package com.twitter.algebird
+package benchmark
 
 import java.util.concurrent.TimeUnit
 import org.openjdk.jmh.annotations._
-import com.twitter.algebird.{ TopPctCMS, CMSHasherImplicits, TopPctCMSMonoid }
+import scala.util.Random.nextString
 
 /**
  * Benchmarks the Count-Min sketch implementation in Algebird.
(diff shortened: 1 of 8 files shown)


algebird-benchmark/src/main/scala/com/twitter/algebird/benchmark/TopCMSBenchmark.scala
```diff
@@ -0,0 +1,91 @@
...
+    var cmsLongMonoid: TopPctCMSMonoid[Long] = _
+    var cmsBigIntMonoid: TopPctCMSMonoid[BigInt] = _
+    var cmsStringMonoid: TopPctCMSMonoid[String] = _
+    var inputsBigInt: Vector[BigInt] = _
+    var inputsString: Vector[String] = _
+
+    @Setup(Level.Trial)
+    def setup(): Unit = {
+      // Required import of implicit values (e.g. for BigInt- or Long-backed CMS instances)
+      import CMSHasherImplicits._
```
non: This should go away now. I'll remove it.

algebird-test/src/main/scala/com/twitter/algebird/BaseProperties.scala
```diff
@@ -44,7 +44,7 @@ object BaseProperties {
   def isAssociativeEq[T: Semigroup, U <: T: Arbitrary](eqfn: (T, T) => Boolean) = {
     'isAssociativeEq |: forAll { (a: U, b: U, c: U) =>
       val semi = implicitly[Semigroup[T]]
-      eqfn(semi.plus(a, semi.plus(b, c)), semi.plus(semi.plus(a, b), c))
+      eqfn(semi.plus(a, semi.plus(b, c)), semi.plus(semi.plus(a, b), c)) //&& false
```
johnynek: can we remove this (which we added to test that the tests were running)?

algebird-test/src/main/scala/com/twitter/algebird/BaseProperties.scala
```diff
@@ -53,10 +53,8 @@ object BaseProperties {
 
   def isAssociative[T: Semigroup: Arbitrary] = isAssociativeDifferentTypes[T, T]
 
-  def semigroupSumWorks[T: Semigroup: Arbitrary: Equiv] = 'semigroupSumWorks |: forAll { (in: List[T]) =>
-    in.isEmpty || {
-      Equiv[T].equiv(Semigroup.sumOption(in.iterator).get, in.reduceLeft(Semigroup.plus(_, _)))
-    }
+  def semigroupSumWorks[T: Semigroup: Arbitrary: Equiv] = 'semigroupSumWorks |: forAll { (head: T, tail: List[T]) =>
+    Equiv[T].equiv(Semigroup.sumOption(head :: tail).get, tail.foldLeft(head)(Semigroup.plus(_, _)))
```
johnynek: actually, this is not such a great test, we should probably do:
```scala
Semigroup.sumOption(list.iterator) match {
  case None => list.isEmpty
  case Some(r) => Equiv[T].equiv(r, in.reduce(Semigroup.plus(_, _)))
}
```
but that is an actual change to the law, so we probably should not do it here.

algebird-test/src/test/scala/com/twitter/algebird/CountMinSketchTest.scala
```diff
@@ -172,13 +180,16 @@ class CMSContraMapSpec extends WordSpec with Matchers with GeneratorDrivenProper
     val a = 4
     val b = 0
     val width = 1234
-    val x = Array(1.toByte).toSeq // same as Seq(1.toByte)
-    targetHasher.hash(a, b, width)(x) should be(677)
+    val x = Array(113.toByte).toSeq // same as Seq(1.toByte)
```
johnynek: comment is out of date. Why did this change?

algebird-test/src/test/scala/com/twitter/algebird/CountMinSketchTest.scala
```diff
@@ -347,7 +358,7 @@ class CmsTotalCountProperty[K: CMSHasher: Gen] extends CmsProperty[K] {
 class CmsProperties extends ApproximateProperties("CountMinSketch") {
   import ApproximateProperty.toProp
 
-  implicit val intGen = Gen.choose(1, 100)
+  implicit val intGen = Gen.choose(Int.MinValue, Int.MaxValue)
```
johnynek: didn't we find that `Int.MinValue` was not good, or was it fine in the end?

actually, `(1, 100)` might be better to leave it since it will have collisions, which otherwise we aren't really exercising counting (since all counts will be 0 or 1 with an enormous range).

algebird-test/src/test/scala/com/twitter/algebird/CountMinSketchTest.scala
non: This changed with the new hashing code -- the previous test had hardcoded the old hash result, so it needed to change anyway since we are using a different hasher.

algebird-test/src/main/scala/com/twitter/algebird/BaseProperties.scala
non: OK, we can add that in another PR.

non:
I think that last failure was transient. It passed on 2.10 but failed on 2.11.

rubanm:
Seems to be reproducible on 2.11 odd :/ Something to do with SparkContext not booting I presume.

```
[info] AlgebirdRDDTest:
[info] Exception encountered when attempting to run a suite with class name: com.twitter.algebird.spark.AlgebirdRDDTest *** ABORTED ***
[info]   java.util.concurrent.TimeoutException: Futures timed out after [10000 milliseconds]
[info]   at scala.concurrent.impl.Promise$DefaultPromise.ready(Promise.scala:219)
[info]   at scala.concurrent.impl.Promise$DefaultPromise.result(Promise.scala:223)
[info]   at scala.concurrent.Await$$anonfun$result$1.apply(package.scala:190)
[info]   at scala.concurrent.BlockContext$DefaultBlockContext$.blockOn(BlockContext.scala:53)
[info]   at scala.concurrent.Await$.result(package.scala:190)
[info]   at akka.remote.Remoting.start(Remoting.scala:180)
[info]   at akka.remote.RemoteActorRefProvider.init(RemoteActorRefProvider.scala:184)
[info]   at akka.actor.ActorSystemImpl.liftedTree2$1(ActorSystem.scala:618)
[info]   at akka.actor.ActorSystemImpl._start$lzycompute(ActorSystem.scala:615)
[info]   at akka.actor.ActorSystemImpl._start(ActorSystem.scala:615)
[info]   ...
```

rubanm:
I restarted https://travis-ci.org/twitter/algebird/builds/138421411 to see if this reproduces on develop

rubanm:
The changes LGTM btw.

There are a couple jobs at Twitter from https://github.com/twitter/scalding/issues/1565 that we can run this on. I can cut a build and give this a whirl. I guess this will also need the CMSHasherImplicits import removed from https://github.com/twitter/scalding/blob/develop/scalding-core/src/main/scala/com/twitter/scalding/typed/Sketched.scala#L25

isnotinvain:
What's meant by:

"The only difference is that more efficient CMSHasher instances are available without an import (the import is still available for the legacy instances, which may be needed for old serialized data)."

Does this change the serialization format of something we usually store for long term storage? We don't expect CMSHasher to be serialized + stored right?

non:
@rubanm Even if you don't remove the `import CMSHasher._` statement things should still work. But removing it will potentially speed up your code that is using the CMS.

@isnotinvain My understanding is that `CMS` contains a `CMSParams` which contains `CMSHasher` instances. So if you serialized a `CMS` (at least using Java serialization) you'd probably end up serializing the hashers. I'm honestly not sure if this is something that's likely to happen, or is a problem. But it's something @johnynek mentioned when I discussed this change with him.

non:
@rubanm By the way -- I am working on a PR which should make the sketch code you linked even faster. It's not quite ready (and will require an algebird milestone) but just figured you should know.

johnynek:
The hash function you use is implicitly part of the serialization because
it controls which buckets get set. So changing the hash will scramble and
existing CMS.

This is why the current PR leaves the old hashes where thy were imported,
but made some optimized ones default. Any old code required the import so
this should not break anyone.

On Friday, June 17, 2016, Erik Osheim <notifications@github.com> wrote:

> @rubanm <https://github.com/rubanm> By the way -- I am working on a PR
> which should make the sketch code you linked even faster. It's not quite
> ready (and will require an algebird milestone) but just figured you should
> know.
>
> —
> You are receiving this because you were mentioned.
> <https://github.com/twitter/algebird/pull/533#issuecomment-226921233>, or mute
> the thread
> <https://github.com/notifications/unsubscribe/AAEJdg9Brnv2G1dLcpFT4EbCV2zKdyxxks5qM3QXgaJpZM4I3n3i>
> .
>


-- 

johnynek:
Any objections to merging this?

I'm +1, but there are (remote) compatibility concerns (enabling defaults without imports of hashing of String and BigInt and making them different than they are unimported).

algebird-core/src/main/scala/com/twitter/algebird/CMSHasher.scala
```diff
@@ -0,0 +1,150 @@
...
+    positiveHash % width
+  }
+
+  // This CMSHasher[BigInt] is newer, and faster, than the old version
+  // found in CMSHasherImplicits. Unless you have serialized data that
+  // requires the old implementation for correctness, you should be
+  // using this instance.
+  implicit object CMSHasherBigInt extends CMSHasher[BigInt] {
+    override def hash(a: Int, b: Int, width: Int)(x: BigInt): Int =
+      ((a * x.hashCode + b) & Int.MaxValue) % width
```
johnynek: actually, one concern I have about this hashCode is that if you collide on the `Int` hashCode, you are already sunk, no more mixing will help. You are not going to be using `BigInt` unless you really have big numbers (bigger than Long) for instance, but actually this hash will have a lot more collisions than `Long`.

Since `BigInt` hashing is not super common, I'm not sure we should make this a default when we have a concern about loss of precision.

An alternate approach might be to: `(a * x + b).hashCode & Int.MaxValue` so that we don't hash until we have already done the linear transform. This gives two different a,b values the chance to not collide. I wonder how much slower that will be.

algebird-core/src/main/scala/com/twitter/algebird/CountMinSketch.scala
```diff
@@ -578,8 +652,19 @@ case class CMSInstance[K](countsTable: CMSInstance.CountsTable[K],
...
   }
 
-  def frequency(item: K): Approximate[Long] =
-    makeApprox(countsTable.counts.iterator.zip(params.hashes.iterator).map { case (row, hash) => row(hash(item)) }.min)
+  def frequency(item: K): Approximate[Long] = {
+    var freq = Long.MaxValue
+    val hs = params.hashes
+    val it = countsTable.counts.iterator
+    var i = 0
+    while (it.hasNext) {
```
johnynek: in the case of `freq == 0` we could stop, since we know that `freq >= 0` and `min` can only decrease it. It will only help in the case of missing keys so it might not be worth adding the `while(it.hasNext && (freq > 0))` which might slightly slow down non-zero keys.

non: I'll try this out -- we should have some benchmarks that test `frequency` anyway.

algebird-core/src/main/scala/com/twitter/algebird/CMSHasher.scala
non: I can think of three ways to address this concern:

1. Roll the change back and have hashing `BigInt` be really slow/expensive.
2. Use `.toLong` first so we get at the same # of collisions as `Long` would.
3. Mix in `(x >>> 32).hashCode` and `(x >>> 64).hashCode`. I think the shifting should still be significantly faster than `.toByteArray`.

johnynek: let's roll back the BigInt change for now. It is uncommon to use anyway.

algebird-core/src/main/scala/com/twitter/algebird/CMSHasher.scala
```diff
@@ -121,13 +121,9 @@ object CMSHasher {
...
   }
 
-  // This CMSHasher[BigInt] is newer, and faster, than the old version
-  // found in CMSHasherImplicits. Unless you have serialized data that
-  // requires the old implementation for correctness, you should be
-  // using this instance.
   implicit object CMSHasherBigInt extends CMSHasher[BigInt] {
     override def hash(a: Int, b: Int, width: Int)(x: BigInt): Int =
-      ((a * x.hashCode + b) & Int.MaxValue) % width
+      CMSHasher.hashBytes(a, b, width)(x.toByteArray)
```
johnynek: actually, I think we should move this all back to CMSHasherImplicits and not have a default here until we can see about speeding up the perf (the long trick or maybe `(a * x).hashCode`)

johnynek: I can fix this in a follow up.

non:
It looks to me like the failing property for `DecayedVector` is not related to this PR.

PR #534, opened 2016-06-24
Add a Aggregator.numericSum function

joshualande:
I needed this for some work I was doing, and I don't think it already exists. It is somewhat trivial, but I think it could be useful to put into the Aggregator object in case other people need something similar.

### DIFF
--- a/algebird-core/src/main/scala/com/twitter/algebird/Aggregator.scala
+++ b/algebird-core/src/main/scala/com/twitter/algebird/Aggregator.scala
@@ -284,6 +284,11 @@ object Aggregator extends java.io.Serializable {
    */
   def approximatePercentileBounds[T](percentile: Double, k: Int = QTreeAggregator.DefaultK)(implicit num: Numeric[T]): QTreeAggregator[T] =
     QTreeAggregator[T](percentile, k)
+
+  /**
+   * An aggregator that sums the values using Numeric addition.
+   */
+  def numericSum[T: Numeric]: MonoidAggregator[T, T, T] = Aggregator.fromMonoid(new NumericRing[T])
 }
 
 /**
--- a/algebird-test/src/test/scala/com/twitter/algebird/AggregatorLaws.scala
+++ b/algebird-test/src/test/scala/com/twitter/algebird/AggregatorLaws.scala
@@ -76,6 +76,16 @@ class AggregatorLaws extends CheckProperties {
     }
   }
 
+  def checkNumericSum[T: Numeric: Arbitrary] =
+    forAll { in: List[T] =>
+      val aggregator = Aggregator.numericSum
+      aggregator(in) == in.sum
+    }
+  property("Aggregator.numericSum is correct for Ints") { checkNumericSum[Int] }
+  property("Aggregator.numericSum is correct for Longs") { checkNumericSum[Long] }
+  property("Aggregator.numericSum is correct for Doubles") { checkNumericSum[Double] }
+  property("Aggregator.numericSum is correct for Floats") { checkNumericSum[Float] }
+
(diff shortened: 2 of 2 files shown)


avibryant:
Does it not already work to use `Aggregator.fromMonoid[T]` for this case? I would expect the implicit `NumericRing[T]` to kick in for that.

joshualande:
Hi Avi!

I think what you said is totally correct, and my code is redundant/unnecessary.

As long as you make sure to import the Ring.numericRing implicit def, things work:

```scala
scala> import com.twitter.algebird.Ring._
import com.twitter.algebird.Ring._
scala> def mySum[T: Numeric] = {
     |   val num = implicitly[Numeric[T]]
     |   val numbers: Seq[T] = (1 to 3).map(num.fromInt)
     |   val agg: Aggregator[T, T, T] = Aggregator.fromMonoid[T]
     |   agg(numbers)
     | }
mySum: [T](implicit evidence$1: Numeric[T])T
scala> mySum[Int]
res28: Int = 6
scala> mySum[Long]
res29: Long = 6
scala> mySum[Float]
res30: Float = 6.0
scala> mySum[Double]
res31: Double = 6.0
```

Given that fromMonoid is doing the right thing, I think this pull request is unnecessary and I will close it.

That being said, I am somewhat familiar with the Aggregator code base and couldn't figure it out myself :). So hopefully these comments can help people out if they need to do something similar in the future!

Thanks for your help!

johnynek:
By the way, I'm not opposed to adding this method. I agree it may help
discoverability. That said, Monoid is actually faster (probably) than
Numeric because Monoid is specialized (meaning the primitive types don't
need to be boxed in Objects to be used in a generic).

On Friday, June 24, 2016, Joshua Lande <notifications@github.com> wrote:

> Closed #534 <https://github.com/twitter/algebird/pull/534>.
>
> —
> You are receiving this because you are subscribed to this thread.
> <https://github.com/twitter/algebird/pull/534#event-703460161>, or mute
> the thread
> <https://github.com/notifications/unsubscribe/AAEJdnV_CINTLZEOfFLU7A-_rrBY80NZks5qPBFTgaJpZM4I9nGs>
> .
>


-- 

joshualande:
@johnynek thanks for your suggestions! 

The benefit of Numeric over Monoid for me was that Numeric has a toDouble function which I needed for an internal class I was writing (i.e. to be able to perform a t-test on Ints, Longs, Floats, or Doubles). But overall, when possible I agree it is more general/efficient to allow Monoids.

In terms of this PR, I would be happy to add this function to the Aggregator class if people found it useful-enough. Or alternately, I would be happy to add a comment to Aggregator.fromMonoid explaining how it can be used for Numerics via the required import. Or alternately, I am happy to leave the discussion in this PR and hope that Google makes it discoverable :) Whatever is fine with me.

Thanks for your help!

Joshua

avibryant:
How about this, then?

````scala
def numericSum[T](implicit monoid: Monoid[T], numeric: Numeric[T]) =
  Aggregator.fromMonoid[T].composePrepare(numeric.toDouble(_))
````

That gets you the specialized `Monoid` instance for the summing, but the standard `Double` output from `Numeric`. And totally justifies having this as a special method.

joshualande:
@avibryant that's a great suggestion! I will give it a shot and see how it goes!

johnynek:
I think that works if you add `def numericSum[@specialize(Int, Long, Float, Double) T]` otherwise you lose the specialized.

joshualande:
@johnynek @avibryant I had to create a new PR with the suggested changes (due to the issue discussed [here](https://github.com/isaacs/github/issues/361)). See the new PR [here](https://github.com/twitter/algebird/pull/539).

PR #535, opened 2016-06-25
Minor fixer for backwards compatibility (proposed 0.12.1)

johnynek:
This does a few things:

1. adds mima backwards compatibility check as part of CI (long LONG overdue)
2. fixes some issues mima found.
3. removes the default `CMSHasher[BigInt]` which I am concerned was too weak of a hasher as it projects to `Int` before doing any mixing of the value `a`. We have a couple of proposals to improve that, but we should be sure we have something fast and correct before we make it default.

/cc @rubanm @isnotinvain @non @ianoc

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,4 +6,4 @@ matrix:
       script: ./sbt ++$TRAVIS_SCALA_VERSION clean test
 
     - scala: 2.11.7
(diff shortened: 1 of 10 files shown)


ianoc:
👍

rubanm:
👍  the implicit wiring is probably missing something going by the travis logs

non:
👍 cms changes seem fine to me

PR #537, opened 2016-06-27
optimize CMS.create(Seq[K])

johnynek:
overlooked this in #533 

/cc @non

### DIFF
--- a/algebird-core/src/main/scala/com/twitter/algebird/CountMinSketch.scala
+++ b/algebird-core/src/main/scala/com/twitter/algebird/CountMinSketch.scala
@@ -163,6 +163,7 @@ class CMSSummation[K](params: CMSParams[K]) {
       offset += width
       row += 1
     }
+    totalCount += count
   }
 
   def updateAll(sketches: TraversableOnce[CMS[K]]): Unit =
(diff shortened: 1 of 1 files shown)


non:
LGTM 👍

Good catch! 😊

algebird-core/src/main/scala/com/twitter/algebird/CountMinSketch.scala
```diff
@@ -124,7 +124,11 @@ class CMSMonoid[K: CMSHasher](eps: Double, delta: Double, seed: Int, maxExactCou
   /**
    * Creates a sketch out of multiple items.
    */
-  def create(data: Seq[K]): CMS[K] = data.foldLeft(zero) { case (acc, x) => plus(acc, create(x)) }
+  def create(data: Seq[K]): CMS[K] = {
+    val summation = new CMSSummation(params)
+    data.foreach { k => summation.insert(k, 1L) }
```
ianoc: Don't think it really matters, but doing an iterator here would be faster to avoid closure creation?

ianoc:
LGTM

rubanm:
👍

algebird-core/src/main/scala/com/twitter/algebird/CountMinSketch.scala
isnotinvain: yeah, while loop here too?

isnotinvain:
+1

algebird-core/src/main/scala/com/twitter/algebird/CountMinSketch.scala
johnynek: since this is only one closure for the whole data and since the insert call is already expensive, I'm going to punt on this (since I'm not sure it will actually be faster). I can imagine the jit inlining this.

PR #538, opened 2016-06-27
Add sumOption to Short,Int,Long,Float,Double

johnynek:
This can avoid some cases of double boxing. When we use for instance, `Semigroup[T]` and `T` happens to be `Long`, then we can still box despite the specialized call because `T` was generic where the `Semigroup` was used (I believe that is true).

Doing `sum` with plus means, unboxing the accumulator and the next item, adding them, boxing the result and starting again. This gives about `N` boxing operations and `2N` unboxing operations for `N` items to sum. If we do sum internally, we know the type, so we only have to unbox from the `TraverableOnce`, which gives `N` unboxing operations.

### DIFF
--- a/algebird-core/src/main/scala/com/twitter/algebird/Field.scala
+++ b/algebird-core/src/main/scala/com/twitter/algebird/Field.scala
@@ -54,6 +54,14 @@ object FloatField extends Field[Float] {
     assertNotZero(r)
     l / r
   }
+  override def sum(t: TraversableOnce[Float]): Float = {
+    var sum = 0.0f
+    t.foreach(sum += _)
+    sum
+  }
+  override def sumOption(t: TraversableOnce[Float]): Option[Float] =
+    if (t.isEmpty) None
+    else Some(sum(t))
 }
 
 object DoubleField extends Field[Double] {
@@ -67,6 +75,14 @@ object DoubleField extends Field[Double] {
     assertNotZero(r)
     l / r
   }
+  override def sum(t: TraversableOnce[Double]): Double = {
+    var sum = 0.0
+    t.foreach(sum += _)
+    sum
+  }
+  override def sumOption(t: TraversableOnce[Double]): Option[Double] =
+    if (t.isEmpty) None
+    else Some(sum(t))
 }
 
 object BooleanField extends Field[Boolean] {
--- a/algebird-core/src/main/scala/com/twitter/algebird/Ring.scala
+++ b/algebird-core/src/main/scala/com/twitter/algebird/Ring.scala
@@ -86,9 +86,9 @@ object ShortRing extends Ring[Short] {
   override def minus(l: Short, r: Short) = (l - r).toShort
   override def times(l: Short, r: Short) = (l * r).toShort
   override def sum(t: TraversableOnce[Short]): Short = {
-    var sum: Short = 0
-    t.foreach { s => sum = (sum + s).toShort }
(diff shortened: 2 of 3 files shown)


johnynek:
mima crashed on this. Trying to bump to the latest and see if that helps.

ianoc:
LGTM -- did you manage to get any benchmark/profile to illustrate?

algebird-core/src/main/scala/com/twitter/algebird/Field.scala
```diff
@@ -54,6 +54,14 @@ object FloatField extends Field[Float] {
     assertNotZero(r)
     l / r
   }
+  override def sum(t: TraversableOnce[Float]): Float = {
+    var sum = 0.0f
+    t.foreach(sum += _)
```
isnotinvain: did we find out in the CMS PR that foreach doesn't compile down to a while?
should this be a while loop? I could imagine dynamic dispatch on a closure costingmore than boxing, but who knows

ianoc: Closure usage is far far cheaper than boxing. I guess its to do with inlining or new gen allocations. But boxing will show up orders of magnitude higher than it. But seems reasonable to eliminate it here

algebird-core/src/main/scala/com/twitter/algebird/Ring.scala
```diff
@@ -77,6 +85,14 @@ object ShortRing extends Ring[Short] {
   override def plus(l: Short, r: Short) = (l + r).toShort
   override def minus(l: Short, r: Short) = (l - r).toShort
   override def times(l: Short, r: Short) = (l * r).toShort
+  override def sum(t: TraversableOnce[Short]): Short = {
+    var sum: Short = 0
+    t.foreach { s => sum = (sum + s).toShort }
```
isnotinvain: is .toShort a method call or is it like a cast in java? 

maybe sum into an Int and then .toShort at the end?

algebird-core/src/main/scala/com/twitter/algebird/Field.scala
isnotinvain: OK thanks, didn't really have an intuition for what the cost of method dispatch vs boxing (object allocation?) is. But yeah, seems like here we might as well use a while anyway.

isnotinvain:
+1, 2 minor comments if you think they are worth addressing (but without benchmarks, not really worth too much thought probably)

algebird-core/src/main/scala/com/twitter/algebird/Field.scala
johnynek: The only catch is `TraversableOnce` has no way to use `while` without `toIterator`. In that case, building an iterator can be more expensive that `foreach`. Consider an array. If you need to write `foreach` for an array, you can do:
```scala
def foreach(f: T => Unit): Unit = {
  var idx = 0
  while(idx < arr.length) {
   f(arr(idx))
    idx += 1
  }
}
```
Which can be cheaper than wrapping with an Iterator.

I think we should assume foreach is fine until we really want to geek out on different data structure testing.

isnotinvain: that's a good point, didn't realize that TraversableOnce only has a conversion to Iterator.

isnotinvain:
+1

PR #539, opened 2016-06-29
Add a new Aggregator.numericSum function (second try)

joshualande:
I had to create a new PR with the suggested changes (due to the issue discussed [[here](https://github.com/isaacs/github/issues/361)](https://github.com/isaacs/github/issues/361)). Here was the previous [PR](https://github.com/twitter/algebird/pull/534).

This PR creates a numericSum aggregator which aggregates numeric values into Doubles.

Note that @johnynek suggested making this function specialized, but I tried that and the scala compiler complained.

The code was:
```scala
def numericSum[@specialized(Int, Long, Float, Double) T](implicit num: Numeric[T]): MonoidAggregator[T, Double, Double] =
  Preparer[T].map(num.toDouble).monoidAggregate(Aggregator.fromMonoid)
```
And the error was:
```
[warn] /Users/jlande/work/github/algebird/algebird-core/src/main/scala/com/twitter/algebird/Aggregator.scala:298: type T is unused or used in non-specializable positions.
[warn]   def numericSum[@specialized(Int, Long, Float, Double) T](implicit num: Numeric[T]): MonoidAggregator[T, Double, Double] =

I suspect the issue is that the num.toDouble function is not specialized so the compiler cannot achieve the specialization optimization.

### DIFF
--- a/algebird-core/src/main/scala/com/twitter/algebird/Aggregator.scala
+++ b/algebird-core/src/main/scala/com/twitter/algebird/Aggregator.scala
@@ -284,6 +284,20 @@ object Aggregator extends java.io.Serializable {
    */
   def approximatePercentileBounds[T](percentile: Double, k: Int = QTreeAggregator.DefaultK)(implicit num: Numeric[T]): QTreeAggregator[T] =
     QTreeAggregator[T](percentile, k)
+
+  /**
+   * An aggregator that sums Numeric values into Doubles.
+   *
+   * Note that if you instead wanted to aggregate Numeric values of a type T into the same type T
+   * (e.g. if you want MonoidAggregator[T, T, T] for some Numeric type T), you can directly use
+   * Aggregator.fromMonoid[T] after importing the numericRing implicit:
+   *
+   *   > import com.twitter.algebird.Ring.numericRing
+   *   > def numericAggregator[T: Numeric]: MonoidAggregator[T, T, T] = Aggregator.fromMonoid[T]
+   */
+  def numericSum[T](implicit num: Numeric[T]): MonoidAggregator[T, Double, Double] =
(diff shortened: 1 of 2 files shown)


avibryant:
I'd still prefer the approach of summing with the native `Monoid[T]` and only converting to double as part of the presentation. For one thing, this way is going to be accumulating floating point error.

johnynek:
Okay, this is why specialization is pretty terrible in practice. I don't think `MonoidAggregator` is specialized, so the whole thing falls apart.

It is tough exactly when do to the `n.toDouble`. If the types are all integers, there should not be accumulation of error (Double can store integers of magnitude less that 2^53). If the types is `Float` it is already more noisy. In the case of BigInt or Long with big numbers, you would lose precision.

On the other hand, if you keep it as `T` and have something like `Short` (or even `Int`) it could roll over before the conversion to `Double` which is weird. As soon as you have semantics that we convert to `Double` I think the normal modulus behavior of primitive integers is not what you expect.

What do you both think of this?

I wonder if we should just have an alias of `sum` to point at `fromMonoid`?

joshualande:
Sorry for the delay in response. 

To me, the biggest risk is summing Shorts and Ints which can more easily overflow. In principle, it would likely be best to convert these types to Longs before summing, not Doubles. But I am not sure how much we can except from this general-purpose function.

I am happy to go with whatever we think is best.

johnynek:
I'm happy with this. Can we add to the comment that this is really no more than converting to `Double` and then summing. The conversion to double means we don't have the overflow semantics of integer types on the jvm (e.g. `Int.MaxValue + 1 == Int.MinValue`).

joshualande:
@johnynek I am supper sorry I didn't get around to addressing your comment! Here is another PR which adds your helpful comments https://github.com/twitter/algebird/pull/543.

Thank you for your help on this!

Issue #528, opened 2016-06-08
Many property-based tests are broken


non:
If you change the associativity property to always return `false`, you should see every law check fail. Instead, many of them continue passing, because they are never actually doing any work. This is due to failing to mix-in the correct traits into the test classes.

To fix this, find the tests that pass under a bogus associativity check, change them all so they fail, and then revert the associativity check. The reporter recommends that all the Algebird property-checks extend a single `AlgebirdProperties` trait to ensure that they all have the same mix-ins, to help avoid this issue in the future.

isnotinvain:
Thanks for finding this. @johnynek this has happened before. I think we need to find a way to use our test library that prevents this case of "the tests are quietly not being executed"

isnotinvain:
@non single base trait is a good start, we should definitely standardize everything to that and use only a single scalatest "style"

sritchie:
Looks like we hit this a while ago: https://github.com/twitter/algebird/issues/357

sritchie:
Closed by #584. `TopPCTCmsLaws` and `BatchedLaws` were the remaining broken tests.

PR #541, opened 2016-08-03
Improve the build and test times and add Codecov.io

johnynek:
I have reduced the cost of the CMS tests that were dominating before with (probably needlessly) huge CMS allocations. I also turned down the sizes for QTree and SketchMap that were needlessly huge taking forever.

The hope is that this would expose #528 and shame us into fixing the issue.


rubanm:
The 2.11 runs timed out after running the tests. Restarted them to see if this was transient.

johnynek:
I think maybe the tests are just too slow when instrumentation is on. Maybe need to find slow tests. :/

algebird-test/src/test/scala/com/twitter/algebird/CountMinSketchTest.scala
```diff
@@ -373,6 +373,9 @@ abstract class CMSTest[K: CMSHasher: FromIntLike] extends WordSpec with Matchers
   val EPS = 0.001
   val SEED = 1
 
+  private[this] val maxDepth = 70
```
benpence: Use these in the changes to the `Gen`s above?

algebird-test/src/test/scala/com/twitter/algebird/CountMinSketchTest.scala
```diff
@@ -784,50 +787,30 @@ abstract class CMSTest[K: CMSHasher: FromIntLike] extends WordSpec with Matchers
...
-          cms2.heavyHitters should be(Set(2, 3, 4, 5).toK[K])
-
-          val cms3 = TopNCMS.monoid[K](depth, width, seed, 3).create(data1)
-          cms3.heavyHitters should be(Set(3, 4, 5).toK[K])
-
-          val cms4 = TopNCMS.monoid[K](depth, width, seed, 2).create(data1)
-          cms4.heavyHitters should be(Set(4, 5).toK[K])
-
-          val cms5 = TopNCMS.monoid[K](depth, width, seed, 1).create(data1)
-          cms5.heavyHitters should be(Set(5).toK[K])
```
benpence: Are these being removed because of redundancy?

johnynek: Yes. They are tested by the replaced line (we only check one for each of
the checks, but that is fine).

algebird-test/src/test/scala/com/twitter/algebird/CountMinSketchTest.scala
johnynek: It's a different class and I didn't want to yak shave too much. Just want the minimal change to enable code coverage to work.

PR #542, opened 2016-08-09
Add Committers.md

isnotinvain:
Per https://github.com/twitter/analytics-infra-governance

### DIFF
--- a/COMMITTERS.md
+++ b/COMMITTERS.md
@@ -1,0 +1,19 @@
+# Committers
+
+Please see our [Project Governance](https://github.com/twitter/analytics-infra-governance) page for more details.
+
+## Active
+
+| Name                   | Handle                                                    |
+|------------------------|-----------------------------------------------------------|
+| Alex Levenson          | [@isnotinvain](https://github.com/isnotinvain)            |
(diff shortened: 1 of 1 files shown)


isnotinvain:
If we're happy with this format, I'll send PRs to the rest of the repos as well

johnynek:
Looks pretty good, I guess. I hate making such lists, but I hope people will remember to check the graphs page to see the many contributions from so many generous people.

johnynek:
👍

benpence:
+1 if we're worried about not giving due credit to previous committers, could we rename this to convey that these people have merge permissions (that's the case right)? Or is the file name important?

isnotinvain:
To clarify, this is not a list of contributors, but rather a list of who can +1 a PR (which is a pretty commonly used term). It would be nice if there were an even more clear term for this but committer is pretty well established.

We have this:

It is not necessary to become a committer in order to contribute to the Analytics Infrastructure OSS Ecosystem. Anyone is welcome to contribute to Twitter's OSS projects, and pull requests and bug reports are greatly appreciated. 

in https://github.com/twitter/analytics-infra-governance

But would could make that even more explicit with a pointer to the contributors page

PR #544, opened 2016-08-27
add Semigroup.maybePlus

johnynek:
this comes up often, and we see it in user code quite a bit.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -39,7 +39,7 @@ val sharedSettings = Project.defaultSettings ++ scalariformSettings ++  Seq(
     "-unchecked",
     "-deprecation",
     "-optimize",
-    "-Xlint:-adapted-args",
+    "-Xlint",
     "-language:implicitConversions",
     "-language:higherKinds",
     "-language:existentials"),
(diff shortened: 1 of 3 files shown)


build.sbt
```diff
@@ -35,7 +35,14 @@ val sharedSettings = Project.defaultSettings ++ scalariformSettings ++  Seq(
 
   javacOptions ++= Seq("-target", "1.6", "-source", "1.6"),
 
-  scalacOptions ++= Seq("-unchecked", "-deprecation", "-optimize", "-Xlint", "-language:implicitConversions", "-language:higherKinds", "-language:existentials"),
+  scalacOptions ++= Seq(
+    "-unchecked",
+    "-deprecation",
+    "-optimize",
+    "-Xlint:-adapted-args",
```
benpence: `scala -Xlint:help` says that `adapter-args` is for "Warn if an argument list is modified to match the receiver." Is that to warn on implicit conversions?

johnynek: it is `Right(1, 2)` which is adapted to be `Right((1, 2))` because `Right(t: T)` takes one arg. This code is littered with that and it makes warnings totally overwhelm the errors (so I couldn't even find the errors).

It does not work on 2.10, so I'll roll that back, but we really need to clean this up.

benpence:
+1

Issue #545, opened 2016-08-27
way too many warnings


johnynek:
We have a ton of adapted arg warnings.

We need to fix those and many others. Just compile the code, not the insane number of warnings, and lament.

jnievelt:
Are you actively working on this?  Sounds like something I could pick up as background work.

sritchie:
Thanks @jnievelt !

PR #546, opened 2016-08-29
Cleanup warnings

jnievelt:
This addresses most of the annoying/messy/large warnings (for #545)

Remaining warnings include macro warnings and deprecated methods warnings.


johnynek:
LGTM when it is green.

jnievelt:
It didn't like the `AbstractIterable` change, even though it's a private class.  I've reverted that and we're green.

PR #550, opened 2016-09-23
Fixes #549 - Use repartition instead of coalesce for compat with spark 2

### LINKED ISSUE #549: Incompatibilities with spark 2
Algebird is built against spark 1.3, until spark 2 most of the code was working fine.
However in spark 2 the method signature of [coalesce changed](https://github.com/apache/spark/blob/branch-2.0/core/src/main/scala/org/apache/spark/rdd/RDD.scala#L438). For example we encounter this error when using sumOption:

```
java.lang.NoSuchMethodError: org.apache.spark.rdd.RDD.coalesce$default$3(IZ)Lscala/math/Ordering;
        at com.twitter.algebird.AlgebirdRDD$.sumOption$extension(AlgebirdRDD.scala:115)
```

EugenCepoi:
Fixes https://github.com/twitter/algebird/issues/549.
Luckily repartition is equivalent to coalesce with shuffle true and it's signature didn't change. So the fix is pretty simple...

### DIFF
--- a/algebird-spark/src/main/scala/com/twitter/algebird/spark/AlgebirdRDD.scala
+++ b/algebird-spark/src/main/scala/com/twitter/algebird/spark/AlgebirdRDD.scala
@@ -25,7 +25,7 @@ class AlgebirdRDD[T](val rdd: RDD[T]) extends AnyVal {
         Iterator(agg.appendAll(b, data))
       }
     }, preservesPartitioning = true)
-    pr.coalesce(1, shuffle = true)
+    pr.repartition(1)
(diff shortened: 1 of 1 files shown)


johnynek:
Thanks!

johnynek:
@EugenCepoi I just published 0.12.2 if you want to pick this up.

EugenCepoi:
Thank you @johnynek you have been much faster than what I was hoping for! :)

johnynek:
It is not always so fast! ;) When we don't change binary compatibility, it is easy.

PR #555, opened 2016-10-18
Add Diff data structure to algebird-core

sritchie:
This PR adds `Diff`, a data structure for tracking additions and deletions to sets. Applying a diff to set `S` has the same effect as if you'd individually added each item in the `add` set contained by the diff and removed each item in the `remove` set.

`Diff` forms a monoid!

A couple of examples:

```scala
scala> import com.twitter.algebird._
import com.twitter.algebird._

scala> (Diff.addAll(Set(1,2,3)) merge Diff.removeAll(Set(2,3)))(Set.empty)
res0: Set[Int] = Set(1)

scala> (Diff.addAll(Set(1,2,3)) merge Diff.removeAll(Set(2,3)) merge Diff.add(2))(Set.empty)
res1: Set[Int] = Set(1, 2)
```

`Diff` has a few interesting properties. If you calculate a `Diff` off of two sets, the diff is always invertible:

```scala
val diff = Diff.of(a, b)
a == diff.invert(b)
```

This property allows you to break down and reconstruct the history of state changes of a series of sets, forward from the first set or backward from the final set.

### DIFF
--- a/algebird-core/src/main/scala/com/twitter/algebird/SetDiff.scala
+++ b/algebird-core/src/main/scala/com/twitter/algebird/SetDiff.scala
@@ -17,19 +17,19 @@ limitations under the License.
 package com.twitter.algebird
 
 /**
- * `Diff` is a class that represents changes applied to a set. It is
- * in fact a Set[T] => Set[T], but doesn't extend Function1 since
+ * `SetDiff` is a class that represents changes applied to a set. It
+ * is in fact a Set[T] => Set[T], but doesn't extend Function1 since
  * that brings in a pack of methods that we don't necessarily want.
  */
-sealed abstract case class Diff[T] private (add: Set[T], remove: Set[T]) { self =>
-  def +(t: T): Diff[T] = Diff(add + t, remove - t)
-  def -(t: T): Diff[T] = Diff(add - t, remove + t)
-  def ++(ts: Iterable[T]): Diff[T] = Diff(add ++ ts, remove -- ts)
-  def --(ts: Iterable[T]): Diff[T] = Diff(add -- ts, remove ++ ts)
-  def merge(other: Diff[T]): Diff[T] = {
+sealed abstract case class SetDiff[T] private (add: Set[T], remove: Set[T]) { self =>
+  def +(t: T): SetDiff[T] = SetDiff(add + t, remove - t)
+  def -(t: T): SetDiff[T] = SetDiff(add - t, remove + t)
+  def ++(ts: Iterable[T]): SetDiff[T] = SetDiff(add ++ ts, remove -- ts)
+  def --(ts: Iterable[T]): SetDiff[T] = SetDiff(add -- ts, remove ++ ts)
+  def merge(other: SetDiff[T]): SetDiff[T] = {
     val newAdd = (add ++ other.add) -- other.remove
     val newRem = (remove -- other.add) ++ other.remove
-    Diff(newAdd, newRem)
+    SetDiff(newAdd, newRem)
   }
 
   /**
@@ -41,7 +41,7 @@ sealed abstract case class Diff[T] private (add: Set[T], remove: Set[T]) { self 
    * Returns a diff that, if applied to a set, undoes the effects of
    * this diff.
    */
-  def invert: Diff[T] = Diff(remove, add)
+  def invert: SetDiff[T] = SetDiff(remove, add)
 
   /**
    * Same as apply, but fails to None if the diff's removal set has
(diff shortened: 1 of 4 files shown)


johnynek:
👍 merge when green.

algebird-core/src/main/scala/com/twitter/algebird/SetDiff.scala
```diff
@@ -0,0 +1,89 @@
...
+ */
+
+package com.twitter.algebird
+
+/**
+ * `SetDiff` is a class that represents changes applied to a set. It
+ * is in fact a Set[T] => Set[T], but doesn't extend Function1 since
+ * that brings in a pack of methods that we don't necessarily want.
+ */
+sealed abstract case class SetDiff[T] private (add: Set[T], remove: Set[T]) { self =>
```
jnievelt: This construction seems a bit awkward.  What about this sort of alternative?
```
sealed trait SetDiff[T] {
  ...
}

private case class SetDiffImpl[T](add: Set[T], remove: Set[T]) extends SetDiff[T] {
  ...
}
```

algebird-core/src/main/scala/com/twitter/algebird/SetDiff.scala
```diff
@@ -0,0 +1,89 @@
...
+  def +(t: T): SetDiff[T] = SetDiff(add + t, remove - t)
+  def -(t: T): SetDiff[T] = SetDiff(add - t, remove + t)
+  def ++(ts: Iterable[T]): SetDiff[T] = SetDiff(add ++ ts, remove -- ts)
+  def --(ts: Iterable[T]): SetDiff[T] = SetDiff(add -- ts, remove ++ ts)
+  def merge(other: SetDiff[T]): SetDiff[T] = {
+    val newAdd = (add ++ other.add) -- other.remove
```
jnievelt: Nit: slightly better to always do -- first?

algebird-core/src/main/scala/com/twitter/algebird/SetDiff.scala
```diff
@@ -0,0 +1,89 @@
...
+    SetDiff(newAdd, newRem)
+  }
+
+  /**
+   * Applies the contained diff to the supplied set.
+   */
+  def apply(previous: Set[T]): Set[T] = previous ++ add -- remove
+
+  /**
+   * Returns a diff that, if applied to a set, undoes the effects of
```
jnievelt: May want to be extra careful here, since this is looking very much like a `Bijection[Set[T], Set[T]]`, though of course it would not obey the `Bijection` laws.

sritchie: By careful you mean, change the name of the method?

algebird-core/src/main/scala/com/twitter/algebird/SetDiff.scala
sritchie: Hey! Thanks for the review - I was cueing off of the approach here: https://gist.github.com/tpolecat/a5cb0dc9adeacc93f846835ed21c92d2

Can you explain what you mean by awkward?

johnynek: the benefit of this style is that it preserves pattern matching, which is nice, but otherwise disables all the other means to create a new instance.

algebird-core/src/main/scala/com/twitter/algebird/SetDiff.scala
johnynek: note `add` and `remove` are disjoint. It does not matter which you do first. No GC can happen of any entries until `other` and `this` go out of scope, so, I don't *think* there is an issue.

algebird-core/src/main/scala/com/twitter/algebird/SetDiff.scala
johnynek: I see what you mean. since the `foo(bar)` and `foo.invert(bar)` both work, it can look similar from that perspective to an Injection or Bijection.

I'm not too worried. Are you @jnievelt ?

johnynek:
@jnievelt note this can be used in summingbird to maintain graphs where the fan-out is small. So, if we don't mind paying the cost of fetching the whole `Diff` on each cache miss.

There could be many applications like this. I think it is really cool.

algebird-core/src/main/scala/com/twitter/algebird/SetDiff.scala
jnievelt: Ah just seems odd to me to have everything anonymous class instances.  But maybe it's not a big deal?

W.r.t. pattern matching, I have a hard time weighing that vs. encapsulating the mechanism here (is there a clear use case for decomposing?).

jnievelt:
Yeah agree this could be very useful.  Could also generalize into a keyed structure like:
```
case class MapDiff[K, V](add: Map[K, V], rem: Set[K]) { ... }
```

sritchie:
@jnievelt what do you think? Good to merge?

sritchie:
Very cool observation on the keyed structure by the way.

johnynek:
wow, yeah, `MapDiff` could be very interesting. Good call @jnievelt !

PR #557, opened 2016-10-24
remove seed variable in BloomFilter and rename k to hashIndex

sritchie:
Same as #220!

I'm GUESSING this might not be legit - do we want to nuke the seed, or do we want to actually try and use it?


sritchie:
(fails binary compatibility check, of course)

johnynek:
I think this is a good PR. we accepted this implementation as a contribution and didn't look back too much (or at least accepted a revamp of it).

This looks like a good improvement. I don't know why you need a seed, except to perhaps salt the hash function, for a reason I don't know. Maybe to allow multiple independent blooms, but why not just use one bigger one there, I'm sure it would perform better?

In any case. I would say merge this on the next binary break (which will be when we merge #523

algebird-core/src/main/scala/com/twitter/algebird/BloomFilter.scala
```diff
@@ -361,19 +361,19 @@ case class BFHash(numHashes: Int, width: Int, seed: Long = 0L) extends Function1
     (upper, lower)
   }
 
-  private def nextHash(bytes: Array[Byte], k: Int, digested: Seq[Int] = Seq.empty): Stream[Int] = {
-    if (k == 0)
+  private def nextHash(bytes: Array[Byte], hashIndex: Int, digested: Seq[Int] = Seq.empty): Stream[Int] = {
```
oscar-stripe, line 364: we should optimize this. This is terrible using `Stream` here.

Paging @non

johnynek:
looks like we need to remove mima for the tests.

isnotinvain:
What gets used instead of the passed in seed? A fixed default one? Only thing I can think of is unit tests wanting a fixed seed or something?

johnynek:
We never used the seed at all.

isnotinvain:
Oh I see, there's no randomness involved in the first place right?

sritchie:
@isnotinvain yup, that's correct

sritchie:
Nice, This failed because the doc compilation failed!

johnynek:
Sweet!

johnynek:
:+1:

PR #566, opened 2016-10-25
upgrade to new scalatest and scalacheck

mosesn:
## Problem
I'm trying to upgrade a monorepo to scalatest 3.x and scalacheck 1.13.1.  Because our repo depends on algebird-test, we now need a version of algebird that supports scalacheck 1.13.1.  This will also allow algebird to begin supporting 2.12, which should be released in the next few weeks.

## Solution
I'm cutting you over to scalatest 3.0.0 and scalacheck 1.13.1.  I tried only moving over scalacheck, but scalatest is compiled against a single version of scalacheck, and will break if you switch.

Some of the Prop api has changed–NB that Prop is now sealed, so we need to use the Prop object to create a new Prop.

## Other notable issues
I've run into a problem with the Eventually tests that I'm not really equipped to solve.  It looks like algebird was using the Gen[A => B] implicit from scalacheck, but it used to be really dumb, and only supported constant functions.  It's [less dumb][0] now, which breaks this test.  I don't know how Eventually is actually supposed to behave, so I'd love some help from y'all.  I've enabled edits from maintainers, so please feel free to just fix it.

## Protip
This change has a lot of whitespace changes–append ?w=1 to the url to ignore the whitespace changes.  Thanks!!!

[0]: https://github.com/rickynils/scalacheck/pull/171

### DIFF
--- a/algebird-test/src/main/scala/com/twitter/algebird/ApproximateProperty.scala
+++ b/algebird-test/src/main/scala/com/twitter/algebird/ApproximateProperty.scala
@@ -52,49 +52,47 @@ object ApproximateProperty {
       }
 
   def toProp(a: ApproximateProperty, objectReps: Int, inputReps: Int, falsePositiveRate: Double): Prop =
-    new Prop {
-      def apply(params: Gen.Parameters) = {
-        require(0 <= falsePositiveRate && falsePositiveRate <= 1)
+    Prop { params: Gen.Parameters =>
+      require(0 <= falsePositiveRate && falsePositiveRate <= 1)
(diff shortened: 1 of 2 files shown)


johnynek:
Thanks Moses! I'll look at the test failure and see if I can find the fix.
Ideally it is a real bug! But probably just a bad test.

mosesn:
@johnynek thanks Oscar!  Please let me know if there's anything else you'd like me to do in this branch.

johnynek:
replaced by #567

PR #552, opened 2016-10-05
Bump scalacheck + scalatest versions

piyushnarang:
If this is good, can follow up with PRs for scalding & summingbird as well.

PR #567, opened 2016-10-27
Fix bug found with scalacheck 1.13

johnynek:
This replaces #566 which Moses started.

See that PR for more discussion.

### DIFF
--- a/algebird-core/src/main/scala/com/twitter/algebird/Eventually.scala
+++ b/algebird-core/src/main/scala/com/twitter/algebird/Eventually.scala
@@ -42,12 +42,12 @@ class EventuallySemigroup[E, O](convert: O => E)(mustConvert: O => Boolean)(impl
   override def plus(x: Either[E, O], y: Either[E, O]) = {
     x match {
       case Left(xe) => y match {
-        case Left(ye) => left(Semigroup.plus(xe, ye))
-        case Right(yo) => left(Semigroup.plus(xe, convert(yo)))
+        case Left(ye) => left(eventualSemigroup.plus(xe, ye))
+        case Right(yo) => left(eventualSemigroup.plus(xe, convert(yo)))
       }
       case Right(xo) => y match {
-        case Left(ye) => left(Semigroup.plus(convert(xo), ye))
-        case Right(yo) => conditionallyConvert(Semigroup.plus(xo, yo))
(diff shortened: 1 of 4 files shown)


johnynek:
@mosesn what do you think?

Actually, this is a legit failure. There are subtleties to equivalence that this uncovers

mosesn:
Glad to hear it surfaced a real bug!  I'm sure @non will be pleased.  @isnotinvain do you mind taking a look at this on my behalf?  I'm not sure I understand `Eventually`.

johnynek:
so, one issue is: this is really a binary break, even though mima can't see it. The test module will depend on a new scalacheck that is not compatible with the previous.

So, we should not merge this until we are ready for such a break. I want to do one soon to merge the algebra dependencies, so maybe that is when we do it.

mosesn:
@johnynek can you publish it under a different name?  we would really like to get this in, we need it to upgrade to scalatest 3.x, which we need for 2.12.

johnynek:
I wonder if you can ask Sriram's Core data libraries team to help?

I'm a bit underwater at the moment.

piyushnarang:
hey @mosesn / @johnynek thanks for spinning this up. I took a stab at this a couple of weeks back and got stuck on the Eventually test and got sidetracked (https://github.com/twitter/algebird/pull/552). We could pull in the algebird sha into our monorepo once this PR is merged. We have taken that approach for other libraries like Scalding / Summingbird as well instead of an official RC.

johnynek:
In that case, I'll merge this now, and we can merge the algebra dep next (so it won't complicate things for you folks).

We need to start 2.12-izing our stack here.

mosesn:
Thanks @piyushnarang!

isnotinvain:
Thanks all. @johnynek do you know if any of our other libraries publish test libraries? Does bijection? We'll probably have to do that one at the same time I would think.

johnynek:
By the way, I think the law we need for eventually is something like:

```
if mustConvert(a) then mustConvert(a + b)
```
for all `b`. This would seem to imply EventuallyGroup is not really a thing, because due to the inverse there is always some `b` such that `a+b=c` for any `c`, which implies either `mustConvert ` is a constant.

I think it is not hard to show this to be true using the sumOption laws and associativity, if so, there is no Group, and hence no Ring.

isnotinvain:
ok thanks

PR #556, opened 2016-10-18
Generate Algebird microsite via sbt-microsites

sritchie:
This PR adds a new, beautiful (if I do say so myself!) microsite for Algebird using the [SBT-Microsites plugin](https://github.com/47deg/sbt-microsites).

I've also converted the Algebird wiki into an executable set of markdown files using the wonderful [tut](https://github.com/tpolecat/tut) library from @tpolecat, and added a build step to check that the microsite compiles. Any code change that breaks the documentation will trigger a build failure.

 I was inspired by the [the new Cats microsite](https://typelevel.org/cats), and a long-slumbering urge to write lots of excellent documentation for algebird's traits and data structures. Tut and the new microsite are the perfect foundation for small, crisp, code-reviewable documentation contributions for existing and new data structures.

The PR also:
- Adds a "Resources for Learners" section to the microsite with links to every talk, paper, tutorial and set of slides I could find on Algebird.
- Bumps the `sbt-mima-plugin` to `0.1.11`, fixing this issue that was biting me on this PR and the #568 PR: https://github.com/typesafehub/migration-manager/issues/112
- Adds a bare-bones "cookbook" section. The plan is to add tutorials here from an "You want to get X done? Try these data structures!" viewpoint, vs the typical "Here's what this data structure can do for you!" approach we've taken.

### How to run the microsite
I couldn't figure out how to get this to push to the github pages section of my local fork, so you'll have to run the site yourself to give it a look.

Here's the magical incantation:

```
brew install jekyll
cd $algebird_home_dir
sbt docs/makeMicrosite
cd docs/target/site
jekyll serve
```

PR #568, opened 2016-10-29
Exponential Histogram implementation

sritchie:
This PR implements the exponential histogram, or DGIM, algorithm from this paper: http://ilpubs.stanford.edu:8090/504/1/2001-34.pdf

The code for `ExpHist` and `ExpHist.Canonical` have a good description of what's going on and how the algorithm works. The supplied `tut`-based documentation has a description of the l-canonical representation procedure.

### Next Steps
- [ ] Add a Monoid implementation based on the algorithm presented here http://megaslides.com/doc/1974/ecm-sketches
- [ ] wrap this impl with a queue that buffers N items before pushing them all into the `ExpHist`, and update relevant error calculations.
- [ ] Get some benchmarks going. I'm going to propose that we do this in the next PR.

cc @non @johnynek


algebird-core/src/main/scala/com/twitter/algebird/ExpHist.scala
```diff
@@ -0,0 +1,342 @@
...
+    }
+  }
+
+  /**
+   * Expands `s` out into a list of numbers that all sum to s. Each
+   * entry is a power of 2, and the number of entries of each power
+   * of 2 matches `s`'s l-canonical representation (for the supplied
+   * l).
+   */
+  def bucketsFromLong(s: Long, l: Int): Vector[Long] = {
```
sritchie: @non @johnynek would love some advice on how to efficiently share the calculations going on here and in `fromLong`, without creating an intermediate return tuple for the values that both functions need.

That said... `bucketsFromLong` is the only thing we're actually using in the implementation. I could just rebuild `fromLong` from the results of `bucketsFromLong`.

sritchie: this is exactly `toBuckets(fromLong(s, k))` btw

algebird-core/src/main/scala/com/twitter/algebird/ExpHist.scala
```diff
@@ -0,0 +1,342 @@
...
+  // the right side, and that `items.map(_.size).sum == delta`.
+  private[ExpHist] def addAllWithoutStep(items: Vector[Bucket], delta: Long): ExpHist = {
+    val inputs = items ++ buckets
+    val desiredBuckets = Canonical.bucketsFromLong(total + delta, conf.l)
+    copy(
+      buckets = ExpHist.rebucket(inputs, desiredBuckets),
+      total = total + delta)
+  }
+
+  def oldestBucketSize: Long = if (total == 0) 0L else buckets.last.size
```
sritchie: I wonder if it'd be a good idea to skip the zero checking by adding a `trait ExpHist` and a `case object EmptyExpHist`.

algebird-core/src/main/scala/com/twitter/algebird/ExpHist.scala
```diff
@@ -0,0 +1,342 @@
...
+    implicit val ord: Ordering[Bucket] = Ordering.by { b: Bucket => (b.timestamp, b.size) }
+  }
+
+  /**
+   * ExpHist guarantees that the returned guess will be within
+   * `epsilon` relative error of the true count across a sliding
+   * window of size `windowSize`.
+   */
+  case class Config(epsilon: Double, windowSize: Long) {
+    val k: Int = math.ceil(1 / epsilon).toInt
```
sritchie: we never really need `k`, but it shows up in the paper.

algebird-core/src/main/scala/com/twitter/algebird/ExpHist.scala
```diff
@@ -0,0 +1,342 @@
...
+   * @return Vector with buckets, or pieces of buckets, with sizes
+   *         totalling `toDrop` items removed from the head. If an
+   *         element wasn't fully consumed, the remainder will be
+   *         stuck back onto the head.
+   */
+  @tailrec private[this] def drop(toDrop: Long, input: Vector[Bucket]): Vector[Bucket] = {
+    val (b @ Bucket(count, _)) +: tail = input
+    (toDrop - count) match {
+      case 0 => tail
+      case x if x < 0 => b.copy(size = -x) +: tail
```
sritchie: this is the only reason I'm not able to use an iterator, this pushing back into the vector. I wonder if we can implement this better with some carry Option. thoughts?

algebird-core/src/main/scala/com/twitter/algebird/ExpHist.scala
```diff
@@ -0,0 +1,342 @@
...
+  def addAll(unsorted: Vector[Bucket]): ExpHist =
+    if (unsorted.isEmpty) this
+    else {
+      val sorted = unsorted.sorted(Ordering[Bucket].reverse)
+      val delta = sorted.map(_.size).sum
+      val timestamp = sorted.head.timestamp
+      if (delta == 0)
+        step(timestamp)
+      else {
+        addAllWithoutStep(sorted, delta).step(timestamp)
```
sritchie: this was a little subtle. We have to step AFTER adding because the range of timestamps of the new items we're pushing in might be larger than windowSize.

algebird-core/src/main/scala/com/twitter/algebird/ExpHist.scala
```diff
@@ -0,0 +1,342 @@
...
+   */
+  def inc(timestamp: Long): ExpHist = add(1L, timestamp)
+
+  /**
+   * Increment ExpHist by delta at the supplied timestamp.
+   */
+  def add(delta: Long, timestamp: Long): ExpHist = {
+    val self = step(timestamp)
+    if (delta == 0) self
+    else self.addAllWithoutStep(Vector(Bucket(delta, timestamp)), delta)
```
sritchie: here we can step BEFORE, because we know that there's only a single timestamp coming in.

algebird-core/src/main/scala/com/twitter/algebird/ExpHist.scala
```diff
@@ -0,0 +1,342 @@
...
+
+  /**
+   * Estimate of the count seen across the last conf.windowSize
+   * timestamps. Guaranteed to be within conf.epsilon of the true
+   * count.
+   */
+  def guess: Double =
+    if (total == 0) 0.0
+    else (total - (oldestBucketSize - 1) / 2.0)
+
```
oscar-stripe: can you add a `Approximate[Long]` return value? Those compose somewhat.

algebird-core/src/main/scala/com/twitter/algebird/ExpHist.scala
```diff
@@ -0,0 +1,342 @@
...
+ * - rebucketing newBuckets ++ oldBuckets into those bucket sizes
+ *
+ * The resulting sequence of buckets is a valid exponential
+ * histogram.
+ */
+object Canonical {
+  @inline private[this] def floorPowerOfTwo(x: Long): Int =
+    JLong.numberOfTrailingZeros(JLong.highestOneBit(x))
+
+  @inline private[this] def modPow2(i: Int, exp2: Int): Int = i & ((1 << exp2) - 1)
```
oscar-stripe: what about the bugs on inlining you found? Are we sure none of these exhibit this? Downstream folks still could if they compile with `-optimize`.

algebird-core/src/main/scala/com/twitter/algebird/ExpHist.scala
```diff
@@ -0,0 +1,342 @@
...
+   *
+   * returns a vector of the the coefficients of s^i in the
+   * l-canonical representation of s.
+   *
+   * the "l" means that
+   *
+   * - ret(i) for all i < j == l or l + 1
+   * - ret(j) < l + 1
+   */
+  def fromLong(s: Long, l: Int): Vector[Int] = {
```
oscar-stripe: no curlies...

algebird-core/src/main/scala/com/twitter/algebird/ExpHist.scala
```diff
@@ -0,0 +1,342 @@
...
+
+  /**
+   * (i = vector index, j = index of last entry)
+   *
+   * returns a vector of the the coefficients of s^i in the
+   * l-canonical representation of s.
+   *
+   * the "l" means that
+   *
+   * - ret(i) for all i < j == l or l + 1
```
oscar-stripe: what is `ret`? The returned array? Can we comment to that?

algebird-core/src/main/scala/com/twitter/algebird/ExpHist.scala
```diff
@@ -0,0 +1,342 @@
...
+  @inline private[this] def bit(i: Int, idx: Int): Int = (i >>> idx) & 1
+
+  private[this] def binarize(i: Int, bits: Int, offset: Int): Vector[Int] =
+    (0 until bits).map { idx => offset + bit(i, idx) }.toVector
+
+  /**
+   * (i = vector index, j = index of last entry)
+   *
+   * returns a vector of the the coefficients of s^i in the
+   * l-canonical representation of s.
```
oscar-stripe: this confuses me. If you are returning the coefficients of `s^i` in the `i'th` position of the vector, then why isn't the vector infinite? We can take infinite powers?

algebird-core/src/main/scala/com/twitter/algebird/ExpHist.scala
oscar-stripe: what about `final case class CanonicalRepresentation(toVector: Vector[Int]) extends AnyVal`

Also, maybe add these specialized types under `object ExpHist`.

algebird-core/src/main/scala/com/twitter/algebird/ExpHist.scala
```diff
@@ -0,0 +1,342 @@
...
+    }
+  }
+
+  /**
+   * Expands out an l-canonical representation into the original number.
+   *
+   * @param rep l-canonical representation of some number s for some l
+   * @return The original s
+   */
+  def toLong(rep: Vector[Int]): Long =
```
oscar-stripe: can this be a method on an `AnyVal` class?

algebird-core/src/main/scala/com/twitter/algebird/ExpHist.scala
oscar-stripe: can you add that comment: `bucketsFromLong(s, l) == toBuckets(fromLong(s, k))`

This is a different representation right? Can we have a different AnyVal wrapper to distinguish?

algebird-test/src/test/scala/com/twitter/algebird/ExpHistLaws.scala
```diff
@@ -0,0 +1,281 @@
...
+      val fullViaAdd = buckets.foldLeft(ExpHist.empty(conf)) {
+        case (e, Bucket(c, t)) => e.add(c, t)
+      }
+
+      val actualSum = actualBucketSum(buckets, cutoff)
+      checkCoreProperties(fullViaAdd, actualSum)
+    }
+  }
+
+  def isPowerOfTwo(i: Long): Boolean = (i & -i) == i
```
oscar-stripe: fancy! :) Hacker's Delight in the house.

algebird-core/src/main/scala/com/twitter/algebird/ExpHist.scala
sritchie: fixed up

algebird-core/src/main/scala/com/twitter/algebird/ExpHist.scala
sritchie: boom, will do

sritchie: don't we want `approximate[Double]`, since we want the estimate to fall in the middle of the range?

algebird-core/src/main/scala/com/twitter/algebird/ExpHist.scala
sritchie: huge typo! should have been 2^i. Fixed up the docs too.

algebird-core/src/main/scala/com/twitter/algebird/ExpHist.scala
sritchie: removed

sritchie: I like it. I moved the whole object inside `ExpHist` - will add these as well.

algebird-core/src/main/scala/com/twitter/algebird/ExpHist.scala
sritchie: boom, moved that and `toBuckets`

sritchie:
Btw the test failure in the `update paper link` push is fixed by the sbt-mima-plugin bump to 0.1.11 in https://github.com/twitter/algebird/pull/556.

algebird-core/src/main/scala/com/twitter/algebird/ExpHist.scala
sritchie: okay, added!

algebird-core/src/main/scala/com/twitter/algebird/ExpHist.scala
```diff
@@ -0,0 +1,405 @@
...
+     *
+     * @param rep l-canonical representation of some number s for some l
+     * @return The original s
+     */
+    def toLong: Long =
+      if (rep.isEmpty) 0L
+      else {
+        rep.iterator.zipWithIndex
+          .map { case (i, exp) => i.toLong << exp }
+          .reduce(_ + _)
```
johnynek: there is going to be a fair amount of boxing cost here. Actually calling `Monoid.sum( )` will avoid some of that since `Monoid` is specialized on `Long`

algebird-core/src/main/scala/com/twitter/algebird/ExpHist.scala
```diff
@@ -0,0 +1,405 @@
...
+
+    /**
+     * Expands out the l-canonical representation of some number s into
+     * a list of bucket sizes in ascending order.
+     *
+     * @param rep l-canonical representation of some number s for some l
+     * @return vector of powers of 2 (where ret.sum == the original s)
+     */
+    def toBuckets: Vector[Long] =
+      rep.zipWithIndex.flatMap { case (i, exp) => List.fill(i)(1L << exp) }
```
johnynek: what about:
```scala
rep.iterator
  .zipWithIndex
  .flatMap { case (i, exp) => Iterator.fill(i)(1L << exp) }
  .toVector
```

algebird-core/src/main/scala/com/twitter/algebird/ExpHist.scala
```diff
@@ -0,0 +1,405 @@
...
+        addAllWithoutStep(sorted, delta).step(timestamp)
+      }
+    }
+
+  /**
+   * Returns a [[Fold]] instance that uses `add` to accumulate deltas
+   * into this exponential histogram instance.
+   */
+  def fold: Fold[Bucket, ExpHist] =
+    Fold.foldLeft(this) {
```
johnynek: I wonder about using a `foldMutable` here and accumulating a batch of `Bucket` and then doing an `addAll` which might be a lot faster, but we can punt on this if you like.

algebird-core/src/main/scala/com/twitter/algebird/ExpHist.scala
sritchie: done

algebird-core/src/main/scala/com/twitter/algebird/ExpHist.scala
sritchie: how about

```scala
Monoid.sum {
        rep.iterator.zipWithIndex
          .map { case (i, exp) => i.toLong << exp }
      }
```

algebird-core/src/main/scala/com/twitter/algebird/ExpHist.scala
```diff
@@ -0,0 +1,405 @@
...
+   * Returns an empty instance with the supplied Config.
+   */
+  def empty(conf: Config): ExpHist = ExpHist(conf, Vector.empty, 0L, 0L)
+
+  /**
+   *  Returns an instance directly from a number `i`. All buckets in
+   *  the returned ExpHist will have the same timestamp, equal to
+   *  `ts`.
+   */
+  def from(i: Long, ts: Long, conf: Config): ExpHist = {
```
sritchie: yeah, let me thread it through and see what it looks like

algebird-core/src/main/scala/com/twitter/algebird/ExpHist.scala
```diff
@@ -0,0 +1,405 @@
...
+   * Efficiently add many buckets at once.
+   *
+   * @param unsorted vector of buckets. All timestamps must be >= this.time.
+   * @return ExpHist instance with all buckets added, stepped
+   *         forward to the max timestamp in `unsorted`.
+   */
+  def addAll(unsorted: Vector[Bucket]): ExpHist =
+    if (unsorted.isEmpty) this
+    else {
+      val sorted = unsorted.sorted(Ordering[Bucket].reverse)
```
sritchie: done

algebird-core/src/main/scala/com/twitter/algebird/ExpHist.scala
johnynek: 👍

algebird-core/src/main/scala/com/twitter/algebird/ExpHist.scala
sritchie: added a test and this impl:

```scala
def fold: Fold[Bucket, ExpHist] =
    Fold.foldMutable[Builder[Bucket, Vector[Bucket]], Bucket, ExpHist](
      { case (b, bucket) => b += bucket },
      { _ => Vector.newBuilder[Bucket] },
      { x => addAll(x.result) })
```

way better imo.

sritchie:
Okay @johnynek, addressed all comments. I think we should add the Monoid in the next round, so we can experiment with controlling the relative error. We'll probably need to track the error separately from the calculation I have here...

We might need to create an `ApproximateHistogram` trait that we can use for the `ExpHist`, the `Windows` implementation, the `ZeroExpHist` and the `ExpHist` that keeps a queue of items around before adding.

sritchie:
Amazing, the docs build failed after my change!! Fixing now.

oscar-stripe:
So good! Love tut docs! Thanks for an excellent example for us to follow @sritchie !

👍 

Any concerns @isnotinvain ?

sritchie:
Thanks for the review, @johnynek! @isnotinvain, would love a final sign off.

johnynek:
I think the rules say it is fine to merge now.

PR #575, opened 2016-11-15
Improve HLLSeries performance.

non:
This commit improves the performance of the HLLSeries data
structure. In particular, it introduces two new methods:

    series.insert(bytes, timestamp)
    series.approximateSizeSince(threshold)

Previously to add data to an HLLSeries required allocating another
series and then combining them with the monoid, potentially creating a
lot of garbage. The `insert` method does the same thing but without an
intermediate series.

Similarly, getting an approximate size used to require calling
`.toHLL` to build an HLL structure, and then asking that. By
abstracting out the machinery to actually do the HLL calculation,
we're able to do this directly from HLLSeries with
`approximateSizeSince`.

There are several other internal efficiency improvements. Some
independent benchmarks I've run show a 4-5x speed up in building
series' from many individual events, and a 3-4x speed up in
calculating the approximate sizes.

The data in the series and the serialization format are unchanged.

### DIFF
--- a/algebird-core/src/main/scala/com/twitter/algebird/HyperLogLog.scala
+++ b/algebird-core/src/main/scala/com/twitter/algebird/HyperLogLog.scala
@@ -93,7 +93,7 @@ object HyperLogLog {
     var need = bits
     while (i < bytes.length && need >= 0) {
       val byte = bytes(i) & 0xff
-      val limit = Integer.min(8, need)
+      val limit = java.lang.Math.min(8, need)
       var j = 0
       while (j < limit) {
         sum += ((byte >>> (7 - j)) & 1) << (i * 8 + j)
--- a/algebird-core/src/main/scala/com/twitter/algebird/HyperLogLog.scala
+++ b/algebird-core/src/main/scala/com/twitter/algebird/HyperLogLog.scala
@@ -20,7 +20,7 @@ import java.nio.ByteBuffer
 import java.lang.Math
 
 /** A super lightweight (hopefully) version of BitSet */
-@deprecated("This is no longer used.", since = "1.12.3")
+@deprecated("This is no longer used.", since = "0.12.3")
 case class BitSetLite(in: Array[Byte]) {
   def contains(x: Int): Boolean = {
     // Pretend 'in' is little endian so that the bitstring b0b1b2b3 is
@@ -81,7 +81,7 @@ object HyperLogLog {
   @inline
   def twopow(i: Int): Double = Math.pow(2.0, i)
 
-  @deprecated("This is no longer used. Use j(Array[Byte], Int) instead.", since = "1.12.3")
+  @deprecated("This is no longer used. Use j(Array[Byte], Int) instead.", since = "0.12.3")
   def j(bsl: BitSetLite, bits: Int): Int =
     j(bsl.in, bits)
 
@@ -106,7 +106,7 @@ object HyperLogLog {
     sum
   }
 
-  @deprecated("This is no longer used. Use rhoW(Array[Byte], Int) instead.", since = "1.12.3")
+  @deprecated("This is no longer used. Use rhoW(Array[Byte], Int) instead.", since = "0.12.3")
   def rhoW(bsl: BitSetLite, bits: Int): Byte =
     rhoW(bsl.in, bits)
 
--- a/algebird-core/src/main/scala/com/twitter/algebird/HyperLogLog.scala
+++ b/algebird-core/src/main/scala/com/twitter/algebird/HyperLogLog.scala
@@ -19,6 +19,7 @@ package com.twitter.algebird
 import java.nio.ByteBuffer
 
 /** A super lightweight (hopefully) version of BitSet */
+@deprecated("this is no longer used", since = "1.12.3")
 case class BitSetLite(in: Array[Byte]) {
   def contains(x: Int): Boolean = {
     // Pretend 'in' is little endian so that the bitstring b0b1b2b3 is
@@ -26,7 +27,7 @@ case class BitSetLite(in: Array[Byte]) {
     // 1 is in the bitset.
     val arrayIdx = x / 8
     val remainder = x % 8
-    ((in(arrayIdx) >> (7 - remainder)) & 1) == 1
+    ((in(arrayIdx) >>> (7 - remainder)) & 1) == 1
   }
 }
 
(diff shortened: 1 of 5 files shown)


algebird-core/src/main/scala/com/twitter/algebird/HyperLogLog.scala
```diff
@@ -79,34 +80,65 @@ object HyperLogLog {
   @inline
   def twopow(i: Int): Double = java.lang.Math.pow(2.0, i)
 
+  @deprecated("this is no longer used", since = "1.12.3")
```
thomas-stripe: May be nice to provide an alternative here: `use j(Array[Byte], Int) instead` or something.

algebird-core/src/main/scala/com/twitter/algebird/HyperLogLog.scala
```diff
@@ -231,6 +261,34 @@ object HyperLogLog {
...
+    //
+    // See https://github.com/twitter/algebird/issues/284
+    if (e > smallE || zeroCnt == 0) e
+    else size * scala.math.log(size.toDouble / zeroCnt)
+  }
+
+  def asApprox(bits: Int, v: Double): Approximate[Long] = {
+    val stdev = 1.04 / scala.math.sqrt(twopow(bits))
+    val lowerBound = math.floor(math.max(v * (1.0 - 3 * stdev), 0.0)).toLong
+    val upperBound = math.ceil(v * (1.0 + 3 * stdev)).toLong
```
oscar-stripe: actually I guess the `floor` and `ceil` should be reversed. We can't widen the bound and still claim the probability is the same. I think this is a (minor) off-by-one bug.

algebird-core/src/main/scala/com/twitter/algebird/HyperLogLogSeries.scala
```diff
@@ -45,10 +92,10 @@ case class HLLSeries(bits: Int, rows: Vector[Map[Int, Long]]) {
     if (rows.isEmpty)
       SparseHLL(bits, Map())
     else
-      rows.zipWithIndex.map{
+      rows.iterator.zipWithIndex.map {
```
oscar-stripe: why don't we want `sumOption` here rather than `reduce(_ + _)`? Or some logic that checks if there are more than `k` things to aggregate use `sumOption`?

algebird-core/src/main/scala/com/twitter/algebird/HyperLogLog.scala
```diff
@@ -79,34 +80,65 @@ object HyperLogLog {
...
-        accum
-      } else if (bsl.contains(pos)) {
-        loop(pos + 1, accum + (1 << pos))
-      } else {
-        loop(pos + 1, accum)
+  def j(bytes: Array[Byte], bits: Int): Int = {
+    var i = 0
+    var sum = 0
+    var need = bits
+    while (i < bytes.length && need >= 0) {
```
thomas-stripe: I don't think this does anything if `need == 0`, so you can just `need > 0`.

algebird-core/src/main/scala/com/twitter/algebird/HyperLogLog.scala
```diff
@@ -79,34 +80,65 @@ object HyperLogLog {
...
-      if (bsl.contains(pos)) zeros
-      else loop(pos + 1, zeros + 1)
-    loop(bits, 1).toByte
+  def rhoW(bytes: Array[Byte], bits: Int): Byte = {
+    var i = bits / 8 // tracks the position in bytes
+    var j = 7 - (bits % 8) // tracks the bit position in bytes(i)
+    var zeros = 1 // start with a single zero
+    while (i < bytes.length) {
+      while (j >= 0) {
+        if (((bytes(i) >>> j) & 1) == 1) return zeros.toByte
```
thomas-stripe: nit: feel free to ignore, but I think having `return` on its own line is worth it here. Mainly, I saw the 0.toByte below, figured there was a `return` somewhere, but wasn't able to easily locate the return statement in this code.

algebird-core/src/main/scala/com/twitter/algebird/HyperLogLogSeries.scala
```diff
@@ -31,12 +31,59 @@ package com.twitter.algebird
...
+
+  def maxRhowStats(threshold: Long): (Int, Double) = {
+    val seen = scala.collection.mutable.Set.empty[Int]
+    var sum: Double = 0.0
+    var i = rows.size - 1
+    while (i >= 0) {
+      val it = rows(i).iterator
+      while (it.hasNext) {
+        val (k, t) = it.next
+        if (t >= threshold && !seen(k)) {
```
thomas-stripe: Perf nit: Is the `!seen` worth it here? If the set already contains `k`, then `seen += k` should be functionally equivalent to a containment check, with no mutation anyway. I guess the question is, is `negativePowersOfTwo` expected to be slower than checking for membership in a hash table?

algebird-core/src/main/scala/com/twitter/algebird/HyperLogLog.scala
non: I'm happy to change that, but the bug seems to have been present before. Do you think it's worth fixing it here, or merging this PR (without differences) then making that change?

algebird-core/src/main/scala/com/twitter/algebird/HyperLogLog.scala
non: Agreed.

algebird-core/src/main/scala/com/twitter/algebird/HyperLogLogSeries.scala
non: I think we need to avoid double-counting a particular `k`. If we remove the check then I think `sum` will be wrong.

tixxit: Ah, of course, for some reason I missed that we were adding it into sum :| That said, you can still skip this check by using `add`, which returns `false` if the set already contains the element:

```scala
if (t >= threshold && seen.add(k)) {
  sum += HyperLogLog.negativePowersOfTwo(i + 1)
}
```

non: Ooh, nice!

non:
Looks like I may have broken downsampling HLLs. I'll investigate more, I don't remember seeing this test fail earlier.

non:
I think I've responded to all the review comments. Please let me know if there's anything else you'd like to see.

algebird-core/src/main/scala/com/twitter/algebird/HyperLogLog.scala
```diff
@@ -17,16 +17,18 @@ limitations under the License.
 package com.twitter.algebird
 
 import java.nio.ByteBuffer
+import java.lang.Math
 
 /** A super lightweight (hopefully) version of BitSet */
+@deprecated("This is no longer used.", since = "1.12.3")
```
oscar-stripe: I think this should be `0.12.3`

algebird-core/src/main/scala/com/twitter/algebird/HyperLogLog.scala
```diff
@@ -77,36 +79,69 @@ object HyperLogLog {
   }
 
   @inline
-  def twopow(i: Int): Double = java.lang.Math.pow(2.0, i)
+  def twopow(i: Int): Double = Math.pow(2.0, i)
+
+  @deprecated("This is no longer used. Use j(Array[Byte], Int) instead.", since = "1.12.3")
```
oscar-stripe: `0.12.3`

algebird-test/src/test/scala/com/twitter/algebird/HyperLogLogSeriesTest.scala
```diff
@@ -2,24 +2,67 @@ package com.twitter.algebird
...
+  }
+
+  // this is a deterministic test to ensure that our rates are staying
+  // within the expected error bounds.
+  property("verify error rate") {
+    def verify(limit: Int): Boolean = {
+      val it = (0 until limit).iterator
+      val h = monoid.sum(it.map(i => monoid.create(int2Bytes(i), i)))
+      val n = h.since(0L).toHLL.approximateSize.estimate
+      val delta = (limit * 0.2).toInt
```
oscar-stripe: where is the `20%` coming from? Can you add a comment?

PR #580, opened 2016-11-27
Remove busted Group/Ring/Field for Future/Try, beef up tests

### LINKED ISSUE #548: Remove invalid Future/Try Group/Ring
the exceptional cases have no inverse, they should be removed (we are clearly using bad test generators or not running tests, we should diagnose that as well).

sritchie:
This PR:

- Closes #548 by removing the group/ring instances
- Adds actual TESTS for the monoid instances for `Future` and `Try`
- updates the generators for `Future` and `Try` to include exceptional cases

This fails binary compatibility, so we should merge the next time we plan on releasing a binary-incompatible version.


algebird-util/src/test/scala/com/twitter/algebird/util/UtilAlgebraSpec.scala
```diff
@@ -0,0 +1,61 @@
...
+  property("Future is a monad") { monadLaws[Future, Int, String, Long](futureEq) }
+
+  property("Future[Int] is a monoid") { monoidLawsEq[Future[Int]](futureEq) }
+
+  property("Try is a monad") { monadLaws[Try, Int, String, Long]() }
+
+  property("Try[Int] is a monoid") { monoidLaws[Try[Int]] }
+}
+
+trait UtilGenerators {
```
sritchie: It's in the tests only, so I think it's fine - and we need `trait` to mix it into the tests above.

algebird-util/src/test/scala/com/twitter/algebird/util/UtilAlgebraProperties.scala
```diff
@@ -1,54 +0,0 @@
...
- */
-
-package com.twitter.algebird.util
-
-import com.twitter.algebird.CheckProperties
-import com.twitter.algebird.MonadLaws.monadLaws
-import com.twitter.util.{ Await, Future, Return, Try }
-import org.scalacheck.Arbitrary
-
-class UtilAlgebraProperties extends CheckProperties {
```
sritchie: I did this before our discussion on the other PR. will change it back!

sritchie:
Changed this so it now comes after #584.

sritchie:
Merged into #584, since it's required to fix tests there.

PR #585, opened 2016-11-29
Add tests for generated tuple/product aggregators

sritchie:
This PR increases our coverage for the worst offenders in `algebird-core` - the generated product and tuple aggregators.

I've also modified the generation script so that each data structure extends its parent. This reduces duplication and allows us to test ALL methods with `ringLaws` alone.

The inheritance doesn't break binary compatibility.


scripts/ntuple_generators.rb
```diff
@@ -41,21 +41,35 @@ def get_comment(n, algebraic_structure)
 # algebraic_structure is "monoid", "group", "ring", etc.
 #
 # Example return:
-#   "class Tuple2Monoid[T,U](implicit tmonoid : Monoid[T], umonoid : Monoid[U]) extends Monoid[(T,U)]"
+#   "class Tuple2Monoid[T,U](implicit tmonoid: Monoid[T], umonoid: Monoid[U]) extends Monoid[(T,U)]"
 def get_class_definition(n, algebraic_structure)
   # Example: "T,U"
   type_values_commaed = TYPE_SYMBOLS.first(n).join(", ")
-  "class Tuple#{n}#{algebraic_structure.capitalize}[#{type_values_commaed}](implicit #{get_type_parameters(n, algebraic_structure)}) extends #{algebraic_structure.capitalize}[(#{type_values_commaed})]"
+  extends = ["#{algebraic_structure.capitalize}[(#{type_values_commaed})]"]
```
sritchie: everything except for this change is a formatting thing that removes the extra space before `:`. Strange that the checked-in script didn't generate the checked-in code, by the way.

johnynek: What I think it is: code formatting has run on the code.

PR #588, opened 2016-11-29
sumOption impl for `GeneratedProductAlgebra`

sritchie:
Here's an example of the sumOption implementation:

```scala
override def sumOption(to: TraversableOnce[X]) = {
    val buf = new ArrayBufferedOperation[X, X](1000) with BufferedReduce[X] {
      def operate(items: Seq[X]) = {
        val iter = items.iterator.map(unapply(_).get).toIterable
        apply(asemigroup.sumOption(iter.map(_._1)).get, bsemigroup.sumOption(iter.map(_._2)).get)
      }
    }
    to.foreach(buf.put(_))
    buf.flush
  }
```

The only difference with `GeneratedAbstractAlgebra.scala`'s approach is that we share `iter` so that we don't have to call unapply more than once.

Would love feedback on that line!


scripts/product_generators.rb
```diff
@@ -126,8 +126,26 @@ def get_operation(n, algebraic_structure, operation)
   "override def #{operation}(#{method_params}) = { val lTuple = unapply(l).get; val rTuple = unapply(r).get; #{values_commaed} }"
 end
 
+def get_sumoption(n, bufferSize)
+  # Example: "tsemigroup.sumOption(items.iterator.map(_._1), "
+  values_commaed = TYPE_SYMBOLS.first(n).each_with_index.map do |t, i|
+    "#{t.downcase}semigroup.sumOption(iter.map(_._#{i+1})).get"
```
johnynek: do we want `iter.iterator.map(...` so we maybe avoid a copy of the `Iterable`?

johnynek: I think we do want the outer copy of the iterable, but not the second one that gets made to do this map. For this one I do think we want to pass an iterator to sumOption.

sritchie:
@johnynek fixed! Tests pass locally. Let's see what travis says.

johnynek:
👍

PR #589, opened 2016-11-30
Coverage and documentation for AveragedValue

sritchie:
This PR increases test coverage and adds a documentation page for `AveragedValue`. It also adds a `sumOption` implementation. (This comes after #591, to take advantage of the benchmarking goodies.)

Run the benchmark with:

```
jmh:run -t1 -f 1 -wi 15 -i 20 .*AveragedValueBenchmark.*
```

Here are the results:
```
[info] Benchmark                             (numElements)   Mode  Cnt      Score      Error  Units
[info] AveragedValueBenchmark.timePlus               10000  thrpt   20   6474.320 ±   65.222  ops/s
[info] AveragedValueBenchmark.timeSumOption          10000  thrpt   15  15057.069 ±  305.026  ops/s
[success] Total time: 73 s, completed Dec 1, 2016 2:54:13 PM
```

So, about a 2.3x speedup from the optimized `sumOption` impl.

### DIFF
--- a/algebird-core/src/main/scala/com/twitter/algebird/AveragedValue.scala
+++ b/algebird-core/src/main/scala/com/twitter/algebird/AveragedValue.scala
@@ -17,42 +17,97 @@ limitations under the License.
 package com.twitter.algebird
 
 /**
- * AveragedValue tracks the count and mean, or average value, of a
- * data stream.
+ * Tracks the count and mean value of Doubles in a data stream.
  *
- * Adding two instances of AveragedValue with `+` is equivalent to
- * taking an average of the two streams, with each stream weighted by
- * its count.
+ * Adding two instances of [[AveragedValue]] with [[+]]
+ * is equivalent to taking an average of the two streams, with each
+ * stream weighted by its count.
  *
- * The mean calculation uses an online algorithm suitable for large
- * numbers of records, similar to:
- * http://en.wikipedia.org/wiki/Algorithms_for_calculating_variance#Parallel_algorithm
+ * The mean calculation uses a numerically stable online algorithm
+ * suitable for large numbers of records, similar to Chan et. al.'s
+ * [[http://en.wikipedia.org/wiki/Algorithms_for_calculating_variance#Parallel_algorithm
+ * parallel variance algorithm on Wikipedia]]. As long as your count
+ * doesn't overflow a Long, the mean calculation won't overflow.
  *
- * See [[MomentsGroup.getCombinedMean]] for the code.
- *
- * @param count the number of items aggregated into this instance
- * @param value the average value of all items aggregated into this instance
+ * @see [[MomentsGroup.getCombinedMean]] for implementation of [[+]]
+ * @param count the number of aggregated items
+ * @param value the average value of all aggregated items
  */
 case class AveragedValue(count: Long, value: Double) {
+  /**
+   * Returns a copy of this instance with a negative value. Note that
+   *
+   * {{{
+   * a + -b == a - b
+   * }}}
+   */
   def unary_- : AveragedValue = copy(count = -count)
+
+  /**
+   * Averages this instance with the *opposite* of the supplied
+   * [[AveragedValue]] instance, effectively subtracting out that
+   * instance's contribution to the mean.
+   *
+   * @param r the instance to subtract
+   * @return an instance with `r`'s stream subtracted out
+   */
   def -(r: AveragedValue): AveragedValue = AveragedGroup.minus(this, r)
+
+  /**
+   * Averages this instance with another [[AveragedValue]] instance.
+   * @param r the other instance
+   * @return an instance representing the mean of this instance and `r`.
+   */
   def +(r: AveragedValue): AveragedValue = AveragedGroup.plus(this, r)
 }
 
+/**
+ * Provides a set of operations needed to create and use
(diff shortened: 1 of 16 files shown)


algebird-core/src/main/scala/com/twitter/algebird/AveragedValue.scala
```diff
@@ -17,17 +17,23 @@ limitations under the License.
 package com.twitter.algebird
 
 object AveragedValue {
+  // TODO: Change to Group[AveragedValue] in 0.13.0
```
johnynek: Can we add an issue and a link to the issue. In my experience TODOs in code are generally ignored.

algebird-core/src/main/scala/com/twitter/algebird/AveragedValue.scala
```diff
@@ -44,7 +50,16 @@ object AveragedGroup extends Group[AveragedValue] {
...
   override def isNonZero(av: AveragedValue) = (av.count != 0L)
 
-  override def negate(av: AveragedValue) = AveragedValue(-av.count, av.value)
+  override def negate(av: AveragedValue) = -av
+
+  override def sumOption(iter: TraversableOnce[AveragedValue]): Option[AveragedValue] =
+    if (iter.isEmpty) None
+    else {
+      val (c, v) = iter.foldLeft((0L, 0.0)) {
+        case ((count, sum), AveragedValue(c, v)) => (count + c, sum + (c * v))
```
johnynek: This breaks the stability of the algorithm doesn't it? We go to lengths to not just do the simple thing in plus and here we go back to the simple thing. I don't know if we can speed up sumOption without duplicating code.

docs/src/main/tut/datatypes/averaged_value.md
```diff
@@ -8,8 +8,16 @@ scaladoc: "#com.twitter.algebird.AveragedValue"
 
 # Averaged Value
 
-### Documentation Help
+## Algebraic Properties
```
johnynek: Can we mention it is a Group meaning you can subtract values?

docs/src/main/tut/datatypes/averaged_value.md
```diff
@@ -8,8 +8,16 @@ scaladoc: "#com.twitter.algebird.AveragedValue"
...
+
+## Usage Examples
```
johnynek: Seems like just showing some example of getting the average of a bunch of Longs or Doubles would be fine.

johnynek: Maybe also showing the Aggregator.

docs/src/main/tut/datatypes/averaged_value.md
```diff
@@ -8,8 +8,16 @@ scaladoc: "#com.twitter.algebird.AveragedValue"
 
 # Averaged Value
```
johnynek: We should mention the two main facts: 1) allows you to get the average and the count with one pass over the data. 2) tracks the average using a stable algorithm that won't overflow for big data (as long as there are fewer than 2^63 items generally).

algebird-core/src/main/scala/com/twitter/algebird/AveragedValue.scala
sritchie: yup, I added it to the 0.13.0 cleanup - I'll add the link here.

algebird-core/src/main/scala/com/twitter/algebird/AveragedValue.scala
sritchie: yeah, that's probably right. Probably worth removing...I guess we have to allocate a `Tuple2` anyway even if we abstract out the code. We might as well just create the `AveragedValue` directly.

so kill this?

algebird-core/src/main/scala/com/twitter/algebird/AveragedValue.scala
```diff
@@ -16,60 +16,142 @@ limitations under the License.
...
+  /**
+   * Averages this instance with the *opposite* of the supplied
+   * [[AveragedValue]] instance, effectively subtracting out that
+   * instance's contribution to the mean.
+   *
+   * @param r the instance to subtract
+   * @return an instance with `r`'s stream subtracted out
+   */
+  def -(r: AveragedValue): AveragedValue = AveragedGroup.minus(this, r)
+
```
johnynek: while we are at it, should we add:
```scala
def +(that: Double): AveragedValue = AveragedValue(count + 1L, ....)`
def +[N](that: N)(implicit num: Numeric[N]): AveragedValue = this + (num.toDouble(that))
```

johnynek:
👍 So great!

Epic PR. benchmarks, docs, perf, nice methods!

algebird-core/src/main/scala/com/twitter/algebird/First.scala
```diff
@@ -16,32 +16,63 @@ limitations under the License.
...
+ *
+ * @param get wrapped instance of `T`
  */
 case class First[@specialized(Int, Long, Float, Double) +T](get: T) {
+  /**
+   * Returns this instance, always.
+   *
+   * @param r ignored instance of `First[U]`
+   */
   def +[U >: T](r: First[U]): First[U] = this
```
johnynek: you could actually make this return `First[T]`, I think.

PR #591, opened 2016-12-01
optimize Generated{Abstract,Product}Algebra with benchmarking

sritchie:
This PR

- Adds a benchmark for `Tuple4` (seemed like a nice one to pick :)
- corrects some bad performance in the generated `sumOption` implementations that this benchmark uncovered.

Run the benchmark with:

```
sbt algebird-benchmark/jmh:run -t1 -f 1 -wi 5 -i 15 .*Tuple4Benchmark.*
```

### Benchmark Results

Before the changes:
```
[info] Benchmark                             (numElements)   Mode  Cnt     Score     Error  Units
[info] Tuple4Benchmark.timeProductPlus               10000  thrpt   15  1870.581 ± 149.159  ops/s
[info] Tuple4Benchmark.timeProductSumOption          10000  thrpt   15   492.128 ±  29.016  ops/s
[info] Tuple4Benchmark.timeTuplePlus                 10000  thrpt   15  3911.323 ± 106.611  ops/s
[info] Tuple4Benchmark.timeTupleSumOption            10000  thrpt   15  2682.713 ±  39.312  ops/s
```

After the changes:
```
[info] Benchmark                             (numElements)   Mode  Cnt     Score     Error  Units
[info] Tuple4Benchmark.timeProductPlus               10000  thrpt   15  1934.965 ±  50.165  ops/s
[info] Tuple4Benchmark.timeProductSumOption          10000  thrpt   15  2896.808 ±  71.128  ops/s
[info] Tuple4Benchmark.timeTuplePlus                 10000  thrpt   15  3893.420 ± 139.511  ops/s
[info] Tuple4Benchmark.timeTupleSumOption            10000  thrpt   15  4073.248 ±  60.992  ops/s
[success] Total time: 88 s, completed Dec 1, 2016 12:48:43 PM
```

so ~6x improvement for the ProductN semigroups, 1.5x improvement for the tupleN semigroups. AND, more importantly, they're all more efficient than just using `plus`.

Note that benchmarks only adds longs, so it measures the improvement of `sumOption` on tuples when the tuple elements don't have an efficient `sumOption` implementation. When they do, these `sumOption` implementations will be that much faster.


build.sbt
```diff
@@ -166,6 +166,7 @@ lazy val algebird = Project(
   base = file("."),
   settings = sharedSettings)
   .settings(noPublishSettings)
+  .settings(coverageExcludedPackages := "<empty>;.*\\.benchmark\\..*")
```
sritchie: gotta increase the coverage!

build.sbt
```diff
@@ -216,6 +216,7 @@ lazy val algebirdTest = module("test").settings(
 ).dependsOn(algebirdCore)
 
 lazy val algebirdBenchmark = module("benchmark").settings(JmhPlugin.projectSettings:_*).settings(
+   coverageExcludedPackages := "com\\.twitter\\.algebird\\.benchmark.*",
```
sritchie-stripe: this is hurting our coverage!

algebird-core/src/main/scala/com/twitter/algebird/BufferedOperation.scala
```diff
@@ -53,6 +53,19 @@ abstract class ArrayBufferedOperation[I, O](size: Int) extends Buffered[I, O] {
   def isFlushed = buffer.isEmpty
 }
 
+object ArrayBufferedOperation {
+  /**
+   * Returns an ArrayBufferedOperation instance that internally uses
+   * the `sumOption` implementation of the supplied Semigroup[T]
+   */
+  def fromSumOption[T](size: Int)(implicit sg: Semigroup[T]) =
+    new ArrayBufferedOperation[T, T](1000) with BufferedReduce[T] {
```
oscar-stripe: aren't you ignoring `size`?

algebird-core/src/main/scala/com/twitter/algebird/BufferedOperation.scala
```diff
@@ -53,6 +53,19 @@ abstract class ArrayBufferedOperation[I, O](size: Int) extends Buffered[I, O] {
   def isFlushed = buffer.isEmpty
 }
 
+object ArrayBufferedOperation {
+  /**
+   * Returns an ArrayBufferedOperation instance that internally uses
+   * the `sumOption` implementation of the supplied Semigroup[T]
+   */
+  def fromSumOption[T](size: Int)(implicit sg: Semigroup[T]) =
```
oscar-stripe: can we put a narrow return type on this? Maybe `BufferedReduce[T]`?

sritchie:
@johnynek ah, nice catches. Was letting the coffee get to me. All fixed up.

isnotinvain:
LGTM

Any idea why ignoring `size` didn't trigger a test failure? Is that just a size hint for performance?

Does this now always use sumOption, even when plus would normally be called? I wonder if any `plus(x,y)` implementations are more efficient than `sumOption(Seq(x,y))` in the case of only adding 2 items together? (eg, map monoid's sumOption creating a whole mutable hash map in sumOption)

sritchie-stripe:
@isnotinvain, my mistake was that I left a value of `1000` hardcoded, instead of respecting the `size` parameter. The Tuple and Record `sumOption` implementations also pass a value of `1000`, so there was no difference.

All this PR does is add a more efficient implementation for anything that explicitly calls `sumOption` - the `plus` logic still works exactly like it did before. In fact, the benchmark calls out the difference between using `plus` explicitly vs `sumOption` for larger collections.

If you call `sumOption` on a collection of two things, it's true that it might be more efficient to just add them together with `plus` - `sumOption` operates on `TraversableOnce`, though, so it can't know that information. It's up to the caller to know the difference.

Thanks for the LGTM! Tests timed out, so waiting for the build to complete again and then I'll merge and continue with the averaged value implementation.

isnotinvain (approved):
thanks for explaining!

johnynek:
so, I'm worried about this.

On a deep tuple are we exponentially growing buffers? Previously we were not (I don't think). You had a linear number of buffers. Can we stop and document why this is not exponential here?

isnotinvain:
Sorry, looks like I pulled the trigger too soon. Did this go into the release? 
Do we need to back it out :( ?

sritchie-stripe:
hey, sorry @isnotinvain for not updating the ticket! @johnynek and I talked offline when releasing and decided that this is NOT an issue. Buffers are going to grow linearly with the depth if you have nested tuples, which is fine.

If I have a tuple2 of tuple2s, The other tuple 2 will create 2 buffers, one for each inner tuple - each inner tuple will then create 2 buffers as well. I think this is totally fine, and how it has to work. Does that sound right?

Glad we thought this through!

isnotinvain:
Yeah SGTM!

PR #594, opened 2016-12-03
replace FromIntLike with Ring and toK function

### LINKED ISSUE #554: remove `FromIntLike` typeclass.
this is in algebird-test, but really `Ring[T]` is enough and we never us it in a case where we don't have a `Ring`.

### LINKED ISSUE #569: scalacheck-identified TopPctCMS failure for Short, Int, success for Long
This transient error showed up in the Travis test run for #556. Here's the failure:

```
import com.twitter.algebird._

val (depth, width, seed) = (3, 24, -5063)
val data: Seq[Short] = Seq(1, 2, 2, 3, 3, 3, 4, 4, 4, 4, 5, 5, 5, 5, 5)
val cms = TopPctCMS.aggregator[Short](depth, width, seed, 0.1).apply(data)

assert(cms.heavyHitters == Set(2, 3, 4, 5)) // fails!
```

This ALSO fails when `data: Seq[Int]`. It PASSES when `data: Seq[Long]`.

sritchie:
This PR removes the `FromIntLike` typeclass and uses either `Ring.fromInt` where possible, or a custom `toK` function for `String` and `Bytes`. Closes #554.

This also closes  #569 by fixing the seed of the CMS tests.


johnynek:
👍

sritchie:
Build timed out. I think all these PRs are killing travis.

Issue #508, opened 2016-02-16
Property test failure for `DecayedVector` associativity law


Gabriel439:
The travis run for https://github.com/twitter/algebird/pull/504 failed due to this unrelated property test failure.  I wanted to document this test failure before re-running the build:

```
DecayedVectorProperties:
- DecayedVector[Map[Int, _]] is a monoid *** FAILED ***
  GeneratorDrivenPropertyCheckFailedException was thrown during property evaluation.
   (CheckProperties.scala:12)
    Falsified after 97 successful property evaluations.
    Location: (CheckProperties.scala:12)
    Occurred when passed generated values (
      arg0 = DecayedVector(Map(0 -> 1.092551019411429E-131, -2147483648 -> -5.1395923349246496E-186, 1123520091 -> -1.054823144609479E81, 1729172390 -> 2.7044208603297777E-204, -1025632385 -> -2.6407550299049166E83, -1531213747 -> 9.369914461173615E-108, 2147483647 -> 6.626359682722528E214, -1 -> 9.183701042114227E-121, 1733000002 -> 1.9054796907962035E-69, 1486261998 -> -3.596548607681309E298),163.61309015320558),
[...]

PR #595, opened 2016-12-03
Convert all laws to take Equiv instances, deprecate Equiv versions

sritchie:
This PR converts all laws (and all tests!) to take Equiv instances by default. This is a binary-incompatible change, but in functions that take lots of implicit arguments that shouldn't be passed by default anyway.

All `LawsEquiv` versions have a deprecation warning and point to the actual `Laws`.

This also adds `.jvmopts` to, hopefully, speed up the build.


sritchie:
Holy shit... those jvm opts dropped the build time from 40something minutes to 7 minutes on the `2.11` build.

sritchie:
@ianoc looks like these started running out of permgen space. Probably need to redo them one at a time.

ianoc:
@sritchie Yeah looks like that. Lets not worry about it in this PR. The last one failed though. is that test reasonably flaky?

ianoc:
Thanks for running the tests you did btw!

sritchie:
@ianoc ah, yes, that one is, and we have an issue for it, I believe. kicking it off again

sritchie:
here we go: https://github.com/twitter/algebird/issues/508

sritchie:
💥  tests pass @ianoc @johnynek.

johnynek:
💯

PR #597, opened 2016-12-03
handle empty in Generated{Product,Abstract}Algebra

sritchie:
This checks for an empty iterator first and breaks, rather than create N `ArrayBufferedOperation` instances just to bail out if the sequence has nothing to offer.


johnynek:
:+1:

Note you should probably do this same thing to the case class macros. Shame we have to fix it N times.

sritchie:
@johnynek  Here we go: https://github.com/twitter/algebird/pull/599

PR #584, opened 2016-11-29
Create Equiv versions of all laws, fix broken tests

sritchie:
This PR comes after #579. Closes #528. 

### Description
- Addresses #548 by deprecating the group/ring instances for future/try
- Adds actual tests for the monoid and semigroup instances for `Future` and `Try`
- updates the generators for `Future` and `Try` to include exceptional cases
- Adds an `Equiv` version to all remaining laws
- activates tests for `Batched` and `TopCMS`, which weren't running
- adds implicit `Equiv` instances for SpaceSaver and SketchMap
- tightens some tests to commutative
- moves `MetricProperties` into `src` so users can make use of `metricsLaws[T]`

PR #601, opened 2016-12-03
Attempt to speed up build with JVM opts

sritchie:
The doc build is so slow... cribbing this from Algebra. Maybe it'll help.


sritchie:
Closing and adding this to #584, since that fixes broken tests.

Issue #577, opened 2016-11-21
Widen the visibility of SpaceSaver.SSMany, SpaceSaver.SSOne


ElPicador:
I need to serialize efficiently an instance of `SpaceSaver`. 
For that, I need to access the inner fields `capacity` & `counters`, and the constructors of `SSMany` and `SSOne`.

The issue is both the classes & the constructors are private.

Would it be possible to have them as `public`?

I'm willing to do the PR for this.

johnynek:
Can we instead have a direct serialization method like `HyperLogLog` does?

Exposing inner classes like that makes binary and source compatibility harder.

ElPicador:
It could be an option. 
But the issue I see is that the `counter` in `SpaceSaver` is parametrize with `T`. So how can we encode those objects? Using the default Java serialization is not a good option IMHO.

johnynek:
you can pass something like an injection: `T => Array[Byte], Array[Byte] => Try[T]` and we call as needed.

sritchie:
Thanks for handling this!

PR #603, opened 2016-12-06
SpaceSaver fromBytes & toBytes (#577)

ElPicador:
My implementation is very basic, but it should work

### DIFF
--- a/algebird-core/src/main/scala/com/twitter/algebird/SpaceSaver.scala
+++ b/algebird-core/src/main/scala/com/twitter/algebird/SpaceSaver.scala
@@ -1,6 +1,9 @@
 package com.twitter.algebird
 
+import java.nio.ByteBuffer
+
 import scala.collection.immutable.SortedMap
+import scala.util.{ Failure, Success, Try }
 
 object SpaceSaver {
   /**
@@ -18,6 +21,99 @@ object SpaceSaver {
   private[algebird] val ordering = Ordering.by[(_, (Long, Long)), (Long, Long)]{ case (item, (count, err)) => (-count, err) }
 
   implicit def spaceSaverSemiGroup[T]: Semigroup[SpaceSaver[T]] = new SpaceSaverSemigroup[T]
+
+  /**
+   * Encodes the SpaceSaver as a sequence of bytes containing in order
+   * - 1 byte: 1/2 => 1 = SSOne, 2 = SSMany
+   * - 4 bytes: the capacity
+   * - N bytes: the item/counters (counters as length + N*(item size + item + 2 * counters)
+   */
+  def toBytes[T](ss: SpaceSaver[T], tSerializer: T => Array[Byte]): Array[Byte] = {
+    ss match {
+      case SSOne(capacity, item) =>
+        val itemAsBytes = tSerializer(item)
+        val itemLength = itemAsBytes.length
+        //1 for the type, 4 for capacity, 4 for itemAsBytes.length
+        val buffer = new Array[Byte](1 + 4 + 4 + itemAsBytes.length)
+        ByteBuffer
+          .wrap(buffer)
+          .put(1: Byte)
+          .putInt(capacity)
+          .putInt(itemLength.toByte)
+          .put(itemAsBytes)
+        buffer
+
+      case SSMany(capacity, counters, _) => //We do not care about the buckets are thery are created by SSMany.apply
+        val buffer = scala.collection.mutable.ArrayBuffer.newBuilder[Byte]
+        buffer += (2: Byte)
+
+        var buff = ByteBuffer.allocate(4)
+        buff.putInt(capacity)
+        buffer ++= buff.array()
+
+        buff = ByteBuffer.allocate(4)
+        buff.putInt(counters.size)
+        buffer ++= buff.array()
+        counters.foreach {
+          case (item, (a, b)) =>
+            val itemAsBytes = tSerializer(item)
+
+            buff = ByteBuffer.allocate(4)
+            buff.putInt(itemAsBytes.length)
+            buffer ++= buff.array()
+
+            buffer ++= itemAsBytes
+
+            buff = ByteBuffer.allocate(8 * 2)
+            buff.putLong(a)
+            buff.putLong(b)
+            buffer ++= buff.array()
+        }
+        buffer.result.toArray
+    }
+  }
+
+  // Make sure to be reversible so fromBytes(toBytes(x)) == x
+  def fromBytes[T](bytes: Array[Byte], tDeserializer: Array[Byte] => Try[T]): Try[SpaceSaver[T]] = {
+    fromByteBuffer(ByteBuffer.wrap(bytes), buffer => tDeserializer(buffer.array()))
+  }
+
+  def fromByteBuffer[T](bb: ByteBuffer, tDeserializer: ByteBuffer => Try[T]): Try[SpaceSaver[T]] = {
+    Try {
+      bb.get.toInt match {
+        case 1 =>
+          val capacity = bb.getInt
+          val itemLength = bb.getInt
+          val itemAsBytes = new Array[Byte](itemLength)
+          bb.get(itemAsBytes)
+          tDeserializer(ByteBuffer.wrap(itemAsBytes)).map(item => SSOne(capacity, item))
+        case 2 =>
+          val capacity = bb.getInt
+
+          var countersToDeserialize = bb.getInt
+          val counters = scala.collection.mutable.Map.empty[T, (Long, Long)]
+          while (countersToDeserialize != 0) {
+            val itemLength = bb.getInt()
+            val itemAsBytes = new Array[Byte](itemLength)
+            bb.get(itemAsBytes)
+            val item = tDeserializer(ByteBuffer.wrap(itemAsBytes))
+
+            val a = bb.getLong
+            val b = bb.getLong
+
+            item match {
+              case Failure(e) => return Failure(e)
+              case Success(i) =>
+                counters += ((i, (a, b)))
+            }
+
+            countersToDeserialize -= 1
+          }
+
+          Success(SSMany(capacity, counters.toMap))
+      }
+    }.flatten
+  }
 }
 
 /**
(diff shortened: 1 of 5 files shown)


algebird-test/src/test/scala/com/twitter/algebird/SpaceSaverTest.scala
```diff
@@ -59,5 +61,26 @@ class SpaceSaverTest extends WordSpec with Matchers {
     "be consistent in its constructors" in {
       assert(SpaceSaver(10, "ha") ++ SpaceSaver(10, "ha") ++ SpaceSaver(10, "ha") === SpaceSaver(10, "ha", 3))
     }
+
+    "Correctly serialize" in {
+      //SSOne
+      verifySerialization(SpaceSaver(40, "string that is not empty"))
```
sritchie-stripe: would you mind converting this to a scalacheck test? Something like

```scala
property("SpaceSaver serializes properly") { ss: SpaceSaver[String] =>
  val spaceSaverAsArrayByte = SpaceSaver.toBytes(ss, stringToArrayByte)
  SpaceSaver.fromBytes(spaceSaverAsArrayByte, arrayByteToString).get == ss
}
```

The best way to make this work is to get a scalacheck `Arbitrary` instance into this file:

https://github.com/twitter/algebird/blob/develop/algebird-test/src/main/scala/com/twitter/algebird/scalacheck/Arbitrary.scala

and a `Gen` instance into this file:

https://github.com/twitter/algebird/blob/develop/algebird-test/src/main/scala/com/twitter/algebird/scalacheck/Gen.scala

I know it's a bigger ask, but serialization is a big deal and I think we absolutely need the scalacheck laws in place here.

algebird-core/src/main/scala/com/twitter/algebird/SpaceSaver.scala
```diff
@@ -18,6 +21,98 @@ object SpaceSaver {
...
 
   implicit def spaceSaverSemiGroup[T]: Semigroup[SpaceSaver[T]] = new SpaceSaverSemigroup[T]
+
+  /**
+   * Encodes the SpaceSaver as a sequence of bytes containing in order
+   * - 1 byte: 1/2 => 1 = SSOne, 2 = SSMany
+   * - 4 bytes: the capacity
+   * - N bytes: the item/counters (counters as length + N*(item size + item + 2 * counters)
+   */
+  def toBytes[T](ss: SpaceSaver[T], tSerializer: T => Array[Byte]): Array[Byte] = {
```
sritchie-stripe, line 31: Another ask - could we serialize a couple of examples and put a base64-encoded string of the bytes into the test, so that we can guarantee that the serialization doesn't change? I can see us making a change here that wouldn't break the tests, but would break deserialization of old instances.

sritchie-stripe (changes requested):
Thanks for doing this, and love the coverage increase! A few testing asks, since serialization is a big deal. Really happy that we're getting this done.

sritchie-stripe:
@ElPicador @non @johnynek have any of you used scodec? https://github.com/scodec/scodec Seems like it might be a nice library to use for these serialization as we start writing them.

sritchie-stripe:
Oh, @ElPicador, one more ask - would you please add an example of the new serialization methods to the spacesaver documentation page?

https://github.com/twitter/algebird/blob/develop/docs/src/main/tut/datatypes/approx/space_saver.md

Just something about like this 

```
    ## Serialization
   You can serialize `SpaceSaver` instances using `fromBytes` and `toBytes`:

    ```tut:book
    import com.twitter.algebird._
    val ss = ???
    // keep the example going!
    ```
```
With the new documentation, you can just add lines as you'd type them into the repl. The output will look great:

https://twitter.github.io/algebird/datatypes/averaged_value.html

Thanks again!

algebird-test/src/test/scala/com/twitter/algebird/SpaceSaverTest.scala
ElPicador: Will do. I wanted to do it, but before I wanted to be sure the PR could be accepted 😄

algebird-core/src/main/scala/com/twitter/algebird/SpaceSaver.scala
ElPicador: Good idea, will do

ElPicador:
It should be good

algebird-core/src/main/scala/com/twitter/algebird/SpaceSaver.scala
```diff
@@ -18,6 +21,98 @@ object SpaceSaver {
...
+        buffer.result.toArray
+    }
+  }
+
+  // Make sure to be reversible so fromBytes(toBytes(x)) == x
+  def fromBytes[T](bytes: Array[Byte], tDeserializer: Array[Byte] => Try[T]): Try[SpaceSaver[T]] = {
+    fromByteBuffer(ByteBuffer.wrap(bytes), tDeserializer)
+  }
+
+  def fromByteBuffer[T](bb: ByteBuffer, tDeserializer: Array[Byte] => Try[T]): Try[SpaceSaver[T]] = {
```
johnynek: I think `ByteBuffer => Try[T]` might allow some better performance at a moment.

ElPicador: Changed

algebird-core/src/main/scala/com/twitter/algebird/SpaceSaver.scala
```diff
@@ -18,6 +21,98 @@ object SpaceSaver {
...
+    }
+  }
+
+  // Make sure to be reversible so fromBytes(toBytes(x)) == x
+  def fromBytes[T](bytes: Array[Byte], tDeserializer: Array[Byte] => Try[T]): Try[SpaceSaver[T]] = {
+    fromByteBuffer(ByteBuffer.wrap(bytes), buffer => tDeserializer(buffer.array()))
+  }
+
+  def fromByteBuffer[T](bb: ByteBuffer, tDeserializer: ByteBuffer => Try[T]): Try[SpaceSaver[T]] = {
+    bb.get.toInt match {
```
johnynek: Couldn't other operations on `bb` throw? I think we need an outer Try to catch those exceptions too (or use a try/catch block inside the method.

algebird-core/src/main/scala/com/twitter/algebird/SpaceSaver.scala
```diff
@@ -18,6 +21,98 @@ object SpaceSaver {
...
+            case Failure(e) =>
+              return Failure(e)
+            case Success(i) =>
+              counters += ((i, (a, b)))
+          }
+
+          countersToDeserialize -= 1
+        }
+
+        Try(SSMany(capacity, counters.toMap))
```
johnynek: Why do you need Try here rather than just Success? Could this throw?

algebird-core/src/main/scala/com/twitter/algebird/SpaceSaver.scala
```diff
@@ -18,6 +21,98 @@ object SpaceSaver {
...
+          val itemLength = bb.getInt()
+          val itemAsBytes = new Array[Byte](itemLength)
+          bb.get(itemAsBytes)
+          val item = tDeserializer(ByteBuffer.wrap(itemAsBytes))
+
+          val a = bb.getLong
+          val b = bb.getLong
+
+          item match {
+            case Failure(e) =>
```
johnynek: `case f@Failure(_) => return f` would avoid an allocation.

algebird-test/src/test/scala/com/twitter/algebird/SpaceSaverTest.scala
```diff
@@ -26,22 +31,31 @@ class SpaceSaverLaws extends CheckProperties {
...
             (left consistentWith right) && (right consistentWith left)
           }
+
         commutativeSemigroupLaws[SpaceSaver[Int]]
       }
     }
   }
+
+  property("SpaceSaver can serialize/deserialize itself") {
+    forAll { ss: SpaceSaver[String] =>
```
johnynek, line 41: Can we also add a test that exercises Failure? Like pass an empty array, and see that you get Failure not an exception thrown.

So basically: `forAll { a: Array[Byte] => ` deserialization does not throw? You get Success or Failure.

johnynek (changes requested):

algebird-test/src/test/scala/com/twitter/algebird/SpaceSaverTest.scala
ElPicador: Done

algebird-core/src/main/scala/com/twitter/algebird/SpaceSaver.scala
ElPicador: True, I changed it

algebird-core/src/main/scala/com/twitter/algebird/SpaceSaver.scala
ElPicador: Won't compile. One is a `Try[T]` the other is a `Try[SpaceSaver[T]]`

algebird-core/src/main/scala/com/twitter/algebird/SpaceSaver.scala
ElPicador: You're right. I've changed it

algebird-test/src/test/scala/com/twitter/algebird/SpaceSaverTest.scala
sritchie: Pending this one change this looks great to me! @ElPicador, thanks for responding to these rounds of feedback. With this extra test the new serde code will have 100% coverage :)

ElPicador: @sritchie I'm not sure to understand what change you are talking about.

sritchie-stripe: oh, sorry, I was thinking of the test that @johnynek requested.

sritchie-stripe:
looks great to me! @ElPicador, @johnynek is traveling today - let's let him take one more look, and then we'll get this merged.

The plan is to fix up the remaining two issues in #600 and then prep a release of 0.13.0. That will be binary incompatible in a couple of minor ways that should be easy to fix. Does that work for you?

Thanks so much for the contribution!

ElPicador:
Perfect for me. My step is to add a serializer in twitter-chill.

johnynek (approved):

johnynek:
Thanks!

PR #605, opened 2016-12-11
Remove typeclass from interval constructor

sritchie:
Moves the `Ordering` inputs from the constructor to the methods that need it, as discussed in #600 .


algebird-core/src/main/scala/com/twitter/algebird/Interval.scala
```diff
@@ -22,31 +22,31 @@ package com.twitter.algebird
...
  */
 sealed trait Interval[T] extends java.io.Serializable {
-  def contains(t: T): Boolean
+  def contains(t: T)(implicit ord: Ordering[T]): Boolean
 
-  def intersect(that: Interval[T]): Interval[T]
-  def apply(t: T) = contains(t)
-  def &&(that: Interval[T]) = intersect(that)
+  def intersect(that: Interval[T])(implicit ord: Ordering[T]): Interval[T]
+  def apply(t: T)(implicit ord: Ordering[T]) = contains(t)
```
johnynek: Let's make apply and `&&` final

algebird-core/src/main/scala/com/twitter/algebird/Interval.scala
```diff
@@ -22,31 +22,31 @@ package com.twitter.algebird
...
+  def &&(that: Interval[T])(implicit ord: Ordering[T]) = intersect(that)
 
   /**
    * Map the Interval with a non-decreasing function.
    * If you use a non-monotonic function (like x^2)
    * then the result is meaningless.
    * TODO: It might be good to have types for these properties in algebird.
    */
-  def mapNonDecreasing[U: Ordering](fn: T => U): Interval[U]
+  def mapNonDecreasing[U](fn: T => U): Interval[U]
```
johnynek, line 37: Think we need a test for mapNonDecreasing.

algebird-core/src/main/scala/com/twitter/algebird/Interval.scala
```diff
@@ -283,11 +290,9 @@ case class Intersection[L[t] <: Lower[t], U[t] <: Upper[t], T](lower: L[T], uppe
    * it this way, it does not mean it is empty or universe, etc... (there
    * are other cases).
    */
-  def toLeftClosedRightOpen(implicit s: Successible[T]): Option[Intersection[InclusiveLower, ExclusiveUpper, T]] = {
-    implicit val ord = lower.ordering
+  def toLeftClosedRightOpen(implicit s: Successible[T]): Option[Intersection[InclusiveLower, ExclusiveUpper, T]] =
```
johnynek, line 295: Let's try to see if we can just jump Successible and Predecessible up to Ordering. I can't think of a good reason to have Partial there.

johnynek (changes requested):

algebird-core/src/main/scala/com/twitter/algebird/Interval.scala
johnynek: So a law might be: forAll intervals and values t, if we map t and the interval, contains returns the same value.

To make the function non-decreasing we will need to do something like make a new fn and apple the first and if the result is less return the input.

algebird-core/src/main/scala/com/twitter/algebird/Interval.scala
sritchie: it works great!

algebird-core/src/main/scala/com/twitter/algebird/Interval.scala
```diff
@@ -245,14 +243,23 @@ case class Intersection[L[t] <: Lower[t], U[t] <: Upper[t], T](lower: L[T], uppe
...
+      }
+    Intersection(newLower, newUpper)
+  }
 
   /**
    * Goes from lowest to highest for all items
    * that are contained in this Intersection
    */
-  def leastToGreatest(implicit s: Successible[T]): Iterable[T] = {
+  def leastToGreatest(implicit s: Successible[T], ord: Ordering[T]): Iterable[T] = {
```
oscar-stripe: don't need `ord` now that `Successible` has an Ordering right?

algebird-core/src/main/scala/com/twitter/algebird/Interval.scala
```diff
@@ -264,7 +271,7 @@ case class Intersection[L[t] <: Lower[t], U[t] <: Upper[t], T](lower: L[T], uppe
    * Goes from highest to lowest for all items
    * that are contained in this Intersection
    */
-  def greatestToLeast(implicit p: Predecessible[T]): Iterable[T] = {
+  def greatestToLeast(implicit p: Predecessible[T], ord: Ordering[T]): Iterable[T] = {
```
oscar-stripe: we don't need `ord` here, right?

algebird-core/src/main/scala/com/twitter/algebird/Interval.scala
sritchie: done

algebird-core/src/main/scala/com/twitter/algebird/Interval.scala
sritchie: done

algebird-core/src/main/scala/com/twitter/algebird/Predecessible.scala
```diff
@@ -39,14 +39,13 @@ trait Predecessible[T] extends java.io.Serializable {
   /**
    * The law is:
    * prev(t)
-   *   .map { n => partialOrdering.lteq(n, t) && (!partialOrdering.equiv(t, n)) }
+   *   .map { n => ordering.lteq(n, t) && (!ordering.equiv(t, n)) }
```
oscar-stripe: this can be more succinct now: `prev(t).map(ordering.lt(_, t)).getOrElse(true)`

algebird-core/src/main/scala/com/twitter/algebird/Successible.scala
```diff
@@ -42,14 +42,13 @@ trait Successible[T] {
   /**
    * The law is:
    * next(t)
-   *   .map { n => partialOrdering.lteq(t, n) && (!partialOrdering.equiv(t, n)) }
+   *   .map { n => ordering.lteq(t, n) && (!ordering.equiv(t, n)) }
```
oscar-stripe: similar: `next(t).map(ordering.lt(t, _)).getOrElse(true)`

oscar-stripe (changes requested):

johnynek (approved):

Issue #606, opened 2016-12-12
Making BloomFilter polymophic


johandahlberg:
Hi!

I've been playing around with making the Algebirds Bloom Filters polymorphic, and I'm wondering if there is any interest in a PR for this?

As it stands right now I've implemented this using a implicit view from `A => Array[Byte]`, making things look something like this (translating from the usage example on the website):

```
implicit def stringToBytes(s: String): Array[Byte] = s.getBytes()

val NUM_HASHES = 6
val WIDTH = 32

val bfMonoid = new BloomFilterMonoid[String](NUM_HASHES, WIDTH)
val bf = bfMonoid.create("1", "2", "3", "4", "100")

val approxBool = bf.contains("1")
val res = approxBool.isTrue
```

I think it's kind of neat, however I completely understand if this is not something folks here are interested in since it would introduce a breaking change to the API. This could of course be worked around by making `BloomFilterMonoid` still be bound to `String`, but be backed up by the polymorphic implementation. Before looking into that, I'd be curious to know if there is any interest in making the Bloom Filters in making the Bloom Filters polymorphic in the first place

Cheers,
Johan

johnynek:
Yeah, bloom filter needs a lot of work. We should do what you describe as was done for CMS and HLL.

We should also optimize bloom filter more. It was a contribution that has not received the attention that other aspects of the code have.

We should add benchmarks and improve them.

Lastly, it would be interesting to see if we could remove the dependency on java ewah. I'm not sure it is actually faster than a smallish local and immutable implementation.

sritchie:
@johandahlberg We'd love a PR with these ideas if you're interested!

johandahlberg:
This was closed by #607.

PR #607, opened 2016-12-13
Polymorphic Bloom filters

johandahlberg:
This relates to #606. It makes `BloomFilter` polymorphic, fixes the tests to match that (while still using `String` as the type in the tests), and it adds some basic benchmarks for creating and querying the Bloom filters.

One major change in this is that the interface has been changed so that many of the functions now require a implicit view from `A => Array[Byte]`, leading to usage looking something like this:

```
implicit def stringToBytes(s: String): Array[Byte] = s.getBytes()

val NUM_HASHES = 6
val WIDTH = 32

val bfMonoid = new BloomFilterMonoid[String](NUM_HASHES, WIDTH)
val bf = bfMonoid.create("1", "2", "3", "4", "100")

val approxBool = bf.contains("1")
val res = approxBool.isTrue
```

When it comes to the benchmarks I've never used setup benchmarks using this type of framework (or any framework at all to be honest), so I'm not sure I got it right, nor that I picked sensible values for the parameters to test. However I hope it can still serve as a useful scaffold if one wants to address the performance issues mentioned in #606.

### DIFF
--- a/algebird-core/src/main/scala/com/twitter/algebird/BloomFilter.scala
+++ b/algebird-core/src/main/scala/com/twitter/algebird/BloomFilter.scala
@@ -44,10 +44,10 @@ class RichCBitSet(val cb: CBitSet) {
 
 object BloomFilter {
 
-  def apply[A](numEntries: Int, fpProb: Double)(implicit toBytes: A => Array[Byte]) = {
+  def apply[A](numEntries: Int, fpProb: Double)(implicit hash: Hash128[A]) = {
     val width = BloomFilter.optimalWidth(numEntries, fpProb)
     val numHashes = BloomFilter.optimalNumHashes(numEntries, width)
-    BloomFilterMonoid[A](numHashes, width)(toBytes)
+    BloomFilterMonoid[A](numHashes, width)(hash)
   }
 
   // Compute optimal number of hashes: k = m/n ln(2)
@@ -76,8 +76,8 @@ object BloomFilter {
  * http://en.wikipedia.org/wiki/Bloom_filter
  *
  */
-case class BloomFilterMonoid[A](numHashes: Int, width: Int)(implicit toBytes: A => Array[Byte]) extends Monoid[BF[A]] {
-  val hashes: BFHash[A] = BFHash[A](numHashes, width)(toBytes)
+case class BloomFilterMonoid[A](numHashes: Int, width: Int)(implicit hash: Hash128[A]) extends Monoid[BF[A]] {
+  val hashes: BFHash[A] = BFHash[A](numHashes, width)(hash)
 
   val zero: BF[A] = BFZero[A](hashes, width)
 
@@ -345,10 +345,10 @@ object BFInstance {
     BFInstance(hashes, BitSet.empty, width)
 }
 
-case class BFHash[A](numHashes: Int, width: Int)(implicit toBytes: A => Array[Byte]) extends Function1[A, Iterable[Int]] {
+case class BFHash[A](numHashes: Int, width: Int)(implicit hash: Hash128[A]) extends Function1[A, Iterable[Int]] {
   val size = numHashes
 
-  def apply(s: A) = nextHash(toBytes(s), numHashes)
+  def apply(s: A) = nextHash(s, numHashes)
 
   private def splitLong(x: Long) = {
     def toNonNegativeInt(x: Long) = {
@@ -361,19 +361,19 @@ case class BFHash[A](numHashes: Int, width: Int)(implicit toBytes: A => Array[By
     (upper, lower)
   }
 
-  private def nextHash(bytes: Array[Byte], hashIndex: Int, digested: Seq[Int] = Seq.empty): Stream[Int] = {
+  private def nextHash(valueToHash: A, hashIndex: Int, digested: Seq[Int] = Seq.empty): Stream[Int] = {
     if (hashIndex == 0)
       Stream.empty
     else {
       val d = if (digested.isEmpty) {
-        val (a, b) = MurmurHash128(hashIndex)(bytes)
+        val (a, b) = hash.hashWithSeed(hashIndex.toLong, valueToHash)
         val (x1, x2) = splitLong(a)
         val (x3, x4) = splitLong(b)
         Seq(x1, x2, x3, x4)
       } else
         digested
 
-      Stream.cons(d(0) % width, nextHash(bytes, hashIndex - 1, d.drop(1)))
+      Stream.cons(d(0) % width, nextHash(valueToHash, hashIndex - 1, d.drop(1)))
     }
   }
 }
@@ -386,6 +386,6 @@ case class BloomFilterAggregator[A](bfMonoid: BloomFilterMonoid[A]) extends Mono
 }
 
 object BloomFilterAggregator {
-  def apply[A](numHashes: Int, width: Int)(implicit toBytes: A => Array[Byte]): BloomFilterAggregator[A] =
+  def apply[A](numHashes: Int, width: Int)(implicit hash: Hash128[A]): BloomFilterAggregator[A] =
     BloomFilterAggregator[A](BloomFilterMonoid[A](numHashes, width))
 }
--- a/algebird-core/src/main/scala/com/twitter/algebird/BloomFilter.scala
+++ b/algebird-core/src/main/scala/com/twitter/algebird/BloomFilter.scala
@@ -44,10 +44,10 @@ class RichCBitSet(val cb: CBitSet) {
 
 object BloomFilter {
 
-  def apply(numEntries: Int, fpProb: Double) = {
+  def apply[A](numEntries: Int, fpProb: Double)(implicit toBytes: A => Array[Byte]) = {
     val width = BloomFilter.optimalWidth(numEntries, fpProb)
     val numHashes = BloomFilter.optimalNumHashes(numEntries, width)
-    BloomFilterMonoid(numHashes, width)
+    BloomFilterMonoid[A](numHashes, width)(toBytes)
   }
 
   // Compute optimal number of hashes: k = m/n ln(2)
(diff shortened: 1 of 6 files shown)


algebird-core/src/main/scala/com/twitter/algebird/BloomFilter.scala
```diff
@@ -44,10 +44,10 @@ class RichCBitSet(val cb: CBitSet) {
 
 object BloomFilter {
 
-  def apply(numEntries: Int, fpProb: Double) = {
+  def apply[A](numEntries: Int, fpProb: Double)(implicit toBytes: A => Array[Byte]) = {
```
johnynek: what I would rather do is this:

1. use `Hash128[A]` here.
2. change `Hash128[K]` to have:
```scala
trait Hash128[K] {
  def hashWithSeed(seed: Long, k: K): (Long, Long)
  def hash(k: K): (Long, Long) = hashWithSeed(Hash128.DefaultSeed, k)
}
```

what do you think of making that change?

johandahlberg: Thank you for the feedback.

I'm not sure that I follow completely. Do you mean changing the signature to: `def apply[Hash128[A]](numEntries: Int, fpProb: Double)`?

And I guess that by number 2 you mean making those changes to the definitions in `algebird/algebird-core/src/main/scala/com/twitter/algebird/Hash128.scala`?

 If I understood correctly I can look into if I can get it working.

johnynek: change apply to:

```scala
def apply[A](numEntries: Int, fpProb: Double)(implicit hash: Hash128[A]): BF[A] = ...
```

johandahlberg: :+1:

johandahlberg:
I updated this to use an implicit `Hash128` instead of `A => Array[Byte]`.

This means that the Bloom filter can now be used in this way (cut-and-past from the updated docs):

```
import com.twitter.algebird._

// We need to provide a implicit conversion from of the Bloom filter,
// in this case String, to Hash128. Conveniently Hash128 has a number
// of built in conversions so we do not need to define our own.
implicit def stringToHash128 = Hash128.stringHash

// It's possible to create a Bloom filter with a set number of
// hashes, and with a set width.
val NUM_HASHES = 6
val WIDTH = 32
val bfMonoid1 = new BloomFilterMonoid(NUM_HASHES, WIDTH)
val bf1 = bfMonoid1.create("1", "2", "3", "4", "100")
val approxBool1 = bf1.contains("1")
val res1 = approxBool1.isTrue
```

I'm unsure about why d6c37d9 and 1d861c2 failed in CI. It is not apparent to me if the errors that show up are in any way related to the code that I have edited. Especially since the later of the two was just me fixing a typo. The first one seems to be some property test failing (which they have not done on my local machine), the second one has a: `*** glibc detected *** /usr/lib/jvm/java-8-oracle/bin/java: free(): invalid pointer: 0x00000007c12a4c28 ***`. Any ideas on how to fix these?

Finally these changes have decrease test coverage by 0.01%, would you like me to see if I can add some more tests to make the test coverage go up? I don't remember seeing anything in the contributing docs about what the policy on this is.

algebird-core/src/main/scala/com/twitter/algebird/BloomFilter.scala
```diff
@@ -44,10 +44,10 @@ class RichCBitSet(val cb: CBitSet) {
 
 object BloomFilter {
 
-  def apply(numEntries: Int, fpProb: Double) = {
+  def apply[A](numEntries: Int, fpProb: Double)(implicit hash: Hash128[A]) = {
```
johnynek: can we add the return type to this method please (while we are cleaning up).

algebird-core/src/main/scala/com/twitter/algebird/Hash128.scala
```diff
@@ -21,11 +21,18 @@ package com.twitter.algebird
...
+
+  def hashWithSeed(seed: Long, k: K): (Long, Long)
+
+  def hash(k: K): (Long, Long) = hashWithSeed(DefaultSeed, k)
+
   def contramap[L](fn: L => K): Hash128[L] = {
     val self = this
     new Hash128[L] {
-      def hash(l: L) = self.hash(fn(l))
+      val DefaultSeed = Hash128.DefaultSeed
```
johnynek: this should be `val defaultSeed = self.defaultSeed`

algebird-core/src/main/scala/com/twitter/algebird/Hash128.scala
```diff
@@ -21,11 +21,18 @@ package com.twitter.algebird
  * Used for HLL, but possibly other applications
  */
 trait Hash128[-K] extends java.io.Serializable {
-  def hash(k: K): (Long, Long)
+
+  val DefaultSeed: Long
```
johnynek: can we make this `def defaultSeed: Long`? we can override a `def` with a val, but vals in traits have some strange issues (init order bug) and we just try to avoid them unless there is a compelling reason.

algebird-core/src/main/scala/com/twitter/algebird/Hash128.scala
```diff
@@ -40,14 +47,17 @@ object Hash128 extends java.io.Serializable {
 
   val DefaultSeed = 12345678L
 
-  def murmur128ArrayByte(seed: Long): Hash128[Array[Byte]] = new Hash128[Array[Byte]] {
```
johnynek, line 43: can we keep these `seed` APIs and make the passed `seed` the `defaultSeed` of the instance? I think that will keep the old behavior.

algebird-core/src/main/scala/com/twitter/algebird/Hash128.scala
```diff
@@ -56,18 +66,20 @@ object Hash128 extends java.io.Serializable {
    * but has been more commonly used in HLL.
    */
   def murmur128Utf8String(seed: Long): Hash128[String] =
-    murmur128ArrayByte(seed).contramap(_.getBytes("UTF-8"))
+    murmur128ArrayByte().contramap(_.getBytes("UTF-8"))
 
   def murmur128Int(seed: Long): Hash128[Int] = new Hash128[Int] {
```
johnynek: I think we should rename `seed` to `defSeed` to avoid a collision in names below, and it should be the `defaultSeed` for the instance created.

algebird-test/src/test/scala/com/twitter/algebird/BloomFilterTest.scala
```diff
@@ -93,8 +95,9 @@ class BFHashIndices extends CheckProperties {
 }
 
 class BloomFilterFalsePositives[T: Gen](falsePositiveRate: Double) extends ApproximateProperty {
+
   type Exact = Set[T]
-  type Approx = BF
+  type Approx = BF[String]
```
johnynek: can we make this `BF[T]` and make the type: `[T: Gen: Hash128]`?

algebird-test/src/test/scala/com/twitter/algebird/BloomFilterTest.scala
```diff
@@ -121,12 +124,13 @@ class BloomFilterFalsePositives[T: Gen](falsePositiveRate: Double) extends Appro
 
   def exactResult(s: Set[T], t: T) = s.contains(t)
 
-  def approximateResult(bf: BF, t: T) = bf.contains(t.toString)
+  def approximateResult(bf: BF[String], t: T) = bf.contains(t.toString)
 }
 
 class BloomFilterCardinality[T: Gen] extends ApproximateProperty {
```
johnynek: let's make this `T: Gen: Hash128`.

algebird-test/src/test/scala/com/twitter/algebird/BloomFilterTest.scala
```diff
@@ -121,12 +124,13 @@ class BloomFilterFalsePositives[T: Gen](falsePositiveRate: Double) extends Appro
...
+
   type Exact = Set[T]
-  type Approx = BF
+  type Approx = BF[String]
```
johnynek: let's make this `BF[T]`

docs/src/main/tut/datatypes/approx/bloom_filter.md
```diff
@@ -8,16 +8,50 @@ scaladoc: "#com.twitter.algebird.BloomFilter"
...
+The size of a Bloom filter depends on the number of input elements, and the desired false positive rate.
+
+To read more about Bloom filters see wikipedia: https://en.wikipedia.org/wiki/Bloom_filter
+
 example usage:
 
 ```tut:book
 import com.twitter.algebird._
+
+// We need to provide a implicit conversion from of the Bloom filter,
```
johnynek: I don't think you need this here. This is not an implicit conversion anymore, but an instance for the typeclass `Hash128[A]`. For String, Long, Array[Byte], Int you don't need to do anything. (As a side note: we should implement a macro to do this for case classes and tuples).

johnynek (changes requested):
This is really great! Thanks so much!

I had a few small comments.

I really want to get this into the next version of Algebird which we hope to publish very soon.

docs/src/main/tut/datatypes/approx/bloom_filter.md
johandahlberg: Aha. I see. I had missed specifying the type for `BloomFilterMonoid` here which meant that I god a compilation error about multiple type classes matching. But specifying the type solved that.

johandahlberg:
That you so much for the feedback @johnynek, I've learned a lot, so this has been super interesting. :smile:  I think I have addressed all you comments. Let me know if there is anything else.

johnynek (approved):

johnynek:
Thanks for your work on this @johandahlberg !

PR #610, opened 2017-01-05
Optimize BloomFilter significantly

johnynek:
This refactors and optimizes BloomFilter. Previously, there was no sumOption. Now there is and it is used in create. The existing create benchmarks show a 10-15x speedup:

```
create:
[info] Benchmark                                     (falsePositiveRate)  (nbrOfElements)   Mode  Cnt     Score      Error  Units
[info] BloomFilterCreateBenchmark.createBloomFilter                0.001              100  thrpt    3  1928.527 ± 5596.189  ops/s
[info] BloomFilterCreateBenchmark.createBloomFilter                0.001             1000  thrpt    3   194.313 ±  121.114  ops/s
[info] BloomFilterCreateBenchmark.createBloomFilter                0.001            10000  thrpt    3    11.972 ±   49.377  ops/s
[info] BloomFilterCreateBenchmark.createBloomFilter                 0.01              100  thrpt    3  2958.460 ± 2123.001  ops/s
[info] BloomFilterCreateBenchmark.createBloomFilter                 0.01             1000  thrpt    3   258.893 ±   75.279  ops/s
[info] BloomFilterCreateBenchmark.createBloomFilter                 0.01            10000  thrpt    3    18.209 ±   41.354  ops/s

# after changes:
[info] Benchmark                                     (falsePositiveRate)  (nbrOfElements)   Mode  Cnt      Score     Error  Units
[info] BloomFilterCreateBenchmark.createBloomFilter                0.001              100  thrpt   10  20857.377 ± 842.149  ops/s
[info] BloomFilterCreateBenchmark.createBloomFilter                0.001             1000  thrpt   10   2080.394 ±  58.374  ops/s
[info] BloomFilterCreateBenchmark.createBloomFilter                0.001            10000  thrpt   10    213.486 ±  15.750  ops/s
[info] BloomFilterCreateBenchmark.createBloomFilter                 0.01              100  thrpt   10  30260.699 ± 525.189  ops/s
[info] BloomFilterCreateBenchmark.createBloomFilter                 0.01             1000  thrpt   10   3085.647 ±  40.314  ops/s
[info] BloomFilterCreateBenchmark.createBloomFilter                 0.01            10000  thrpt   10    313.572 ±  23.155  ops/s
```

an issue discovered by this is that scala's "immutable" BitSet is just an `Array[Long]` that is copied on each write. More optimizations are possible by investigating a better immutable bit set.

In the mean time, wrapping `BF[A]` in `Batched[BF[A]]` will allow you to use the optimized `sumOption` which for many cases will be good enough.

@sritchie @non @isnotinvain review?

### DIFF
--- a/algebird-core/src/main/scala/com/twitter/algebird/BloomFilter.scala
+++ b/algebird-core/src/main/scala/com/twitter/algebird/BloomFilter.scala
@@ -20,49 +20,156 @@ import scala.collection.immutable.BitSet
 import scala.collection.JavaConverters._
 
 import com.googlecode.javaewah.{ EWAHCompressedBitmap => CBitSet }
+import com.googlecode.javaewah.IntIterator
 
 object RichCBitSet {
-  def apply(x: Int*) = {
-    CBitSet.bitmapOf(x.sorted: _*)
+  def apply(xs: Int*): CBitSet = fromArray(xs.toArray)
+
+  // this sorts the array in-place
+  def fromArray(x: Array[Int]): CBitSet = {
+    val bs = new CBitSet
+    bs += x
+  }
+  def fromBitSet(bs: BitSet): CBitSet = {
+    val nbs = new CBitSet
+    val it = bs.iterator
+    while (it.hasNext) { nbs.set(it.next) }
+    nbs
   }
   implicit def cb2rcb(cb: CBitSet): RichCBitSet = new RichCBitSet(cb)
 }
 
 // An enrichment to give some scala-like operators to the compressed
 // bit set.
-class RichCBitSet(val cb: CBitSet) {
+class RichCBitSet(val cb: CBitSet) extends AnyVal {
   def ++(b: CBitSet): CBitSet = cb.or(b)
 
   def ==(b: CBitSet): Boolean = cb.equals(b)
 
+  def +=(xs: Array[Int]): cb.type = {
+    var idx = 0
+    java.util.Arrays.sort(xs)
+    while (idx < xs.length) {
+      cb.set(xs(idx))
+      idx += 1
+    }
+    cb
+  }
+
   def toBitSet(width: Int): BitSet = {
-    val a = new Array[Long]((width + 63) / 64)
-    cb.asScala.foreach{ i: java.lang.Integer => a(i.intValue / 64) |= 1L << (i.intValue % 64) }
-    BitSet.fromBitMask(a)
+    val a = LongBitSet.empty(width)
+    val iter = cb.intIterator
+    while (iter.hasNext) {
+      val i = iter.next
+      a.set(i)
+    }
+    a.toBitSetNoCopy
+  }
+}
+
+private[algebird] case class LongBitSet(toArray: Array[Long]) extends AnyVal {
+  def toBitSetNoCopy: BitSet = {
+    BitSet.fromBitMaskNoCopy(toArray)
+  }
+
+  def set(i: Int): Unit = {
+    toArray(i / 64) |= 1L << (i % 64)
+  }
+
+  def +=(xs: Array[Int]): Unit = {
+    var idx = 0
+    while (idx < xs.length) {
+      set(xs(idx))
+      idx += 1
+    }
+  }
+
+  def +=(it: IntIterator): Unit = {
+    while (it.hasNext) { set(it.next) }
+  }
+}
+
+private[algebird] object LongBitSet {
+  def empty(size: Int): LongBitSet = LongBitSet(new Array[Long]((size + 63) / 64))
+  def fromCBitSet(cb: CBitSet, width: Int): LongBitSet = {
+    val lbs = empty(width)
+    lbs += cb.intIterator
+    lbs
   }
 }
 
 object BloomFilter {
 
-  def apply[A](numEntries: Int, fpProb: Double)(implicit hash: Hash128[A]): BloomFilterMonoid[A] = {
-    val width = BloomFilter.optimalWidth(numEntries, fpProb)
-    val numHashes = BloomFilter.optimalNumHashes(numEntries, width)
-    BloomFilterMonoid[A](numHashes, width)(hash)
-  }
+  def apply[A](numEntries: Int, fpProb: Double)(implicit hash: Hash128[A]): BloomFilterMonoid[A] =
+    BloomFilter.optimalWidth(numEntries, fpProb) match {
+      case None =>
+        throw new java.lang.IllegalArgumentException(
+          s"BloomFilter cannot guarantee the specified false positive probability for the number of entries! (numEntries: $numEntries, fpProb: $fpProb)")
+      case Some(width) =>
+        val numHashes = BloomFilter.optimalNumHashes(numEntries, width)
+        BloomFilterMonoid[A](numHashes, width)(hash)
+    }
 
   // Compute optimal number of hashes: k = m/n ln(2)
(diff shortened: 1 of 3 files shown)


johnynek:
@johandahlberg you may also want to review these changes since you may be using BloomFilters.

johnynek:
Note this blog post commented on how slow our implementation is/was:

https://alexandrnikitin.github.io/blog/bloom-filter-for-scala/

Membership checking could still be optimized some I think. We need to make sure if you are going to query a bunch, you do so on a dense representation. A full BFSparse is going to be SUPER slow for that.

I think we should not be creating really dense sparse representations now, but we should follow up with more benchmarks.

The ultimate thing here would be immutable implementations of the EWAH algorithm for very sparse cases then shifting to an unboxed immutable `HashSet[Int]` implementation.

algebird-core/src/main/scala/com/twitter/algebird/BloomFilter.scala
```diff
@@ -20,49 +20,156 @@ import scala.collection.immutable.BitSet
 import scala.collection.JavaConverters._
 
 import com.googlecode.javaewah.{ EWAHCompressedBitmap => CBitSet }
+import com.googlecode.javaewah.IntIterator
 
 object RichCBitSet {
-  def apply(x: Int*) = {
-    CBitSet.bitmapOf(x.sorted: _*)
+  def apply(xs: Int*): CBitSet = fromArray(xs.toArray)
```
isnotinvain, line 26: in scala, is xs.toArray a no-op? xs will already be an array right?

algebird-core/src/main/scala/com/twitter/algebird/BloomFilter.scala
```diff
@@ -20,49 +20,156 @@ import scala.collection.immutable.BitSet
...
+
+  // this sorts the array in-place
+  def fromArray(x: Array[Int]): CBitSet = {
+    val bs = new CBitSet
```
isnotinvain, line 30: does CBitSet take a size hint or anything like that?

algebird-core/src/main/scala/com/twitter/algebird/BloomFilter.scala
```diff
@@ -20,49 +20,156 @@ import scala.collection.immutable.BitSet
...
 
 // An enrichment to give some scala-like operators to the compressed
 // bit set.
-class RichCBitSet(val cb: CBitSet) {
+class RichCBitSet(val cb: CBitSet) extends AnyVal {
   def ++(b: CBitSet): CBitSet = cb.or(b)
 
   def ==(b: CBitSet): Boolean = cb.equals(b)
 
+  def +=(xs: Array[Int]): cb.type = {
```
isnotinvain, line 49: what does `cb.type` do here?

algebird-core/src/main/scala/com/twitter/algebird/BloomFilter.scala
```diff
@@ -20,49 +20,156 @@ import scala.collection.immutable.BitSet
...
+    var idx = 0
+    java.util.Arrays.sort(xs)
```
isnotinvain, line 51: it's ok to mutate the passed in array?

isnotinvain (commented):
It looks like all the tests for this are via property tests? Or are there some unit tests to? 

It might be a good idea to have a few unit tests (with hard-coded inputs). It looks like with just the property tests, if there's a bug in the Equiv implementation, then all the property tests could be false positives right? 

EG, if the equiv always just returned true, all the tests would pass. So maybe just a unit test for the Equiv implementation? Or some properties that assert certain things are not equal?

algebird-core/src/main/scala/com/twitter/algebird/BloomFilter.scala
johnynek: I don't really know the details, but `Int*` has type `_*[Int]` or something, which is a seq, but not an Array as far as I know.

algebird-core/src/main/scala/com/twitter/algebird/BloomFilter.scala
johnynek: not the examples I see from their code. It is doing compression of the differences between the indices. I don't see immediately how such a hit would be used.

algebird-core/src/main/scala/com/twitter/algebird/BloomFilter.scala
johnynek: `cb.type` means you are returning literally `cb`. So this lets you see that though it is mutated, the return value is the mutated thing.

algebird-core/src/main/scala/com/twitter/algebird/BloomFilter.scala
johnynek: In general, I'd say no, but since this is just a little convenience class for this code, and we only pass `Arrays` below that are ephemeral, I think it is okay.

johnynek:
@isnotinvain this PR has 95% test coverage. That is excellent for something of this size. I know there is the twitter review style of finding reasons not to merge rather than merge, but I think test coverage anxiety is particularly misplaced on this code. As you can see in the code, bloom filters are fairly well tested:

https://github.com/twitter/algebird/blob/develop/algebird-test/src/test/scala/com/twitter/algebird/BloomFilterTest.scala

So, the kinds of tests you want are here:

1: we check that bloomfilters are approximately the same as sets:
https://github.com/twitter/algebird/blob/develop/algebird-test/src/test/scala/com/twitter/algebird/BloomFilterTest.scala#L169

https://github.com/twitter/algebird/blob/develop/algebird-test/src/test/scala/com/twitter/algebird/BloomFilterTest.scala#L223

2. We check a number of ad-hoc properties aside from the law checks I added:
https://github.com/twitter/algebird/blob/develop/algebird-test/src/test/scala/com/twitter/algebird/BloomFilterTest.scala#L175

3. we actually test the hashing implementation against a copy of some old code so we won't accidentally change serialized bloom filters: https://github.com/twitter/algebird/blob/develop/algebird-test/src/test/scala/com/twitter/algebird/BloomFilterTest.scala#L87

It's a bit hard to find some negative test to add. Can you think of one? Only an empty bloom filter can't give a false positive.

The new property tests found several bugs while I was developing the code, so I think they are pretty good. I could add another law, like `if (bfa equiv bfb) (bfa.contains(x) == bfa.contains(x))` for all `x`, but frankly, I think that doing so on a PR that has such high coverage already is a misplaced effort (like, I would rather publish chill for 2.12 for 0.7 and 0.8 branches).

algebird-core/src/main/scala/com/twitter/algebird/BloomFilter.scala
isnotinvain: oh, that's neat

isnotinvain: so it guaruantees that the passed in cb will be `eq` to the return value?

algebird-core/src/main/scala/com/twitter/algebird/BloomFilter.scala
isnotinvain: ok

algebird-core/src/main/scala/com/twitter/algebird/BloomFilter.scala
johnynek: yes. And the compiler verifies that.

isnotinvain:
+1 thanks for the clarifications

PR #612, opened 2017-01-09
Bloom filter distance function

johandahlberg:
I've added a function to compute the Hamming distance between two Bloom filters. I'm not entierly sure if this is general enough that it is of interest to others than me, but I though I'd make a PR and hear what you folks think.

My reason for implementing it here (and for my previous PR regarding Bloom filters) is that I'm looking at implementing Bloom trees as described by [Solomon and Kingsford (2016)](http://www.nature.com/nbt/journal/v34/n3/full/nbt.3442.html), and then they use the Hamming distance between Bloom filters to decide on the path through the tree.

Like I said, if you think that it's to niche feel free to close this and I'll simply just place it in the Bloom tree code I'm working on instead.

### DIFF
--- a/algebird-core/src/main/scala/com/twitter/algebird/BloomFilter.scala
+++ b/algebird-core/src/main/scala/com/twitter/algebird/BloomFilter.scala
@@ -171,6 +171,47 @@ object BloomFilter {
 
     Approximate[Long](nl, n, nr, scala.math.max(0, prob))
   }
+
+  /**
+   * Compute the Hamming distance between the two Bloom filters
+   * `a` and `b`. The distance is defined as the number of bits that
+   * need to change to in order to transform one filter into the other.
+   */
+  def hammingDistance[A](a: BF[A], b: BF[A]): Int = {
+    assert((a.width == b.width) && (a.numHashes == b.numHashes),
+      "To compute a distance between two Bloom filters they" +
+        "have to be of equal width and the same number of hashes." +
+        s"A was of width ${a.width} and had ${a.numHashes}, and" +
+        s"A was of width ${b.width} and had ${b.numHashes}")
+
+    def toBitSet(x: BF[A]): BitSet = {
+      x match {
+        case BFZero(_, _) => BitSet()
+        case BFItem(item, hashes, _) => {
+          val hashvalues = hashes(item)
+          BitSet(hashvalues: _*)
+        }
+        case BFSparse(_, cbitset, width) => {
+          import RichCBitSet._
+          cbitset.toBitSet(width)
+        }
+        case BFInstance(_, bitset, _) => {
+          bitset
+        }
+      }
+    }
+
+    val aSet = toBitSet(a)
+    val bSet = toBitSet(b)
+
+    if (aSet.isEmpty) {
+      b.numBits
+    } else if (bSet.isEmpty) {
+      a.numBits
+    } else {
+      (aSet ^ bSet).size
+    }
+  }
 }
 
 /**
--- a/algebird-core/src/main/scala/com/twitter/algebird/BloomFilter.scala
+++ b/algebird-core/src/main/scala/com/twitter/algebird/BloomFilter.scala
@@ -377,7 +377,7 @@ sealed abstract class BF[A] extends java.io.Serializable {
    * `a` and `b`. The distance is defined as the number of bits that
    * need to change to in order to transform one filter into the other.
    */
-  def hammingDistance[A](that: BF[A]): Int = {
+  def hammingDistance(that: BF[A]): Int = {
     assert((this.width == that.width) && (this.numHashes == that.numHashes),
       "To compute a distance between two Bloom filters they" +
         "have to be of equal width and the same number of hashes." +
--- a/algebird-core/src/main/scala/com/twitter/algebird/BloomFilter.scala
+++ b/algebird-core/src/main/scala/com/twitter/algebird/BloomFilter.scala
@@ -379,9 +379,9 @@ sealed abstract class BF[A] extends java.io.Serializable {
    */
   def hammingDistance(that: BF[A]): Int = {
     assert((this.width == that.width) && (this.numHashes == that.numHashes),
-      "To compute a distance between two Bloom filters they" +
-        "have to be of equal width and the same number of hashes." +
-        s"A was of width ${this.width} and had ${this.numHashes}, and" +
+      "To compute a distance between two Bloom filters they " +
+        "have to be of equal width and the same number of hashes. " +
+        s"A was of width ${this.width} and had ${this.numHashes}, and " +
         s"B was of width ${that.width} and had ${that.numHashes}")
 
     val aSet = this.toBitSet
--- a/algebird-core/src/main/scala/com/twitter/algebird/BloomFilter.scala
+++ b/algebird-core/src/main/scala/com/twitter/algebird/BloomFilter.scala
@@ -378,12 +378,6 @@ sealed abstract class BF[A] extends java.io.Serializable {
    * need to change to in order to transform one filter into the other.
    */
   def hammingDistance(that: BF[A]): Int = {
-    assert((this.width == that.width) && (this.numHashes == that.numHashes),
-      "To compute a distance between two Bloom filters they " +
-        "have to be of equal width and the same number of hashes. " +
-        s"A was of width ${this.width} and had ${this.numHashes} hashes, and " +
-        s"B was of width ${that.width} and had ${that.numHashes} hashes.")
-
     val aSet = this.toBitSet
     val bSet = that.toBitSet
 
--- a/algebird-core/src/main/scala/com/twitter/algebird/BloomFilter.scala
+++ b/algebird-core/src/main/scala/com/twitter/algebird/BloomFilter.scala
@@ -171,47 +171,6 @@ object BloomFilter {
 
     Approximate[Long](nl, n, nr, scala.math.max(0, prob))
   }
-
-  /**
-   * Compute the Hamming distance between the two Bloom filters
-   * `a` and `b`. The distance is defined as the number of bits that
-   * need to change to in order to transform one filter into the other.
-   */
-  def hammingDistance[A](a: BF[A], b: BF[A]): Int = {
-    assert((a.width == b.width) && (a.numHashes == b.numHashes),
-      "To compute a distance between two Bloom filters they" +
-        "have to be of equal width and the same number of hashes." +
-        s"A was of width ${a.width} and had ${a.numHashes}, and" +
-        s"A was of width ${b.width} and had ${b.numHashes}")
-
-    def toBitSet(x: BF[A]): BitSet = {
-      x match {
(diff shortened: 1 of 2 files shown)


algebird-core/src/main/scala/com/twitter/algebird/BloomFilter.scala
```diff
@@ -171,6 +171,47 @@ object BloomFilter {
...
+  def hammingDistance[A](a: BF[A], b: BF[A]): Int = {
+    assert((a.width == b.width) && (a.numHashes == b.numHashes),
+      "To compute a distance between two Bloom filters they" +
+        "have to be of equal width and the same number of hashes." +
+        s"A was of width ${a.width} and had ${a.numHashes}, and" +
+        s"A was of width ${b.width} and had ${b.numHashes}")
+
+    def toBitSet(x: BF[A]): BitSet = {
+      x match {
+        case BFZero(_, _) => BitSet()
```
johnynek: this branch is untested. I think we need to generate `BFZero` in the test. Can you explicitly add some tests there?

algebird-test/src/test/scala/com/twitter/algebird/BloomFilterTest.scala
```diff
@@ -51,6 +54,27 @@ class BloomFilterLaws extends CheckProperties {
     }
   }
 
+  property("the distance between a filter and itself should be 0") {
+    forAll { (a: BF[String]) =>
+      BloomFilter.hammingDistance(a, a) == 0
+    }
+  }
```
johnynek, line 68: can we do: `forAll { (a: BF[String]) => BloomFilter.hammingDistance(a, BFZero) == a.numBits }`?

algebird-test/src/test/scala/com/twitter/algebird/BloomFilterTest.scala
```diff
@@ -51,6 +54,27 @@ class BloomFilterLaws extends CheckProperties {
...
+
+  property("all equivalent filters should have 0 Hamming distance") {
+    forAll { (a: BF[String], b: BF[String]) =>
+      if (Equiv[BF[String]].equiv(a, b))
+        BloomFilter.hammingDistance(a, b) == 0
+      else
+        BloomFilter.hammingDistance(a, b) != 0
```
johnynek: can you write `> 0` here?

algebird-test/src/test/scala/com/twitter/algebird/BloomFilterTest.scala
```diff
@@ -376,4 +400,56 @@ class BloomFilterTest extends WordSpec with Matchers {
...
+      val secondBloomFilter = createBFWithItems(Seq("C"))
+
+      val distance1 = BloomFilter.hammingDistance(firstBloomFilter, secondBloomFilter)
+      assert(distance1 === 4)
+
+      val thirdBloomFilter = createBFWithItems(Seq("A", "B", "C"))
+      // Make it dense to make sure that that case is also covered
+      // even though these examples are small and thus sparse.
+      val forthBloomFilter = toDense(createBFWithItems(Seq("C", "D", "E")))
+
```
johnynek, line 434: can you add the `createBFWithItems(List())` case as well, just to test emptiness again?

algebird-core/src/main/scala/com/twitter/algebird/BloomFilter.scala
```diff
@@ -171,6 +171,47 @@ object BloomFilter {
...
+   * need to change to in order to transform one filter into the other.
+   */
+  def hammingDistance[A](a: BF[A], b: BF[A]): Int = {
+    assert((a.width == b.width) && (a.numHashes == b.numHashes),
+      "To compute a distance between two Bloom filters they" +
+        "have to be of equal width and the same number of hashes." +
+        s"A was of width ${a.width} and had ${a.numHashes}, and" +
+        s"A was of width ${b.width} and had ${b.numHashes}")
+
+    def toBitSet(x: BF[A]): BitSet = {
```
johnynek: can you add this to `abstract class BF[A]` instead? I'd rather have `a.hammingDistance(b)`.

algebird-test/src/test/scala/com/twitter/algebird/BloomFilterTest.scala
johnynek: also `<= width`, right?

johnynek (changes requested):
Looks great, just let's add some minor changes.

johandahlberg:
Cool. Good suggestions for changes. I'll most likely look into fixing this when I get home tonight.

johandahlberg:
I'm not familiar with the codecov tool, so I'm a bit surprised that it is reporting a decrease in coverage. When looking at it by eye  it seems like all of the new code introduced is covered, but maybe I'm missing something.

algebird-core/src/main/scala/com/twitter/algebird/BloomFilter.scala
```diff
@@ -369,12 +369,41 @@ sealed abstract class BF[A] extends java.io.Serializable {
...
+
+  def toBitSet: BitSet
+
+  /**
+   * Compute the Hamming distance between the two Bloom filters
+   * `a` and `b`. The distance is defined as the number of bits that
+   * need to change to in order to transform one filter into the other.
+   */
+  def hammingDistance(that: BF[A]): Int = {
+    assert((this.width == that.width) && (this.numHashes == that.numHashes),
```
johnynek: Do we really need this assert? The code below does not seem to actually require it.

johnynek (commented):
I'd like to remove the assert if we can. I don't think hamming actually cares what the number of hashes were nor the width.

johandahlberg:
As you say the assert is not strictly necessary. I added it because I was thinking that even if it is possible it would not be a valid comparison to compare filters with different parameters, however thinking more about it now I think you're right.

I'll remove the assert and related tests, asap.

algebird-core/src/main/scala/com/twitter/algebird/BloomFilter.scala
pankajroark: I wonder how frequently this method will be used. If not frequently then may be this should go into the companion object to avoid cluttering the base class.

algebird-test/src/test/scala/com/twitter/algebird/BloomFilterTest.scala
```diff
@@ -51,6 +54,36 @@ class BloomFilterLaws extends CheckProperties {
     }
   }
 
+  property("the distance between a filter and itself should be 0") {
+    forAll { (a: BF[String]) =>
+      a.hammingDistance(a) == 0
+    }
+  }
+
+  property("the distance between a filter and an empty filter should be the the number of bits" +
```
pankajroark: Extra "the".

algebird-test/src/test/scala/com/twitter/algebird/BloomFilterTest.scala
```diff
@@ -376,4 +409,61 @@ class BloomFilterTest extends WordSpec with Matchers {
       }
     }
   }
+
+  "BloomFilters" should {
+    "should not be able to compute Hamming distance to each other when they are not " +
```
pankajroark: extra "should"

algebird-test/src/test/scala/com/twitter/algebird/BloomFilterTest.scala
```diff
@@ -376,4 +409,61 @@ class BloomFilterTest extends WordSpec with Matchers {
...
+          bf1.hammingDistance(bf2)
+        }
+
+        assertThrows[AssertionError]{
+          bf2.hammingDistance(bf3)
+        }
+
+      }
+
+    "be able to compute compute Hamming distance to each other" in {
```
pankajroark: Extra "compute".

algebird-core/src/main/scala/com/twitter/algebird/BloomFilter.scala
```diff
@@ -369,12 +369,41 @@ sealed abstract class BF[A] extends java.io.Serializable {
...
+      "To compute a distance between two Bloom filters they " +
+        "have to be of equal width and the same number of hashes. " +
+        s"A was of width ${this.width} and had ${this.numHashes} hashes, and " +
+        s"B was of width ${that.width} and had ${that.numHashes} hashes.")
+
+    val aSet = this.toBitSet
```
pankajroark, line 381: Note that this can be made more efficient. BFZero and BFItem can be handled specially and underlying compressed bitmap of BFSparse has a xorCardinality method that avoids intermediate allocations.

pankajroark (changes requested):

algebird-core/src/main/scala/com/twitter/algebird/BloomFilter.scala
johandahlberg: I'd be happy to move this anywhere you think it should go. I started out keeping it in the `BloomFilter` and I can place it back in there, or move it to the `BF` companion object if you prefer that.

johandahlberg:
@pankajroark Thank you for the suggestions, I've fixed the typos. Let me know if you do want me to move the function into the `BloomFilter` or `BF` objects. I guess it also depends on what @johnynek prefers?

When it comes to the performance problems you mentioned I'm not 100% certain how to actually solve them. Did you mean computing the distance in a specialized way for each of the cases? I.e. something like adding a specific `def hammingDistance(that: BF[A]): Int` which would then pattern match on the input and perform the most performant distance calculation for each case?

pankajroark:
@johandahlberg Yeah, I was thinking that special cases could be optimized. Not necessarily by adding specific hemming distance functions for all cases but you could pattern match in the companion function:
(x, y) match {
  case (BFZero, BFZero) => 
  case (BFSparse, BFSparse) => 
  case - => // fallback to creating bitsets
...
}

It doesn't seem necessary for this change, just wanted to raise the possibility. I'll let @johnynek suggest about whether it should go into the base class or companion.

johnynek:
@johandahlberg up to you if you want to optimize now. The signatures do not prevent optimization.

One thing I am thinking is that we could move the `Equiv` code to this. That code is somewhat optimized. So, `a == b` iff `a.hammingDistance(b) == 0`. So Equiv could just call an optimized hammingDistance.

If you don't have time, I can try to find time before we release. Ideally we will soon release algebird for 2.12, and we should go ahead and get this in that branch.

johnynek:
(sorry, clicked the wrong button).

johandahlberg:
@johnynek Depending on how pressed for time you are on this it might be better if you have a look at optimizing it. While I could look into it I'm not sure how much time I will be able to put into this in the next few days, and since it will probably require a bit more back and forth between us before it's completed it might be faster if you look at it.

And using the the hamming distance to determine equality sounds very reasonable to me.

johnynek (approved):

johnynek:
yes. Let's merge and optimize as needed. This won't slow the most common use cases and I'll try to take a stab at it before release.

PR #614, opened 2017-01-12
Catch OOM and consider it a pass in SpaceSaverTest

johnynek:
we generate random bytes and try to deserialize. On travis this sometimes OOMs


piyushnarang:
👍

PR #620, opened 2017-02-02
Use more algebra types

johnynek:
This labels many (I hope all) types we have with more precise types:

see:
https://github.com/typelevel/algebra/blob/master/docs/src/main/tut/typeclasses/overview.md

for an overview of the types.

@sritchie @piyushnarang @non please take a look. I want to get as many of these as possible added if we can so we don't break compatibility by wanting to add some we missed.

### DIFF
--- a/algebird-core/src/main/scala/com/twitter/algebird/Max.scala
+++ b/algebird-core/src/main/scala/com/twitter/algebird/Max.scala
@@ -17,6 +17,8 @@ package com.twitter.algebird
 
 import scala.annotation.tailrec
 
+import algebra.{ BoundedSemilattice, Semilattice }
+
 /**
  * Tracks the maximum wrapped instance of some ordered type `T`.
  *
@@ -58,20 +60,16 @@ object Max extends MaxInstances {
   /**
    * Returns a [[Semigroup]] instance with a `plus` implementation
    * that always returns the maximum `T` argument.
-   *
-   * @todo Note that this is a semilattice, for when we merge  typelevel/algebra.
    */
-  def maxSemigroup[T](implicit ord: Ordering[T]): Semigroup[T] =
-    Semigroup.from { (l: T, r: T) => ord.max(l, r) }
+  def maxSemigroup[T](implicit ord: Ordering[T]): Semigroup[T] with Semilattice[T] =
+    new Semigroup[T] with Semilattice[T] {
+      def plus(l: T, r: T): T = ord.max(l, r)
+    }
 }
 
 private[algebird] sealed abstract class MaxInstances {
   implicit def equiv[T](implicit eq: Equiv[T]): Equiv[Max[T]] = Equiv.by(_.get)
   implicit def ordering[T: Ordering]: Ordering[Max[T]] = Ordering.by(_.get)
-
-  private[this] def plus[T](implicit ord: Ordering[T]) = {
-    (l: Max[T], r: Max[T]) => if (ord.gteq(l.get, r.get)) l else r
-  }
 
   /**
    * Returns a [[Monoid]] instance for [[Max]][T] that combines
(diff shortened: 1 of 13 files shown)


algebird-core/src/main/scala/com/twitter/algebird/Max.scala
```diff
@@ -58,51 +60,55 @@ object Max extends MaxInstances {
...
-
   /**
    * Returns a [[Monoid]] instance for [[Max]][T] that combines
    * instances using [[Max.max]] and uses `zero` for its identity.
    *
    * @param zero identity of the returned [[Monoid]] instance
    * @note `zero` must be `<=` every element of `T` for the returned instance to be lawful.
    */
-  def monoid[T: Ordering](zero: => T): Monoid[Max[T]] = Monoid.from(Max(zero))(plus)
+  def monoid[T: Ordering](z: => T): Monoid[Max[T]] with BoundedSemilattice[Max[T]] =
```
piyushnarang: leave the name as zero?

algebird-core/src/main/scala/com/twitter/algebird/Min.scala
```diff
@@ -57,44 +59,51 @@ object Min extends MinInstances {
...
-
   /**
    * Returns a [[Monoid]] instance for [[Min]][T] that combines
    * instances using [[Min.min]] and uses `zero` for its identity.
    *
    * @param zero identity of the returned [[Monoid]] instance
    * @note `zero` must be `>=` every element of `T` for the returned instance to be lawful.
    */
-  def monoid[T: Ordering](zero: => T): Monoid[Min[T]] = Monoid.from(Min(zero))(plus)
+  def monoid[T: Ordering](z: => T): Monoid[Min[T]] with BoundedSemilattice[Min[T]] =
```
piyushnarang: s/z/zero?

algebird-core/src/main/scala/com/twitter/algebird/Ring.scala
```diff
@@ -161,6 +161,35 @@ class FromAlgebraRing[T](r: ARing[T]) extends Ring[T] {
   override def product(ts: TraversableOnce[T]): T = r.product(ts)
 }
 
+class UnsafeFromAlgebraRig[T](r: Rig[T]) extends Ring[T] {
```
piyushnarang: curious as to why this is extending Ring? can we not just extend Rig? Maybe document the rationale?

algebird-core/src/main/scala/com/twitter/algebird/Ring.scala
```diff
@@ -161,6 +161,35 @@ class FromAlgebraRing[T](r: ARing[T]) extends Ring[T] {
...
+    sys.error(s"$r is a Rig, there is no additive inverse")
+  override def minus(a: T, b: T): T =
+    sys.error(s"$r is a Rig, there is no additive inverse")
+  override def sum(ts: TraversableOnce[T]): T = r.sum(ts)
+  override def sumOption(ts: TraversableOnce[T]): Option[T] = r.trySum(ts)
+  override def times(a: T, b: T): T = r.times(a, b)
+  override def product(ts: TraversableOnce[T]): T = r.product(ts)
+}
+
+class UnsafeFromAlgebraRng[T](r: Rng[T]) extends Ring[T] {
```
piyushnarang: here too?

piyushnarang (approved):
This is pretty cool stuff :-). Would wait for others to take a look too as I'm not the most familiar with the code.

algebird-core/src/main/scala/com/twitter/algebird/Max.scala
johnynek: then whe we override `zero` below you have the chance for confusion (in fact, you can create an infinite loop if you are not careful).

johnynek:
@piyushnarang addressed your concerns, I think.

algebird-core/src/main/scala/com/twitter/algebird/Max.scala
piyushnarang: ah yeah :-)

piyushnarang:
Great thanks for the updates.

algebird-core/src/main/scala/com/twitter/algebird/Ring.scala
```diff
@@ -202,8 +243,14 @@ object Ring extends GeneratedRingImplicits with ProductRings with RingImplicits0
   implicit def doubleRing: Ring[Double] = DoubleRing
   implicit def jdoubleRing: Ring[JDouble] = JDoubleRing
   implicit def indexedSeqRing[T: Ring]: Ring[IndexedSeq[T]] = new IndexedSeqRing[T]
+  /**
+   * This is actually a Rng but we leave the unsafe Ring for legacy reasons
+   */
   implicit def mapRing[K, V](implicit ring: Ring[V]): Ring[Map[K, V]] =
```
tixxit, line 249: Did you mean to have the implicit parameter be `Rng[V]` here?

algebird-core/src/main/scala/com/twitter/algebird/Ring.scala
```diff
@@ -202,8 +243,14 @@ object Ring extends GeneratedRingImplicits with ProductRings with RingImplicits0
...
-    new MapRing[K, V]()(ring)
+    new UnsafeFromAlgebraRng(new MapRing[K, V]()(ring))
+  /**
+   * This is actually a Rng but we leave the unsafe Ring for legacy reasons
+   */
   implicit def scMapRing[K, V](implicit ring: Ring[V]): Ring[scala.collection.Map[K, V]] =
```
tixxit, line 254: Same here.

tixxit (changes requested):
Looks good - just a minor typo I think.

algebird-core/src/main/scala/com/twitter/algebird/Ring.scala
johnynek: sadly no, because algebra does not use the import-to-get-instances style. If we did that it would break source compatibility for people that expect to get Map rings.

piyushnarang:
Ping @sritchie / @non I think this PR is the last one blocking our new release and it would be nice to get that out soon.

johnynek:
@tixxit said it looked good just noticed two ugly issues. Not easy to fix
and keep compatibility for source so I suggest we don't.

Unless there are objections, I think we should merge and publish 0.13.0
tomorrow.

PR #622, opened 2017-02-16
Address ambiguous implicit for Semigroup[Max[String]]

alexarchambault:
Currently, `Max.stringMonoid ` and `Max.semigroup` collide when looking up for a `Semigroup[Max[String]]`. This PR addresses this ambiguity by giving `Max.stringMonoid` a higher priority.

Note that this problem seems to happen only for `String`, not the primitive types that have specific implicit instances too, but with more specific types. This PR still moves the latter instances along with `stringMonoid`, for readability.

### DIFF
--- a/algebird-core/src/main/scala/com/twitter/algebird/Max.scala
+++ b/algebird-core/src/main/scala/com/twitter/algebird/Max.scala
@@ -67,7 +67,32 @@ object Max extends MaxInstances {
     }
 }
 
-private[algebird] sealed abstract class MaxInstances {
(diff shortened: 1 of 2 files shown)


johnynek:
thanks for catching this. Bummer we didn't notice sooner.

alexarchambault:
@johnynek It was fine in `0.12.3`, I just ran into this when switching to `0.13.0`.

johnynek:
Huh. That's strange. Wish it was easier to auto generate implicit
resolution tests.

PR #624, opened 2017-02-17
Removed claim for 2.12 support from readme

slouc:
It's not true since algebird_2.12 is not out afaik, plus https://index.scala-lang.org/twitter/algebird says 2.10 and 2.11 only. Let me know if I'm wrong.

### DIFF
--- a/README.md
+++ b/README.md
@@ -66,7 +66,7 @@ A list of contributors to the project can be found here: [Contributors](https://
 
 Algebird modules are available on maven central. The current groupid and version for all modules is, respectively, `"com.twitter"` and  `0.13.0`.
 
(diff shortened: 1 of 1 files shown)


johnynek:
I'm sorry that index is out of date, I don't maintain it.

2.12 is ready:

https://search.maven.org/#search%7Cga%7C1%7CAlgebird%20

johnynek:
See: https://search.maven.org/#search%7Cga%7C1%7CAlgebird%20

slouc:
Ah ok. I was trying to get the full "algebird" artifact and it wasn't found, but now I see that separate modules are on 2.12. Thanks! Do you know if the whole algebird artifact will be there too?

johnynek:
Yes. You are using the wrong artifact ID. It has not been Algebird for
quite some time. algebird-core is the artifact ID.

slouc:
Got it, thanks!

johnynek:
👍

PR #625, opened 2017-02-17
Re-enable mima on 0.13.0 branch

johnynek:
would be nice to publish frequent updates and hopefully keep binary compatibility for a long time.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,19 +3,19 @@ sudo: false
 matrix:
   include:
     - scala: 2.10.6
-      # script: ./sbt ++$TRAVIS_SCALA_VERSION clean test mimaReportBinaryIssues"
-      script: ./sbt ++$TRAVIS_SCALA_VERSION clean test
+      script: ./sbt ++$TRAVIS_SCALA_VERSION clean test mimaReportBinaryIssues"
+      #script: ./sbt ++$TRAVIS_SCALA_VERSION clean test
 
     - scala: 2.11.8
-      #script: ./sbt ++$TRAVIS_SCALA_VERSION clean coverage test coverageReport mimaReportBinaryIssues
-      script: ./sbt ++$TRAVIS_SCALA_VERSION clean coverage test coverageReport
+      script: ./sbt ++$TRAVIS_SCALA_VERSION clean coverage test coverageReport mimaReportBinaryIssues
+      #script: ./sbt ++$TRAVIS_SCALA_VERSION clean coverage test coverageReport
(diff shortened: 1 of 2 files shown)


ulyssepence:
From the MiMa wiki https://github.com/typesafehub/migration-manager/wiki :

> A key aspect of MiMa to be aware of is that it only looks for syntactic binary incompatibilities. The semantic binary incompatibilities (such as adding or removing a method invocation) are not considered

I don't really understand this distinction. Sorry for my ignorance, @johnynek, can you describe what issues this will prevent (and/or cite any issues this would have avoided in the past)?

johnynek:
We have used this for years. We turn it off when we make an incompatible change. This prevents us from changing, for instance, the return type of a method, which then gets pulled into a runtime classpath by the maven/ivy approach of taking the latest version, then crashing at runtime. The idea is the version numbers should tell you when you are safe to upgrade (but the tooling is not so great on the jvm, honestly).

The main thing it does is prevent us from making gratuitous binary incompatible changes.

The issue of a semantic vs a syntactic change, consider the following:
```scala
// syntactically incompatible:
def foo(a: Int): String = a.toString
// change to:
def foo(a: Int): Long = a.toLong
// Long is not a String!

// semantically incompatible
def foo(a: Int): String = a.toString
def foo(a: Int): String = s"The int is $a"
// the types are the same, but the meaning of the method is different.
// should never really make this kind of change since it massively confuses consumers.
```

sritchie-stripe:
👍  LGTM @johnynek

.travis.yml
```diff
@@ -3,19 +3,19 @@ sudo: false
 matrix:
   include:
     - scala: 2.10.6
-      # script: ./sbt ++$TRAVIS_SCALA_VERSION clean test mimaReportBinaryIssues"
-      script: ./sbt ++$TRAVIS_SCALA_VERSION clean test
+      script: ./sbt ++$TRAVIS_SCALA_VERSION clean test mimaReportBinaryIssues
+      #script: ./sbt ++$TRAVIS_SCALA_VERSION clean test
```
piyushnarang, line 7: should we just delete this line? and the two other commented ones?

piyushnarang (approved):
Thanks for turning this back on!

ulyssepence:
+1 Thanks @johnynek

.travis.yml
johnynek: I left them to make it easier in the future since we will, probably, turn it off again.

PR #627, opened 2017-03-02
Add methods to Interval

johnynek:
Working with Interval in some cases is painful to deal with.

This tries to add a few methods that make sense.

I am on the fence about introducing a type like:
```scala
sealed abstract class MaybeUnbounded[+T]
object MaybeUnbounded {
  // This does not exist
  case object Empty extends MaybeUnbounded[Nothing]
  // Infinite
  case object Unbounded extends MaybeUnbounded[Nothing]
  case class Bounded(get: T) extends MaybeUnbounded[T]
}
```
In this way, `def least: MaybeUnbounded[T]` makes sense, since `Universe()` has no least, so it is unbounded, empty is not unbounded, but something with a known lowerbound is Bounded.

cc @non

### DIFF
--- a/algebird-core/src/main/scala/com/twitter/algebird/Interval.scala
+++ b/algebird-core/src/main/scala/com/twitter/algebird/Interval.scala
@@ -101,6 +101,69 @@ object Interval extends java.io.Serializable {
     if (Ordering[T].lt(lower, upper))
       MaybeEmpty.NotSoEmpty[T, ExLowInUp](Intersection(ExclusiveLower(lower), InclusiveUpper(upper)))
     else MaybeEmpty.SoEmpty[T, ExLowInUp]()
+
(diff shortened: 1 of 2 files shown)


non:
These methods seem useful to me, so 👍 

I don't have strong feelings regarding your API question.

johnynek:
@pankajroark @isnotinvain review?

sritchie-stripe:
LGTM except for the shitty coverage ;)

johnynek:
actually good point. 100% or bust.

pankajroark:
Don't have any opinions on introducing MaybeUnbounded, looks ok otherwise. 👍

johnynek:
@sritchie-stripe how does the coverage look? (should be 100% when coverage updates).

johnynek:
diff coverage 100% and I didn't change code (just added tests) since @pankajroark gave me the thumbs up, so I'm going to merge.

Issue #583, opened 2016-11-28
AdaptiveVector `isNonZero` is broken


sritchie-stripe:
Here's an example of the breakage:

```scala
import com.twitter.algebird._
val a = SparseVector(Map.empty, 0, 5)
val b = AdaptiveVector.fromVector(Vector(1), 0)
Monoid.zeroEquiv[AdaptiveVector[Int]].equiv(Monoid.plus(a, b), b)
// => false
```

`Monoid.plus` returns `SparseVector(1, 0, 0, 0, 0)`, extending the first vector. We want to be able to add zero-only vectors of any length and treat them as identity.

sritchie-stripe:
This is only a problem when the sparse value is the same as the zero of the underlying monoid.

johnynek:
this is still flakey:

`AdaptiveVector[Int] has a monoid`

mio-stripe:
Hi all, two questions

1. Could we add two more cases to [`AVSemigroup.plus`](https://github.com/twitter/algebird/blob/1520068ae296d65ce3c4af7a0dc40137349f76f0/algebird-core/src/main/scala/com/twitter/algebird/AdaptiveVector.scala#L98-L112) to check if either argument `isZero`
```
def plus(left: AdaptiveVector[V], right: AdaptiveVector[V]) = {
...
  (left, right) match {
    case _ if isZero(left) => right
    case _ if isZero(right) => left
    case (DenseVector(lv, ls, ld), DenseVector(rv, rs, rd)) => ...
}
```
This bug is happening because the size of the result is getting set to the larger of the two arguments which is the right thing to do as long as neither argument `isZero`.

2. While we're at it, could we simplify [`AVMonoid.isZero`](https://github.com/twitter/algebird/blob/1520068ae296d65ce3c4af7a0dc40137349f76f0/algebird-core/src/main/scala/com/twitter/algebird/AdaptiveVector.scala#L121-L124) to
```
def isZero(v: AdaptiveVector[V]) = (v.size == 0) || v.denseIterator.isEmpty
```

mio-stripe:
Whoops! Scratch (2), I get it now: `isZero` should return `true` on `val c = SparseVector(Map(0 -> 0, 1 -> 0, 2 -> 0), 2, 3)` where the `sparseValue` is not `Monoid.zero`, but all of its values _are_ `Monoid.zero`.

[...]

PR #629, opened 2017-04-04
Mio/issue 583

mio-stripe:
Fixing [issue 583](https://github.com/twitter/algebird/issues/583) that `AdaptiveVector` monoid fails monoid laws when the `sparseValue` is `monoid.zero`.

### DIFF
--- a/algebird-test/src/test/scala/com/twitter/algebird/CollectionSpecification.scala
+++ b/algebird-test/src/test/scala/com/twitter/algebird/CollectionSpecification.scala
@@ -1,6 +1,6 @@
 package com.twitter.algebird
 
-import org.scalacheck.{ Arbitrary, Gen }
+import org.scalacheck.{ Arbitrary, Gen, Prop }
 
 import scala.collection.mutable.{ Map => MMap }
 import scala.collection.{ Map => ScMap }
@@ -299,21 +299,28 @@ class CollectionSpecification extends CheckProperties {
         size <- Gen.posNum[Int]
       } yield AdaptiveVector.fromMap(Map.empty, sparse, size))
 
+
+  def dontSumSparseValues[T: Semigroup: Arbitrary]: Prop =
+    'dontSumSparseValues |: forAll { (a: T) =>
+      val sem: Semigroup[T] = implicitly
+      def denseCount(v: T): Int = v match {
+        case v: AdaptiveVector[_] => v.denseCount
+        case _ => 0
+      }
+      denseCount(sem.plus(a,a)) == denseCount(a)
+    }
+
   property("AdaptiveVector[Int] has a semigroup") {
     implicit val arb = Arbitrary(arbAV(2))
     semigroupLaws[AdaptiveVector[Int]]
   }
 
   property("AdaptiveVector[Int] has a monoid when sparseValue IS monoid.zero") {
-    // TODO: remove this equiv instance once #583 is resolved.
-    // implicit val equiv = AdaptiveVector.denseEquiv[Int]
     implicit val arb = Arbitrary(arbAV(0))
     monoidLaws[AdaptiveVector[Int]]
   }
 
   property("AdaptiveVector[Int] has a monoid when sparseValue is NOT monoid.zero") {
-    // TODO: remove this equiv instance once #583 is resolved.
-    // implicit val equiv = AdaptiveVector.denseEquiv[Int]
     implicit val arb = Arbitrary(arbAV(2))
     monoidLaws[AdaptiveVector[Int]]
   }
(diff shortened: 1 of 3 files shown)


algebird-test/src/test/scala/com/twitter/algebird/CollectionSpecification.scala
```diff
@@ -299,27 +299,50 @@ class CollectionSpecification extends CheckProperties {
...
 
+
+  def dontSumSparseValues[T: Semigroup: Arbitrary]: Prop =
+    'dontSumSparseValues |: forAll { (a: T) =>
+      val sem: Semigroup[T] = implicitly
+      def denseCount(v: T): Int = v match {
+        case v: AdaptiveVector[_] => v.denseCount
+        case _ => 0
+      }
+      denseCount(sem.plus(a,a)) == denseCount(a)
```
johnynek: don't need to capture `sem`. `Semgroup.plus(a, a)` would work.

algebird-test/src/test/scala/com/twitter/algebird/CollectionSpecification.scala
```diff
@@ -299,27 +299,50 @@ class CollectionSpecification extends CheckProperties {
...
+
   property("AdaptiveVector[Int] has a semigroup") {
     implicit val arb = Arbitrary(arbAV(2))
     semigroupLaws[AdaptiveVector[Int]]
   }
 
-  property("AdaptiveVector[Int] has a monoid") {
-    // TODO: remove this equiv instance once #583 is resolved.
-    implicit val equiv = AdaptiveVector.denseEquiv[Int]
+  property("AdaptiveVector[Int] has a monoid when sparseValue IS monoid.zero") {
```
johnynek: what is up with the `IS`?

build.sbt
```diff
@@ -183,8 +183,8 @@ lazy val algebird = Project(
   algebirdCore,
   algebirdUtil,
   algebirdBijection,
-  algebirdBenchmark,
-  algebirdSpark
+  algebirdBenchmark//,
+  //algebirdSpark
```
johnynek: let's not comment out algebirdSpark. It is a pain that `2.12.x` won't build, but CI knows about it. Can build with `sbt ++2.11.8 compile` on the command line.

johnynek (commented):
Thanks for doing this! very helpful.

A few minor comments.

build.sbt
mio-stripe: Will change, didn't mean to commit this. Cool, I'll build that way.

algebird-test/src/test/scala/com/twitter/algebird/CollectionSpecification.scala
mio-stripe: I'll change these, I got into doing that as a note-to-self.

algebird-test/src/test/scala/com/twitter/algebird/CollectionSpecification.scala
mio-stripe: 👍

mio-stripe:
So, what I discovered is that this is a bit confusing because the algebra of sparse values is entwined with the algebra of values you care about. The clean solution to this would be to make this `AdpativeVector[V, SV]` where the type of the sparse value is distinct from the type of the values, then be able to define the algebra of sparse values however you want. This way, `AVSemigroup.plus` could add sparse values or not depending on your semigroup structure on `SV`.  This is not what I did, but I'm happy to make that generalization if there is interest. What I did is update `plus` and `equiv` so that

* adding AdpativeVectors does not add their sparseValues

* explicitly check if a summand to AVSemigroup.plus is AVMonoid.zero or if it is a zero vector so that both `AVMonoid.zero` and  other zero vectors all act as additive identities

* came up what I think is a reasonable definition of `equiv`

* got all the tests to pass except the `AdaptiveVector[Int] is a group (when sparseValue is 1)` because the group structure for AdaptiveVector relies on adding the sparseValues.
I changed this test to check the group properties when the sparse value is 0 and it passes.

johnynek:
@mio-stripe is this ready to merge? Seems like it might be?

No longer WIP?

johnynek:
this looks good, but now I'm worried that the plus is so expensive we are killing any optimizations that this class intends to represent.

Maybe we should write a benchmark in the algebird-benchmarks and compare `Vector[T]` with `AdaptiveVector[T]` when we have a lot of sparsity.

mio-stripe:
@johnynek ready to merge! Can we add the benchmark after or should we do it before merging this?

johnynek:
I needed to publish a version. I don't want to publish this without a benchmark, so I'm going to keep this open until we have the benchmark. I can find time to pair with you if you want sometime soon.

PR #631, opened 2017-04-19
Add Scio to "Projects using Algebird"

nevillelyh:


johnynek:
Thanks and congrats on scio! looks great!

PR #633, opened 2017-05-30
Binary Classification Confusion Matrix and AUC Aggregators

richwhitjr:
Some of this works is derived from similar functions in the Spark library. For Binary Classifications tasks it will compute a confusion matrix.  You can also aggregator over these confusions matrices with different thresholds to compute an Area under the Curve Stat for both PR and ROC.

### DIFF
--- a/algebird-core/src/main/scala/com/twitter/algebird/BinaryClassificationAUC.scala
+++ b/algebird-core/src/main/scala/com/twitter/algebird/BinaryClassificationAUC.scala
@@ -1,0 +1,94 @@
+package com.twitter.algebird
+
+/**
+ * Curve is a list of Confusion Matrices with different
+ * thresholds
+ *
+ * @param matrices List of Matrices
+ */
+case class Curve(matrices: List[ConfusionMatrix])
+
+/**
+ * Given a List of (x,y) this functions computes the
+ * Area Under the Curve
+ */
+object AreaUnderCurve {
+  private def trapezoid(points: Seq[(Double, Double)]): Double = {
+    require(points.length == 2)
+    val x = points.head
+    val y = points.last
+    (y._1 - x._1) * (y._2 + x._2) / 2.0
+  }
+
+  def of(curve: List[(Double, Double)]): Double = {
+    curve.toIterator.sliding(2).withPartial(false).aggregate(0.0)(
+      seqop = (auc: Double, points: Seq[(Double, Double)]) => auc + trapezoid(points),
+      combop = _ + _)
+  }
+}
+
+sealed trait AUCMetric
+case object ROC extends AUCMetric
+case object PR extends AUCMetric
+
+/**
+ * Sums Curves which are a series of Confusion Matrices
+ * with different thresholds
+ */
+case object CurveMonoid extends Monoid[Curve] {
+  def zero = Curve(Nil)
+  override def plus(left: Curve, right: Curve): Curve = {
+    val sg = BinaryClassificationConfusionMatrixMonoid
+    Curve(
+      left.matrices.zip(right.matrices)
+        .map{ case (cl, cr) => sg.plus(cl, cr) })
+  }
+}
+
+/**
+ * AUCAggregator computes the Area Under the Curve
+ * for a given metric by sampling along that curve.
+ *
+ * The number of samples is taken and is used to compute
+ * the thresholds to use. A confusion matrix is then computed
+ * for each threshold and finally that is used to compute the
+ * Area Under the Curve.
+ *
+ * Note this is for Binary Classifications Tasks
+ *
+ * @param metric Which Metric to compute
+ * @param samples Number of samples, defaults to 100
+ */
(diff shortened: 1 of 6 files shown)


sritchie-stripe:
Hey @richwhitjr, thank you for this! We should have some time to look today.

algebird-core/src/main/scala/com/twitter/algebird/BinaryClassificationAUC.scala
```diff
@@ -0,0 +1,110 @@
...
+ */
+case class Curve(matrices: List[ConfusionMatrix])
+
+/**
+ * Given a List of (x,y) this functions computes the
+ * Area Under the Curve
+ */
+object AreaUnderCurve {
+  private def trapezoid(points: Seq[(Double, Double)]): Double = {
+    require(points.length == 2)
```
sritchie-stripe: do you want to take the two points directly instead of using this assertion, head, last combo?

algebird-core/src/main/scala/com/twitter/algebird/BinaryClassificationAUC.scala
```diff
@@ -0,0 +1,110 @@
...
+sealed trait AUCMetric
+case object ROC extends AUCMetric
+case object PR extends AUCMetric
+
+/**
+ * Sums Curves which are a series of Confusion Matrices
+ * with different thresholds
+ */
+case object CurveMonoid extends Monoid[Curve] {
+  def zero = Curve(Nil)
```
sritchie-stripe: looked like we don't have a test yet that exercises the `zero`.

algebird-core/src/main/scala/com/twitter/algebird/BinaryClassificationAUC.scala
```diff
@@ -0,0 +1,110 @@
...
+  override def plus(left: Curve, right: Curve): Curve = {
```
sritchie-stripe: is there a sumOption implementation that would be more efficient that we should add here as well, for adding multiple curves at once?

algebird-core/src/main/scala/com/twitter/algebird/BinaryClassificationAUC.scala
```diff
@@ -0,0 +1,110 @@
...
+ * Note this is for Binary Classifications Tasks
+ *
+ * @param metric Which Metric to compute
+ * @param samples Number of samples, defaults to 100
+ */
+case class BinaryClassificationAUCAggregator(metric: AUCMetric, samples: Int = 100)
+  extends Aggregator[BinaryPrediction, Curve, Double]
+  with Serializable {
+
+  private def linspace(a: Double, b: Double, length: Int = 100): Array[Double] = {
```
sritchie-stripe: we never call this without default args - should we remove the default 100?

algebird-test/src/test/scala/com/twitter/algebird/BinaryClassificationAUCTest.scala
```diff
@@ -0,0 +1,63 @@
...
+
+  implicit val semigroup = CurveMonoid
+  implicit val gen = Arbitrary {
+    for (
+      v <- choose(0, 10000)
+    ) yield Curve(List(ConfusionMatrix(truePositive = v)))
+  }
+
+  property("Curve is associative") {
+    isAssociative[Curve]
```
sritchie-stripe: can we use the monoid laws?

algebird-core/src/main/scala/com/twitter/algebird/BinaryClassificationAUC.scala
```diff
@@ -0,0 +1,110 @@
...
+
+package com.twitter.algebird
+
+/**
+ * Curve is a list of Confusion Matrices with different
+ * thresholds
+ *
+ * @param matrices List of Matrices
+ */
+case class Curve(matrices: List[ConfusionMatrix])
```
sritchie-stripe: if you add an object with an `implicit def monoid: Monoid[Curve] = CurveMonoid` you won't need to declare the same in the tests.

sritchie-stripe:
@richwhitjr this is getting there - I think it's looking good.

One ask that we have for all new data structures is that you add a section with a few examples to the documentation site.

here are the instructions on how to do this: https://github.com/twitter/algebird/blob/develop/CONTRIBUTING.md#contributing-documentation

All data structure examples are compiled by CI, so it's extremely helpful to get a mini tutorial in at the same time the data structure is merged.

here are a few examples of existing pages:
- https://twitter.github.io/algebird/datatypes/approx/exponential_histogram.html
- https://twitter.github.io/algebird/datatypes/averaged_value.html

richwhitjr:
I looked at adding sumOption but couldn't think of a way to make it more efficient without a lot of complexity and not sure if it is entirely worth it.  It would help with object creation but on the other side we would have to aggregate the confusion matrix variables outside into some type of dictionary and reconstruct it at the end.

PR #637, opened 2017-07-26
Create a Bloom Filter from an iterator

johandahlberg:
This PR simply adds the option of creating a Bloom Filter directly from an iterator. I found this useful in my own code, so perhaps someone else will also find it useful.


sritchie:
looks great!

sritchie:
Thanks for this, @johandahlberg, and thanks for the test!

PR #639, opened 2017-09-01
Maven publish shouldn't inline code from dependencies

Lazu13:
Library is built with compiler option ```-opt:l:classpath``` but as it is stated in release note http://www.scala-lang.org/news/2.12.0/#new-optimizer : ```If you are building a library to publish on Maven Central, you should not inline code from dependencies. Users of your library might have different versions of those dependencies on the classpath, which breaks binary compatibility```, it shouldn't when publishing to Maven
Algebird library depends on scala-library and in connection with ```-opt:l:classpath``` flag, it inlines some methods which which uses Scala in version defined in build.sbt, and at the end it might results in binary incompatibility (when using other Scala versions than defined in build.sbt).

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -79,7 +79,7 @@ val sharedSettings = scalariformSettings ++  Seq(
 
   scalacOptions ++= {
     if (scalaVersion.value startsWith "2.12")
-      Seq("-opt:l:classpath")
+      Seq()
     else
       Seq("-optimize")
   },
@@ -251,7 +251,7 @@ lazy val algebirdBijection = module("bijection").settings(
 ).dependsOn(algebirdCore, algebirdTest % "test->test")
 
 lazy val algebirdSpark = module("spark").settings(
-    libraryDependencies += "org.apache.spark" %% "spark-core" % "1.3.0" % "provided",
+    libraryDependencies += "org.apache.spark" % "spark-core_2.10" % "1.3.0" % "provided",
     crossScalaVersions := crossScalaVersions.value.filterNot(_.startsWith("2.12"))
   ).dependsOn(algebirdCore, algebirdTest % "test->test")
 


build.sbt
```diff
@@ -251,7 +251,7 @@ lazy val algebirdBijection = module("bijection").settings(
 ).dependsOn(algebirdCore, algebirdTest % "test->test")
 
 lazy val algebirdSpark = module("spark").settings(
-    libraryDependencies += "org.apache.spark" %% "spark-core" % "1.3.0" % "provided",
+    libraryDependencies += "org.apache.spark" % "spark-core_2.10" % "1.3.0" % "provided",
```
johnynek, line 254: why is this change made?

build.sbt
```diff
@@ -79,7 +79,7 @@ val sharedSettings = scalariformSettings ++  Seq(
 
   scalacOptions ++= {
     if (scalaVersion.value startsWith "2.12")
-      Seq("-opt:l:classpath")
```
johnynek, line 82: there is an option for inlining from the sources. Can we change to that?
How about:
https://gist.github.com/retronym/29932ea76712ff374d1363c9eda6eabe

`-optimize -opt-inline-from:com.twitter.algebird.*` or something, so we can inline from the current project?

build.sbt
Lazu13: I had a problem with ```%%``` which instructs sbt to substitute current scala version to artifact name so I set the version of spark-core by myself, leaving the library version. Checking out the Maven repository, I couldn't find ```spark-core_2.12```.
```
[warn] 	::::::::::::::::::::::::::::::::::::::::::::::
[warn] 	::          UNRESOLVED DEPENDENCIES         ::
[warn] 	::::::::::::::::::::::::::::::::::::::::::::::
[warn] 	:: org.apache.spark#spark-core_2.12;1.3.0: not found
[warn] 	::::::::::::::::::::::::::::::::::::::::::::::
```

However, it could be even 2.11

build.sbt
Lazu13: There is an option ```-opt:l:project``` which seems like the one that can be used, however it requires some research. Doing basic testing, it looks like it doesn't break binary compatibility as it doesn't inline code from dependencies.

johnynek:
There is no spark for 2.12 but we also don't build that module for 2.12 (or don't intend to).

build.sbt
johnynek: We have to leave this as is, or we can't publish this module for 2.10 and 2.11. Note this here:

https://github.com/twitter/algebird/blob/develop/build.sbt#L255

should prevent spark from being set up for 2.12.

johnynek:
indeed just running `compile` does not work for me.

doing:
```
++2.11.9 compile
```
does. I guess the sbt magic is screwed.

build.sbt
lrytz: `-opt:l:project` is also deprecated in 2.12.3, I recommend upgrading to 2.12.3 and using `-opt:l:inline` with `-opt-inline-from:com.twitter.algebird.**`

johnynek:
closing in favor of #643 

plan to release that soon.

PR #641, opened 2017-09-08
Snoble/fix bigdecimal semigroup

snoble:
Fixes bug where there was no Semigroup for BigDecimal. With tests.

### DIFF
--- a/algebird-core/src/main/scala/com/twitter/algebird/Semigroup.scala
+++ b/algebird-core/src/main/scala/com/twitter/algebird/Semigroup.scala
@@ -195,6 +195,7 @@ object Semigroup extends GeneratedSemigroupImplicits with ProductSemigroups with
   implicit def jshortSemigroup: Semigroup[JShort] = JShortRing
   implicit def longSemigroup: Semigroup[Long] = LongRing
   implicit def bigIntSemigroup: Semigroup[BigInt] = BigIntRing
+  implicit def bigDecimalSemigroup: Semigroup[BigDecimal] = implicitly[Ring[BigDecimal]]
   implicit def jlongSemigroup: Semigroup[JLong] = JLongRing
   implicit def floatSemigroup: Semigroup[Float] = FloatRing
   implicit def jfloatSemigroup: Semigroup[JFloat] = JFloatRing
--- a/algebird-test/src/main/scala/com/twitter/algebird/BaseProperties.scala
+++ b/algebird-test/src/main/scala/com/twitter/algebird/BaseProperties.scala
@@ -25,9 +25,15 @@ import scala.math.Equiv
  * Base properties useful for all tests using Algebird's typeclasses.
  */
 object BaseProperties extends MetricProperties {
+  /**
+   * We generate a restricted set of BigDecimals for our tests because if we use
+   * the full range then the way we lose precision in addition does not satisfy
+   * the distributive property perfectly. This means BigDecimal isn't truly
+   * a Ring under it's very strict Equiv.
+   */
   val arbReasonableBigDecimals: Arbitrary[BigDecimal] = Arbitrary(
     for {
-      scale <- Gen.choose(-4, +4)
+      scale <- Gen.choose(-8, +8)
       base <- implicitly[Arbitrary[Int]].arbitrary
     } yield {
       (BigDecimal(base) * BigDecimal(10).pow(scale))
--- a/algebird-test/src/main/scala/com/twitter/algebird/BaseProperties.scala
+++ b/algebird-test/src/main/scala/com/twitter/algebird/BaseProperties.scala
@@ -33,7 +33,7 @@ object BaseProperties extends MetricProperties {
    */
   val arbReasonableBigDecimals: Arbitrary[BigDecimal] = Arbitrary(
     for {
-      scale <- Gen.choose(-8, +8)
+      scale <- Gen.choose(-7, +7)
       base <- implicitly[Arbitrary[Int]].arbitrary
     } yield {
       (BigDecimal(base) * BigDecimal(10).pow(scale))
--- a/algebird-test/src/main/scala/com/twitter/algebird/BaseProperties.scala
+++ b/algebird-test/src/main/scala/com/twitter/algebird/BaseProperties.scala
@@ -27,8 +27,8 @@ import scala.math.Equiv
 object BaseProperties extends MetricProperties {
   val arbReasonableBigDecimals: Arbitrary[BigDecimal] = Arbitrary(
     for {
-      scale <- Gen.choose(-128, +128)
-      base <- implicitly[Arbitrary[BigInt]].arbitrary
+      scale <- Gen.choose(-4, +4)
+      base <- implicitly[Arbitrary[Int]].arbitrary
     } yield {
       (BigDecimal(base) * BigDecimal(10).pow(scale))
     })
--- a/algebird-test/src/test/scala/com/twitter/algebird/NumericAlgebraTests.scala
+++ b/algebird-test/src/test/scala/com/twitter/algebird/NumericAlgebraTests.scala
@@ -1,0 +1,29 @@
+package com.twitter.algebird
+
+import org.scalatest._
+
+import org.scalatest.prop.PropertyChecks
+import org.scalacheck.{ Gen, Arbitrary, Prop, Properties }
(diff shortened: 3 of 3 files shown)


algebird-core/src/main/scala/com/twitter/algebird/Semigroup.scala
```diff
@@ -195,6 +195,7 @@ object Semigroup extends GeneratedSemigroupImplicits with ProductSemigroups with
   implicit def jshortSemigroup: Semigroup[JShort] = JShortRing
   implicit def longSemigroup: Semigroup[Long] = LongRing
   implicit def bigIntSemigroup: Semigroup[BigInt] = BigIntRing
+  implicit def bigDecimalSemigroup: Semigroup[BigDecimal] = implicitly[Ring[BigDecimal]]
```
oscar-stripe, line 198: can we add this for `Monoid` and `Group` too?

algebird-test/src/test/scala/com/twitter/algebird/BatchedTest.scala
```diff
@@ -52,6 +52,30 @@ class BatchedLaws extends CheckProperties {
   testBatchedMonoid[String]("String", 10)
   testBatchedMonoid[String]("String", 100)
   testBatchedMonoid[String]("String", 1000000)
+
+  def testBatchedSemigroup[A: Arbitrary: Semigroup](name: String, size: Int): Unit = {
+    implicit val m: Semigroup[Batched[A]] = Batched.compactingSemigroup[A](size)
+    property(s"CountMinSketch[$name] batched at $size is a Semigroup") {
```
oscar-stripe: what is countminsketch got to do with anything here?

algebird-test/src/test/scala/com/twitter/algebird/BatchedTest.scala
```diff
@@ -52,6 +52,30 @@ class BatchedLaws extends CheckProperties {
   testBatchedMonoid[String]("String", 10)
   testBatchedMonoid[String]("String", 100)
   testBatchedMonoid[String]("String", 1000000)
+
+  def testBatchedSemigroup[A: Arbitrary: Semigroup](name: String, size: Int): Unit = {
```
oscar-stripe: why was this added here. Seems a weird way to test bigdecimal?

Can we instead add a file like `NumericAlgebraTests.scala` that has something like:
```scala
class NumericAlgebraTests extends CheckProperties {
  property("test bigdecimal") {
    // check resolution:
    implicitly[Semigroup[BigDecimal]]
    implicitly[Monoid[BigDecimal]]
    implicitly[Group[BigDecimal]]
    // now test all the laws:
    ringLaws[BigDecimal]
  }
```
and similar for the numeric types we care about?

algebird-core/src/main/scala/com/twitter/algebird/Semigroup.scala
snoble: they're already there... this was the one gap

algebird-test/src/test/scala/com/twitter/algebird/BatchedTest.scala
snoble: yeah... I wanted to add _something_ as a test... even if it was crude... I'll add this instead

algebird-test/src/test/scala/com/twitter/algebird/BatchedTest.scala
snoble: copy-pasta... I didn't read at all but it comes directly from `def testBatchedMonoid`

algebird-test/src/test/scala/com/twitter/algebird/NumericAlgebraTests.scala
```diff
@@ -0,0 +1,23 @@
...
+      // check resolution:
+      implicitly[Semigroup[BigDecimal]]
+      implicitly[Monoid[BigDecimal]]
+      implicitly[Group[BigDecimal]]
+      // now test all the laws:
+      ringLaws[A]
+    }
+
+  testRingAndBelowProperties[Int]("int")
+  testRingAndBelowProperties[BigDecimal]("bigdecimal")
```
snoble: This one keeps failing for me with errors like
```
[info] NumericAlgebraTests:
[info] - test int
[info] - test bigdecimal *** FAILED ***
[info]       arg0 = 1.627209798210513,
[info]       arg1 = 606303697535274687,
[info]       arg2 = -0.003402952952309675094511817686947572
[info]     )
[info]     Label of failing property:
[info]       isDistributiveDifferentTypes
```

johnynek: it may not be distributive

did you check?

Float and Double are not.

snoble: when I test with
```
scala> def foo[T](a: T, b: T, c: T)(implicit rng: Numeric[T]) = {
     |     Equiv[T].equiv(rng.times(a, rng.plus(b, c)), rng.plus(rng.times(a, b), rng.times(a, c))) &&
     |     Equiv[T].equiv(rng.times(rng.plus(b, c), a), rng.plus(rng.times(b, a), rng.times(c, a)))
     | }
```
it seems to work with the failing values as BigDecimal

snoble: ahhh... it's rounding errors
```
scala> val a = BigDecimal(1.627209798210513)
a: scala.math.BigDecimal = 1.627209798210513

scala> val b = BigDecimal("606303697535274687")
b: scala.math.BigDecimal = 606303697535274687

scala> val c = BigDecimal("-0.003402952952309675094511817686947572")
c: scala.math.BigDecimal = -0.003402952952309675094511817686947572
scala> def bar[T](a: T, b: T, c: T)(implicit rng: Numeric[T]) = {
     | (rng.times(a, rng.plus(b, c)), rng.plus(rng.times(a, b), rng.times(a, c)), rng.times(rng.plus(b, c), a), rng.plus(rng.times(b, a), rng.times(c, a)))
     | }

scala> bar(a,b,c)
res9: (scala.math.BigDecimal, scala.math.BigDecimal, scala.math.BigDecimal, scala.math.BigDecimal) = (986583317320662231.5814891877975833,986583317320662231.581489187797583304121164227599838248,986583317320662231.5814891877975833,986583317320662231.581489187797583304121164227599838248)
```

How are those situations dealt with?

johnynek: you can set up a custom `Equiv` instance:

```scala
implicit val closeEnough: Equiv[BigDecimal] = new Equiv[BigDecimal] {
  val eps = BigDecimal(1e-10) // or something.
  def equiv(a: BigDecimal, b: BigDecimal) = 2 * (b - a).abs / (b.abs + a.abs) < eps
}
```

algebird-test/src/main/scala/com/twitter/algebird/BaseProperties.scala
```diff
@@ -27,8 +27,8 @@ import scala.math.Equiv
 object BaseProperties extends MetricProperties {
   val arbReasonableBigDecimals: Arbitrary[BigDecimal] = Arbitrary(
     for {
-      scale <- Gen.choose(-128, +128)
-      base <- implicitly[Arbitrary[BigInt]].arbitrary
+      scale <- Gen.choose(-4, +4)
+      base <- implicitly[Arbitrary[Int]].arbitrary
```
snoble: @johnynek: This may not be a good idea. I can go into it but when you add two BigDecimals that have sufficiently different scales it is possible to lose precision. And when testing out the distributive properties of rings the way in which the precision is lost is not distributive. I suspect nobody is surprised by this though so it's reasonable to limit the tests to cases where precision can't be lost.

johnynek: This is fine, but can we do the biggest scale we can do with the tests passing consistently? Would 24 work for instance?

snoble:
@johnynek: `(-8,+8)` was as far as I could get and consistently pass

snoble:
turns out I didn't try hard enough to break it :/

johnynek:
👍

Issue #640, opened 2017-09-05
Using the inliner (-opt:l:classpath) causes binary incompatible releases


lrytz:
Algebird 0.13.0 for Scala 2.12 (built with Scala 2.12.1) is not binary compatible with Scala 2.12.3, which causes a `NoSuchMethodError` at run-time. This is because inlining from the classpath is enabled in the build (https://github.com/twitter/algebird/blob/0.13.0/build.sbt#L80-L85).

From the [2.12.0 release notes](http://www.scala-lang.org/news/2.12.0/):

> When inlining code from the classpath, you must ensure that all dependencies have exactly the same versions at compile time and run time.
> If you are building a library to publish on Maven Central, you should not inline code from dependencies. Users of your library might have different versions of those dependencies on the classpath, which breaks binary compatibility.

Note that 2.12.3 deprecates `-opt:l:classpath`, introduces `-opt:l:inline`, and a new setting `-opt-inline-from` which tells the inliner where from which packages it is allowed to inline from. See https://github.com/scala/scala/pull/5964.

lrytz:
Details on the binary incompatibility here: https://github.com/scala/bug/issues/10489#issuecomment-327097429

johnynek:
we have an open PR on this now:

https://github.com/twitter/algebird/pull/639

PR #643, opened 2017-09-13
Fix some issues uncovered by 2.12.3

johnynek:
this addresses #640 cc @lrytz 

review @piyushnarang @ianoc ?

### DIFF
--- a/algebird-test/src/test/scala/com/twitter/algebird/CountMinSketchTest.scala
+++ b/algebird-test/src/test/scala/com/twitter/algebird/CountMinSketchTest.scala
@@ -930,7 +930,7 @@ class CMSFunctionsSpec extends PropSpec with PropertyChecks with Matchers {
   }
 
   property("throw IAE when deriving depth from invalid delta values") {
-    val invalidDeltas = Table("invalidDelta", 0.0, 1E-330, 1E-400)
+    val invalidDeltas = Table("invalidDelta", 0.0, 1.0, 2.0, -1.0)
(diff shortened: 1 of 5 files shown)


algebird-test/src/test/scala/com/twitter/algebird/CountMinSketchTest.scala
```diff
@@ -930,7 +930,7 @@ class CMSFunctionsSpec extends PropSpec with PropertyChecks with Matchers {
   }
 
   property("throw IAE when deriving depth from invalid delta values") {
-    val invalidDeltas = Table("invalidDelta", 0.0, 1E-330, 1E-400)
```
johnynek, line 933: these numbers were invalid doubles and 2.12.3 won't parse them. scala 2.11 I think parses them to 0.0.

build.sbt
```diff
@@ -79,7 +79,7 @@ val sharedSettings = scalariformSettings ++  Seq(
 
   scalacOptions ++= {
     if (scalaVersion.value startsWith "2.12")
-      Seq("-opt:l:classpath")
+      Seq("-opt-inline-from:com.twitter.algebird.**")
```
johnynek, line 82: make sure we are only inlining from the local project. I *think* this is the correct invocation.

ianoc:
lg

build.sbt
lrytz: you **also** need the `-opt:l:inline` flag, `-opt-inline-from` alone doesn't enable the optimizer, it just configures it.

PR #647, opened 2017-10-03
Make derivation macros work for generic case classes

travisbrown:
`returnType` doesn't work appropriately if the member type is generic.


johnynek:
I fracking hate sbt.

something in the build is somehow now wrong. The 2.12 build is looking for a non-existent 2.11 version of algebird.

I would love to get bids from people to see what I would have to pay to never have to worry about `sbt +++` vs `sbt ++` and get sane cross building support. Also, while this sbt expert is here, can they look at an scalajs build too.

johnynek:
going to merge assuming #648 fixes the build failure.

Thanks for fixing this!

johnynek:
this is published in 0.13.3: https://github.com/twitter/algebird/releases/tag/v0.13.3

PR #648, opened 2017-10-04
Upgrade to sbt 1.0.2

gkossakowski:
Removes sbt-doge. This plugin is now folded into sbt master.
Upgrade a bunch of plugins to versions compatible with sbt 1.0. Versions
plucked from https://github.com/sbt/sbt/wiki/sbt-1.x-plugin-migration


gkossakowski:
This PR touches plugins that I either never used or used a little so it would be good to check if they still work as expected:

- [ ] sbt-unidoc
- [ ] sbt-microsites
- [ ] sbt-scalariform
- [ ] sbt-ghpages

johnynek:
we want to remove scalariform, so we can ditch that. The others are documentation stuff @sritchie set up. I actually don't know how to publish the docs, so maybe he can chime in and we can get some documentation on what the release process is.

PR #649, opened 2018-01-26
Add trailing window aggregator

cdg-stripe:
#### Description

Add WindowMonoid utility for aggregating over finite window. Window case class stores elements with a queue. The monoid combines windows while trimming their elements to the appropriate size. The monoid is also geared to aggregate elements with the group property more efficiently, as we can subtract from the total elements we push out of the queue upon aggregation.

#### Testing

Unit tests

r? @johnynek

### DIFF
--- a/algebird-core/src/main/scala/com/twitter/algebird/Window.scala
+++ b/algebird-core/src/main/scala/com/twitter/algebird/Window.scala
@@ -17,6 +17,7 @@ limitations under the License.
 package com.twitter.algebird
 
 import scala.collection.immutable.Queue
+import Operators._
 
 /**
  *
--- a/algebird-core/src/main/scala/com/twitter/algebird/Window.scala
+++ b/algebird-core/src/main/scala/com/twitter/algebird/Window.scala
@@ -112,18 +112,15 @@ case class WindowMonoid[T](
 
   def plusM(a: Window[T], b: Window[T])(implicit m: Monoid[T]): Window[T] =
     if (b.items.size >= windowSize) {
-      var q = b.items
-      while (q.size > windowSize) {
-        q = q.tail
-      }
-      val total = m.sum(q)
-      Window(total, q)
+      val items = b.items.takeRight(windowSize)
+      val total = m.sum(items)
+      Window(total, items)
     } else {
       // we need windowSize - b.items.size from `a`
       val fromA = a.items.takeRight(windowSize - b.items.size)
-      val res = fromA ++ b.items
+      val items = fromA ++ b.items
       val total = m.sum(fromA) + b.total
-      Window(total, res)
+      Window(total, items)
     }
 
     def fromTraversable(ts: Traversable[T]): Window[T] = {
--- a/algebird-core/src/main/scala/com/twitter/algebird/Window.scala
+++ b/algebird-core/src/main/scala/com/twitter/algebird/Window.scala
@@ -76,17 +76,11 @@ case class WindowMonoid[T](
     windowSize: Int
 )(implicit p: Priority[Group[T], Monoid[T]])
     extends Monoid[Window[T]] {
-  val zero =
-    p match {
-      case Priority.Preferred(g) => Window(g.zero)
-      case Priority.Fallback(m)  => Window(m.zero)
-    }
+
+  def zero = p.fold(g => Window(g.zero))(m => Window(m.zero))
 
   def plus(a: Window[T], b: Window[T]): Window[T] =
-    p match {
-      case Priority.Preferred(g) => plusG(a, b)(g)
-      case Priority.Fallback(m)  => plusM(a, b)(m)
-    }
+    p.fold(g => plusG(a, b)(g))(m => plusM(a, b)(m))
 
   def plusG(a: Window[T], b: Window[T])(implicit g: Group[T]): Window[T] =
     if (b.items.size >= windowSize) {
--- a/algebird-core/src/main/scala/com/twitter/algebird/Window.scala
+++ b/algebird-core/src/main/scala/com/twitter/algebird/Window.scala
@@ -77,6 +77,8 @@ case class WindowMonoid[T](
 )(implicit p: Priority[Group[T], Monoid[T]])
     extends Monoid[Window[T]] {
 
+  require(windowSize >= 1, "Windows must have positive sizes")
+
   def zero = p.fold(g => Window(g.zero))(m => Window(m.zero))
 
   def plus(a: Window[T], b: Window[T]): Window[T] =
--- a/algebird-core/src/main/scala/com/twitter/algebird/Window.scala
+++ b/algebird-core/src/main/scala/com/twitter/algebird/Window.scala
@@ -62,7 +62,7 @@ case class Window[T](total: T, items: Queue[T]) {
 
 object Window {
   def apply[T](v: T): Window[T] = Window[T](v, Queue[T](v))
-  def from[T](ts: Traversable[T])(implicit m: WindowMonoid[T]) = m.fromTraversable(ts)
+  def from[T](ts: Iterable[T])(implicit m: WindowMonoid[T]) = m.fromIterable(ts)
 }
 
 /**
@@ -124,22 +124,22 @@ case class WindowMonoid[T](
       Window(total, items)
     }
 
-  def fromTraversable(ts: Traversable[T]): Window[T] = {
+  override def sumOption(ws: TraversableOnce[Window[T]]): Option[Window[T]] =
+    if (ws.isEmpty) None
+    else {
+      val it = ws.toIterator
+      var queue = Queue.empty[T]
+      while (it.hasNext) {
+        queue = (queue ++ it.next.items).takeRight(windowSize)
+      }
+      val monT: Monoid[T] = p.join
+      Some(Window(monT.sum(queue), queue))
+    }
+
+  def fromIterable(ts: Iterable[T]): Window[T] = {
     val monT: Monoid[T] = p.join
     val right = ts.toList.takeRight(windowSize)
     val total = monT.sum(right)
     Window(total, Queue(right: _*))
-  }
-
-  override def sumOption(ws: TraversableOnce[Window[T]]): Option[Window[T]] = {
-    if(ws.isEmpty) None
-    else {
-      val it = ws.toIterator
-      var queue = Queue[T]()
-      while (it.hasNext) {
-        queue = (queue ++ it.next.items).takeRight(windowSize)
-      }
-      Some(fromTraversable(queue))
-    }
   }
 }
--- a/algebird-core/src/main/scala/com/twitter/algebird/Window.scala
+++ b/algebird-core/src/main/scala/com/twitter/algebird/Window.scala
@@ -98,6 +98,18 @@ case class WindowMonoid[T](
       val total = monT.sum(right)
       Window(total, Queue(right: _*))
     }
+
+    override def sumOption(ws: TraversableOnce[Window[T]]): Option[Window[T]] = {
+      if(ws.isEmpty) None
+      else {
+        val it = ws.toIterator
+        var queue = Queue[T]()
+        while (it.hasNext) {
+          queue = (queue ++ it.next.items).takeRight(windowSize)
+        }
+        Some(fromTraversable(queue))
+      }
+    }
 }
 
 /*
--- a/algebird-core/src/main/scala/com/twitter/algebird/Window.scala
+++ b/algebird-core/src/main/scala/com/twitter/algebird/Window.scala
@@ -25,6 +25,34 @@ import scala.collection.immutable.Queue
  *
  * @param total Known running total of `T`
  * @param items queue of known trailing elements.
+ *
+ *  Example usage:
+ *
+ *  case class W28[T](window: Window[T]) {
+ *    def total = this.window.total
+ *    def items = this.window.items
+ *    def size = this.window.size
+ *  }
+ *
+ *  object W28 {
+ *    val windowSize = 28
+ *    def apply[T](v: T): W28[T] = W28[T](Window(v))
+ *
+ *    implicit def w28Monoid[T](implicit p: Priority[Group[T], Monoid[T]]): Monoid[W28[T]] =
+ *      new Monoid[W28[T]] {
+ *        private val WT: Monoid[Window[T]] = WindowMonoid[T](windowSize)
+ *        def zero = W28[T](WT.zero)
+ *        def plus(a: W28[T], b: W28[T]): W28[T] =
+ *          W28[T](WT.plus(a.window, b.window))
+ *      }
+ *  }
+ *  val elements = getElements()
+ *
+ *  val trailing90Totals =
+ *    elements
+ *      .map{ W90 }
+ *      .scanLeft(W90(0)) { (a, b) => a + b }
+ *      .map{ _.total }
  */
 
 case class Window[T](total: T, items: Queue[T]) {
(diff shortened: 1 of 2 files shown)


algebird-core/src/main/scala/com/twitter/algebird/Window.scala
```diff
@@ -0,0 +1,118 @@
...
+/**
+ * Provides a natural monoid for combining windows truncated to some window size.
+ *
+ * @param windowSize
+ */
+
+case class WindowMonoid[T](
+    windowSize: Int
+)(
+    implicit m: Monoid[T],
```
johnynek: Let’s not use this m. You can get the zero from the p by case matching.

johnynek (commented):
This is great.

Look in algebird-core/src/test/... to find the tests for this module. We want to make sure it passes the monoidLaws. You can search for monoidLaw for some examples.

After we add test coverage we should merge this.

algebird-core/src/main/scala/com/twitter/algebird/Window.scala
```diff
@@ -0,0 +1,118 @@
...
+/**
+ *
+ * Convenience case class defined with a monoid for aggregating elements over
+ * a finite window.
+ *
+ * @param total wrapped instance of `T`
+ * @param q (optional) queue to override default q containing input "total"
+ */
+
+case class Window[T](total: T, q: Queue[T] = Queue()) {
```
MansurAshraf: This looks great! What do you think about allowing user to pass a `PriorityQueue` instead of just a Queue (AFAIK scala PQ doesnt extend Queue). The way code is written right now, its implying that I need a commutative monoid to guarantee deterministic behavior or I need to control the order in  which items are inserted in the window. When I work with windows they usually have  some sort of ordering defined (last 3 distinct IPs used by a user in last one hour for example), It would be nice to allow users to  preserve the ordering somehow IMO

algebird-core/src/main/scala/com/twitter/algebird/Window.scala
```diff
@@ -0,0 +1,118 @@
...
+    p match {
+      case Priority.Preferred(g) => plusG(a, b)(g)
+      case Priority.Fallback(m)  => plusM(a, b)(m)
+    }
+
+  def plusG(a: Window[T], b: Window[T])(implicit g: Group[T]): Window[T] =
+    if (b.items.size >= windowSize) {
+      var total: T = b.total
+      var q = b.items
+      while (q.size > windowSize) {
```
MansurAshraf, line 91: Not sure if its possible to do it in this implementation but it would be nice to generalize how elements are evicted from the window. Right now it seems like I can only do fixed windows but I would really like to do time based rolling windows as well. For example: Rolling window of 1 hour and we evict element based on time instead of size

algebird-core/src/main/scala/com/twitter/algebird/Window.scala
johnynek: actually, I think `Window[T](total: T, items: Queue[T])` is better. We don't really want people to do something like `Window(100)` do we? that will have no items in the queue, and I imagine break the laws.

johnynek (commented):
in addition to the monoidLaws, I think we want something like:
```scala
forAll { (ts0: List[Int], n: Int) =>
  val ts = ts0.takeRight(n)
  val mon = new WindowMonoid(n)
  assert(mon.sum(ts.map(Window(_))).total == ts.sum)
}
```
That's the behavior you want, right?

algebird-core/src/main/scala/com/twitter/algebird/Window.scala
cdg-stripe: If you look at `items`, it fills in the queue if none was provided, i.e meant for initialization. The intent is to have a simple interface where you can wrap a window around a single object and sum in a sensible way:
```
implicit def wm[T] = WindowMonoid[T](2)

Window(1) + Window(2) + Window(3) == Window(5)
```
Needing to specify the queue would make this redundant 
```
Window(1, Queue(1)) + ...
```

Tbh I'm not a big fan of even having to expose the queue, since you could change the implementation in the future (e.g. a PriorityQueue). A more general approach might use a `PushPopper` interface, but that's probably overkill.

algebird-core/src/main/scala/com/twitter/algebird/Window.scala
cdg-stripe: My thought was to eventually have two classes/monoids, based on whether we want aggregations over fixed items or fixed time.

```
// Monoid defines aggregation for fixed window length, e.g. last 28 elements. Queue fixed to 28.
case class Window[T](total:T, q: Queue[T])

// Monoid defines aggregation for fixed time, e.g. last 28 days. Queue can grow arbitrarily.
case class WindowWithTs(lastestTime: Double, total: T, q: Queue[(Double, T)])
```

Does that seem sensible?

MansurAshraf: That seems reasonable but I am still not sure why we cant use `ProrityQueue` instead of `Queue`? As a user I would like to control the ordering and if I am passing you a PQ, I can do that by specifying my own comparator

cdg-stripe:
@MansurAshraf: I'm trying to grasp whether we need a fully general implementation up front, where the user inputs their own queue/priority. As mentioned above my thoughts are to have trailing windows based on one of:

- Fixed number of elements
- Fixed time interval (not implemented here, but can in a separate PR)

The "last N ip-addresses" problem is interesting. But rather than making this implementation more flexible/abstract, it might make sense to have a `LastNDistinct` monoid. I suspect this approach will be easier to test and implement.

cdg-stripe:
@johnynek:

Added tests and removed the redundant `implicit m: Monoid[T]`. I also tried out making `Window` a `class` instead of a `case class`. 
```
class Window[T](t: T, q: Queue[T] = Queue()) {
  def total = t
  def items = if (q.size > 0) q else Queue(total)
  def size = items.size
}
```
This makes it clear that q is optional and never accessible after instantiation. If no q is given, items default `Queue(total)`, the expected behavior when instantiating a window from a single item.

algebird-core/src/main/scala/com/twitter/algebird/Window.scala
```diff
@@ -0,0 +1,120 @@
...
+      // we need windowSize - b.items.size from `a`
+      val fromA = a.items.takeRight(windowSize - b.items.size)
+      val res = fromA ++ b.items
+      val total = m.sum(fromA) + b.total
+      Window(total, res)
+    }
+}
+
+/*
+  Example usage:
```
johnynek: can you make this a scaladoc comment on the case class?

algebird-test/src/test/scala/com/twitter/algebird/WindowLawsTest.scala
```diff
@@ -0,0 +1,20 @@
...
+import org.scalacheck.Prop.forAll
+
+
+class WindowLaws extends CheckProperties {
+  property("We aggregate over only n items") {
+    forAll { (ts0: List[Int], n: Int) =>
+      val ts = ts0.takeRight(n)
+      val mon = new WindowMonoid(n)
+      assert(mon.sum(ts0.map( Window(_) )).total == ts.sum)
+    }
```
johnynek: can we have a second one:
```scala
forAll { (ts0: List[Int], ts1: List[Int], n: Int) =>
  val expected = Queue((ts0 ::: ts1).takeRight(n): _*)
  val mon = new WindowMonoid(n)
  val got = mon.plus(Window(ts0.sum, Queue(ts0: _*)), Window(ts1.sum, Queue(ts1:_*)))
  assert(got == expected)
}
```

algebird-core/src/main/scala/com/twitter/algebird/Window.scala
```diff
@@ -0,0 +1,120 @@
...
+}
+
+object Window {
+  def apply[T](v: T): Window[T] = Window[T](v, Queue[T](v))
+}
+
+/**
+ * Provides a natural monoid for combining windows truncated to some window size.
+ *
+ * @param windowSize
```
johnynek: can you add more comment to this javaDoc param?

algebird-core/src/main/scala/com/twitter/algebird/Window.scala
```diff
@@ -0,0 +1,120 @@
...
+case class WindowMonoid[T](
+    windowSize: Int
+)(implicit p: Priority[Group[T], Monoid[T]])
+    extends Monoid[Window[T]] {
+  val zero =
+    p match {
+      case Priority.Preferred(g) => Window(g.zero)
+      case Priority.Fallback(m)  => Window(m.zero)
+    }
+
```
johnynek: what about a
```scala
  def fromIterable[T](ts: Iterable[T]): Window[T] = {
    val monT: Monoid[T] = p.join
    val right = ts.toList.takeRight(windowSize)
    val total = monT.sum(right)
    Window(total, Queue.empty[T] ++ right)
  }
```

algebird-core/src/main/scala/com/twitter/algebird/Window.scala
```diff
@@ -0,0 +1,120 @@
...
+  def plus(a: Window[T], b: Window[T]): Window[T] =
+    p match {
+      case Priority.Preferred(g) => plusG(a, b)(g)
+      case Priority.Fallback(m)  => plusM(a, b)(m)
+    }
+
```
johnynek: I think you can optimize `sumOption` somewhat here: you can take blocks of data and ignore the left if the right is full. PS: scalding uses this if it exists, so overridding sumOption on Semigroup (or any subclass) can be a very nice performance win on map/reduce.

```scala
def sumOption(ws: TraversableOnce[Window[T]]): Option[Window[T]] =
  if (ws.isEmpty) None
  else {
    val it = ws.toIterator
    var size = 0
    var queue = Queue.empty[Window[T]]
    while (it.hasNext) {
      val n = it.next
      queue = queue :+ n
      size += n.size
      val tailSize = size - queue.head.size
      if (tailSize >= windowSize) {
        queue = queue.tail
        size = tailSize
      }
    }
    // now we only have to merge the queue:
    queue.tail.foldLeft(queue.head)(plus(_, _))
  }
```
I think this can be a pretty big win.

johnynek (changes requested):

johnynek:
this PR: https://github.com/twitter/algebird/pull/650 seems to fix the ruby issue on travis.

johnynek:
(nothing like non-hermetic builds to make you feel real good, all day long).

algebird-core/src/main/scala/com/twitter/algebird/Window.scala
cdg-stripe: @johnynek: Overwrote sumOption below.

algebird-test/src/test/scala/com/twitter/algebird/WindowLawsTest.scala
cdg-stripe: Done

algebird-core/src/main/scala/com/twitter/algebird/Window.scala
cdg-stripe: @johnynek: I did this, and generalized to `Traversable`.

algebird-core/src/main/scala/com/twitter/algebird/Window.scala
```diff
@@ -0,0 +1,145 @@
...
+    val right = ts.toList.takeRight(windowSize)
+    val total = monT.sum(right)
+    Window(total, Queue(right: _*))
+  }
+
+  override def sumOption(ws: TraversableOnce[Window[T]]): Option[Window[T]] = {
+    if(ws.isEmpty) None
+    else {
+      val it = ws.toIterator
+      var queue = Queue[T]()
```
johnynek: .empty is more efficient since you avoid allocating the varargs wrapper.

algebird-core/src/main/scala/com/twitter/algebird/Window.scala
```diff
@@ -0,0 +1,145 @@
...
+      while (it.hasNext) {
+        queue = (queue ++ it.next.items).takeRight(windowSize)
+      }
+      Some(fromTraversable(queue))
```
johnynek: Using fromTraversable here is ignoring that you already have a queue. I think summing the queue and reusing it will be a significant performance improvement

algebird-core/src/main/scala/com/twitter/algebird/Window.scala
```diff
@@ -0,0 +1,145 @@
...
+ *      .map{ _.total }
+ */
+
+case class Window[T](total: T, items: Queue[T]) {
+  def size = this.items.size
+}
+
+object Window {
+  def apply[T](v: T): Window[T] = Window[T](v, Queue[T](v))
+  def from[T](ts: Traversable[T])(implicit m: WindowMonoid[T]) = m.fromTraversable(ts)
```
johnynek: Can we use Iterable? Scala 2.13 is changing the collections to avoid Traversable: https://www.scala-lang.org/blog/2017/02/28/collections-rework.html#traversable-and-iterable

cdg-stripe:
@johnynek:

Added more tests. Caught that the zero element should have an empty queue, where originally it was `Queue[T](T.zero)`. Also made the monoid on group elements more efficient. I dug around to see if there's anything else worth tweaking, but I think it's in a decent spot.

Also what are you thoughts about Scala 2.13, when we can have a signature `Window[Size, T]`? Might be okay have `Window[T]` and `WindowN[Size, T]` case classes.

johnynek:
yeah, I think in scala 2.13 we will have to revisit a lot of stuff, but the compatibility story is open since we will need to support 2.12 for a while, I guess (we aren't even on 2.12 yet at Stripe).

algebird-core/src/main/scala/com/twitter/algebird/Window.scala
```diff
@@ -0,0 +1,144 @@
...
+
+
+case class WindowMonoid[T](
+    windowSize: Int
+)(implicit p: Priority[Group[T], Monoid[T]])
+    extends Monoid[Window[T]] {
+
+  require(windowSize >= 1, "Windows must have positive sizes")
+
+  def zero = p.fold(g => Window[T](g.zero, Queue.empty[T]))(m => Window(m.zero, Queue.empty[T]))
```
johnynek, line 82: can we make this a `val` so we don't have to do the computation and allocation again on each call?

algebird-core/src/main/scala/com/twitter/algebird/Window.scala
```diff
@@ -0,0 +1,144 @@
...
+
+  def plus(a: Window[T], b: Window[T]): Window[T] =
+    p.fold(g => plusG(a, b)(g))(m => plusM(a, b)(m))
```
johnynek, line 85: for performance (not calling fold on each plus call) we are probably better to have two subclasses of `WindowMonoid` which could be an abstract class. We can decide once when we allocate which path to choose. Then the jit should be able to inline both instances. Sadly, the jvm can't optimize the above very well.

johnynek (commented):
Two comments, but let's address them in a follow up PR.

Thanks for this!

johnynek (approved):

johnynek:
published in 0.13.4:

https://github.com/twitter/algebird/releases/tag/v0.13.4

johnynek:
@MansurAshraf note if you want the most recent K things, TopKMonoid should work:

https://github.com/twitter/algebird/blob/develop/algebird-core/src/main/scala/com/twitter/algebird/TopKMonoid.scala

(so would priorityqueue I guess).

PR #651, opened 2018-02-13
Correct Min.semigroup

kellen:
`Min.semigroup` was incorrectly returning the *greater* element.

This PR corrects the semigroup and adds tests to check the results from both `Min.semigroup` and `Max.semigroup` (which was already correct).

### DIFF
--- a/algebird-test/src/test/scala/com/twitter/algebird/MaxLaws.scala
+++ b/algebird-test/src/test/scala/com/twitter/algebird/MaxLaws.scala
@@ -29,6 +29,13 @@ class MaxLaws extends CheckProperties {
     }
   }
 
+  property("Max.semigroup returns the maximum item") {
+    forAll { v: NonEmptyVector[Int] =>
+      val maxItems = v.items.map { Max(_) }
+      v.sorted.last == Max.semigroup[Int].combineAllOption(maxItems).get.get
+    }
+  }
+
   property("Max[Long] is a commutative monoid") {
     commutativeMonoidLaws[Max[Long]]
   }
--- a/algebird-test/src/test/scala/com/twitter/algebird/MaxLaws.scala
+++ b/algebird-test/src/test/scala/com/twitter/algebird/MaxLaws.scala
@@ -13,6 +13,12 @@ class MaxLaws extends CheckProperties {
       l + r == realMax && (l max r) == realMax
     }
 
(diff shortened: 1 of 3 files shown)


kellen:
Hmm, build failure seems unrelated to the PR changes.

johnynek:
Amazing. It looks like the monoid was correct and since all the numeric types use that we were saved.

algebird-test/src/test/scala/com/twitter/algebird/MaxLaws.scala
```diff
@@ -29,6 +29,13 @@ class MaxLaws extends CheckProperties {
     }
   }
 
+  property("Max.semigroup returns the maximum item") {
+    forAll { v: NonEmptyVector[Int] =>
+      val maxItems = v.items.map { Max(_) }
+      v.sorted.last == Max.semigroup[Int].combineAllOption(maxItems).get.get
```
johnynek: Can you also test this for a custom type that does not have a special instance defined?

Also, scala has .max on Iterable. Can we just use that rather than sorted.last?

algebird-test/src/test/scala/com/twitter/algebird/MinLaws.scala
```diff
@@ -26,6 +26,13 @@ class MinLaws extends CheckProperties {
     }
   }
 
+  property("Min.semigroup returns the minimum item") {
+    forAll { v: NonEmptyVector[Int] =>
+      val minItems = v.items.map { Min(_) }
+      v.sorted.head == Min.semigroup[Int].combineAllOption(minItems).get.get
+    }
```
johnynek: Same comments as above.

algebird-test/src/test/scala/com/twitter/algebird/MaxLaws.scala
kellen: Done.

johnynek:
👍 

will merge when green.

johnynek:
Thank you for taking the time to send a PR here.

oscar-stripe:
This was broken in 0.13.2 by me. :( I'm sorry y'all...

https://github.com/twitter/algebird/pull/620

I just hit this in tests trying to update to the latest algebird. At least our tests caught it....

Will publish 0.13.4 presently.

johnynek:
published in https://github.com/twitter/algebird/releases/tag/v0.13.4

PR #652, opened 2018-02-22
Optimize Window monoid a bit

johnynek:
Rather than switching inside the monoid, switch on creation so the JIT can optimize better.

### DIFF
--- a/algebird-core/src/main/scala/com/twitter/algebird/Window.scala
+++ b/algebird-core/src/main/scala/com/twitter/algebird/Window.scala
@@ -17,6 +17,7 @@ limitations under the License.
 package com.twitter.algebird
 
 import scala.collection.immutable.Queue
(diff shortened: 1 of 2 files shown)


algebird-core/src/main/scala/com/twitter/algebird/Window.scala
```diff
@@ -128,17 +143,41 @@ case class WindowMonoid[T](
...
+      while (it.hasNext) {
+        // avoid materializing the whole list in memory
+        // at one time
+        queue = queue :+ it.next
+        size = size + 1
+        if (size > windowSize) {
+          queue = queue.tail
+          size = size - 1
+        }
+      }
```
non, line 164: It's probably possible to geek out on this even further, but this looks good to me for now.

PR #655, opened 2018-05-11
Add algebird-generic module to work with shapeless

johnynek:
This is a long overdue algbird shapeless module.

For now, it gives instances for semigroup, monoid, group and ring, but it could also give some nice combinators on Aggregator, which I will send in a follow up.

### DIFF
--- a/algebird-shapeless/src/main/scala/com/twitter/algebird/shapeless/EquivOrdering.scala
+++ b/algebird-shapeless/src/main/scala/com/twitter/algebird/shapeless/EquivOrdering.scala
@@ -1,0 +1,40 @@
+package com.twitter.algebird.shapeless
+
+import shapeless._
+
+object EquivOrdering extends EquivOrdering1 {
+  implicit def hconsOrdering[A, B <: HList](implicit
+    la: Lazy[Ordering[A]],
+    b: Ordering[B]): Ordering[A :: B] =
+    new Ordering[A :: B] {
+      val a = la.value
+      def compare(x: A :: B, y: A :: B): Int = {
+        val c = a.compare(x.head, y.head)
+        if (c == 0) b.compare(x.tail, y.tail) else c
+      }
+    }
+
+  implicit val hnilOrdering: Ordering[HNil] =
+    new Ordering[HNil] {
+      def compare(x: HNil, y: HNil): Int = 0
+      override def equiv(a: HNil, b: HNil) = true
+    }
+
+  implicit def genericOrdering[A, Repr](implicit gen: Generic.Aux[A, Repr], r: Ordering[Repr]): Ordering[A] =
+    r.on(gen.to _)
+}
+
+abstract class EquivOrdering1 {
+
+  implicit def hconsEquiv[A, B <: HList](implicit la: Lazy[Equiv[A]], b: Equiv[B]): Equiv[A :: B] =
+    new Equiv[A :: B] {
+      val a = la.value
+      def equiv(x: A :: B, y: A :: B): Boolean =
+        a.equiv(x.head, y.head) && b.equiv(x.tail, y.tail)
+    }
+
+  implicit def genericEquiv[A, Repr](implicit gen: Generic.Aux[A, Repr], r: Equiv[Repr]): Equiv[A] =
+    new Equiv[A] {
+      def equiv(left: A, right: A) = r.equiv(gen.to(left), gen.to(right))
+    }
+}
--- a/algebird-shapeless/src/main/scala/com/twitter/algebird/shapeless/HListSum.scala
+++ b/algebird-shapeless/src/main/scala/com/twitter/algebird/shapeless/HListSum.scala
@@ -1,0 +1,14 @@
+package com.twitter.algebird.shapeless
+
+import shapeless._
+import shapeless.ops.hlist.ToList
+import com.twitter.algebird._
+
+object HListSum {
+  def sum[L <: HList, T](hlist: L)(implicit toList: ToList[L, T], m: Monoid[T]): T =
+    m.sum(toList(hlist))
+
+  def sum1[L <: HList, T](cons: T :: L)(implicit toList: ToList[L, T], s: Semigroup[T]): T =
+    // this get can never fail because there is at least one item
+    s.sumOption(cons.head :: toList(cons.tail)).get
+}
--- a/algebird-shapeless/src/main/scala/com/twitter/algebird/shapeless/Instances.scala
+++ b/algebird-shapeless/src/main/scala/com/twitter/algebird/shapeless/Instances.scala
@@ -1,0 +1,126 @@
+package com.twitter.algebird.shapeless
+
+import shapeless._
+import com.twitter.algebird._
+
+object Shapeless extends Shapeless3 {
+
+  implicit val hnilRing: Ring[HNil] =
+    new ConstantRing(HNil)
+
+}
+
+abstract class Shapeless3 extends Shapeless2 {
+  /**
+   * Pairwise ring for arbitrary heterogenuous lists (HList).
+   */
+  implicit def hconsRing[A, B <: HList](implicit
+    la: Lazy[Ring[A]],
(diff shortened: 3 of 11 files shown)


algebird-shapeless/src/main/scala/com/twitter/algebird/shapeless/EquivOrdering.scala
```diff
@@ -0,0 +1,40 @@
+package com.twitter.algebird.shapeless
+
+import shapeless._
+
+object EquivOrdering extends EquivOrdering1 {
+  implicit def hconsOrdering[A, B <: HList](implicit
+    la: Lazy[Ordering[A]],
+    b: Ordering[B]): Ordering[A :: B] =
```
travisbrown-stripe: Isn't it usually the tail where you need to worry about the divergence checker?

algebird-shapeless/src/main/scala/com/twitter/algebird/shapeless/Instances.scala
```diff
@@ -0,0 +1,126 @@
...
+object Shapeless extends Shapeless3 {
+
+  implicit val hnilRing: Ring[HNil] =
+    new ConstantRing(HNil)
+
+}
+
+abstract class Shapeless3 extends Shapeless2 {
+  /**
+   * Pairwise ring for arbitrary heterogenuous lists (HList).
```
travisbrown-stripe: Nitpick: there are [two ways](https://github.com/milessabin/shapeless/issues/233) to spell the h-word but this isn't one of them. 😄

algebird-shapeless/src/main/scala/com/twitter/algebird/shapeless/EquivOrdering.scala
```diff
@@ -0,0 +1,40 @@
+package com.twitter.algebird.shapeless
```
travisbrown-stripe: I know (from experience) that people are going to complain about the following not doing what you might expect:
```scala
import com.twitter.algebird._
import shapeless._
```

I'd probably call the package `generic` or `derivation` to avoid the whole issue. I really wish Shapeless would use a proper package path.

MansurAshraf:
I know its not ideal but is it possible to shade shapeless jar... its painful using Circe on Spark because spark is using an old version of shapeless and I suspect same thing will happen with this module

https://github.com/circe/circe/issues/713

johnynek:
@MansurAshraf if you shade the jar it is not useful except for Generic, which we already have macros for in core.

I think you just can't use the module, right? What could be done about it?

algebird-shapeless/src/main/scala/com/twitter/algebird/shapeless/HListSum.scala
```diff
@@ -0,0 +1,14 @@
...
+
+import shapeless._
+import shapeless.ops.hlist.ToList
+import com.twitter.algebird._
+
+object HListSum {
+  def sum[L <: HList, T](hlist: L)(implicit toList: ToList[L, T], m: Monoid[T]): T =
+    m.sum(toList(hlist))
+
+  def sum1[L <: HList, T](cons: T :: L)(implicit toList: ToList[L, T], s: Semigroup[T]): T =
```
travisbrown-stripe: I find the following little confusing. Given:

```scala
import com.twitter.algebird.Monoid
import shapeless.{ ::, HNil }

abstract class Base { def x: Int }
case class Foo(x: Int) extends Base
object Base {
  implicit val baseMonoid: Monoid[Base] =
    Monoid.from[Base](Foo(0))((a, b) => Foo(a.x + b.x))
}
```
And then:
```scala
scala> import com.twitter.algebird.shapeless.HListSum
import com.twitter.algebird.shapeless.HListSum

scala> val l = Foo(1) :: (Foo(2): Base) :: HNil
l: shapeless.::[Foo,shapeless.::[Base,shapeless.HNil]] = Foo(1) :: Foo(2) :: HNil

scala> HListSum.sum(l)
res0: this.Lub = Foo(3)

scala> HListSum.sum1(l)
<console>:16: error: could not find implicit value for parameter toList: shapeless.ops.hlist.ToList[shapeless.::[Base,shapeless.HNil],Foo]
       HListSum.sum1(l)
                    ^
```
I _think_ switching the order of the type parameters would fix it (?).

travisbrown-stripe (commented):
This looks reasonable to me, apart from some minor things noted below and the fact that I really do see this as a problem:

```scala
import com.twitter.algebird.Monoid
import com.twitter.algebird.Operators._
import com.twitter.algebird.shapeless.Shapeless._

case class Foo(x: Int)

object Foo {
  implicit val fooMonoid: Monoid[Foo] = 
    Monoid.from(Foo(1))((a, b) => Foo(a.x * b.x))
}

object Bar {
  val result = Foo(1) + Foo(1) // Foo(2)
}
```
…especially since it's not to hard to avoid with the export pattern.

travisbrown-stripe:
@MansurAshraf Shading the Shapeless jar would cause problems for us here, because we want to use this in conjunction with other libraries (like [scrooge-shapes](https://github.com/stripe/scrooge-shapes)) that provide generic instances.

algebird-shapeless/src/main/scala/com/twitter/algebird/shapeless/EquivOrdering.scala
johnynek: yeah, not really sure that is needed at all... probably insufficiently reviewed internally.

Should we add the Lazy if it is not needed? I really don't know when it is (e.g. it must not be here, since we are not using it on the tail).

travisbrown-stripe: @oscar I think this might be causing the failures on 2.10? Checking now…

algebird-shapeless/src/main/scala/com/twitter/algebird/shapeless/HListSum.scala
johnynek: how did you even discover that?

I'll add a test and try to fix it.

algebird-shapeless/src/main/scala/com/twitter/algebird/shapeless/EquivOrdering.scala
johnynek: yeah, I hit this myself. Generic would be consistent with circe, right? derivation is your custom macros, right?

travisbrown-stripe: Yeah, `generic` would be consistent with circe, and I think it'd have my vote.

algebird-shapeless/src/main/scala/com/twitter/algebird/shapeless/EquivOrdering.scala
travisbrown-stripe: Hmm, swapping the `Lazy` from heads to tails everywhere doesn't fix 2.10. Taking a look now.,

travisbrown-stripe: Removing the `Lazy` altogether makes at least the `Semigroup` one compile. My guess is some lingering version of [this bug](https://github.com/milessabin/shapeless/issues/367)—I can try to minimize this weekend and open a Shapeless issue.

johnynek:
side note @MansurAshraf you can use circe-derivation and not use shapeless. You know that right?

PS: spark is holding the whole fucking community hostage.

johnynek:
@travisbrown-stripe I changed to generic, fixed the spelling, added some tests. 2.10 still fails.

travisbrown:
Oh, right, the 2.10 failure is probably just a matter of a missing Paradise plugin.

travisbrown-stripe:
👍

At some point I think it'd be good to have tests that fail without the `Lazy`s, but I don't think we need to hold up this PR for that.

PR #656, opened 2018-05-12
Use scalafmt to test formatting after running tests

johnynek:
remove scala reform, add scalafmt not automatic, but as part of the CI.

The first PR should fail before I run fmt, then we will see if it succeeds.


johnynek:
I'm just going to merge this if it is green. It is only formatting changes.

PR #657, opened 2018-05-12
remove implicit from the caseclass macros

johnynek:
This is interestingly, not source compatible (since it removes implicit) but it is binary compatible.

This will likely be a small migration for people using it, but I think it is the right direction.

Notice the tests I added failed if we didn't remove the implicit. We got burned at Stripe on this.

cc @ianoc @travisbrown @MansurAshraf what do you think?

### DIFF
--- a/algebird-core/src/main/scala/com/twitter/algebird/macros/caseclass.scala
+++ b/algebird-core/src/main/scala/com/twitter/algebird/macros/caseclass.scala
@@ -4,9 +4,9 @@ import com.twitter.algebird._
 import scala.language.experimental.macros
 
 object caseclass {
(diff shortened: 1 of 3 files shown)


ianoc:
👍

travisbrown:
:+1: (and if enough people really want the fully-automatic version reinstated at some point, I'd recommend the approach circe takes with `io.circe.export`).

ianoc:
I think your original fix exposed for algebird fully automatic can lead to just dragons that are probably best avoided... (i.e. user needs to manually configure a semigroup with some settings, forgets, auto supplies something nuts. -- unit tests should catch, but alas low priority implicits wouldn't help).

PR #659, opened 2018-05-15
Add MonoidAggregator combinators

johnynek:
this some code we have been using at Stripe written by @non and myself.

We should support probably non-monoid (that is Semigroup based Aggregator) as well, although we try to use MonoidAggregator primarily at Stripe.

cc @erik-stripe


johnynek:
it looks like internally `Parallel.oneOf` and `ApplicativeAggregators` are the only ones we actually use, which maybe makes sense. Maybe we can only do those two for semigroups and maybe even remove the others from the PR in the interest of keeping things simpler...

algebird-generic/src/main/scala/com/twitter/algebird/generic/CombinedAggregators.scala
```diff
@@ -0,0 +1,76 @@
...
+ * `b1 => c` could be weighting the feature into the return space. We then
+ * sum the weights (this would be something like a linear of logistic regression).
+ *
+ * Let's use the following example code to demonstrate the usage here:
+ *
+ *   val a: MonoidAggregator[Animal, ColorStats, Result] = ...
+ *   val b: MonoidAggregator[Animal, ShapeStats, Result] = ...
+ *
+ *   val m1: Aggregator[Animal, ColorStats :: ShapeStats :: HNil, Result] =
+ *     CombinedAggregators(a :: b :: HNil)
```
MansurAshraf, line 30: Not sure about all the `HList` stuff but this seems like a combinator on `ApplicativeAggregators`. I would think that if I have:

```scala
val a: MonoidAggregator[Animal, ColorStats, Result] = ...
val b: MonoidAggregator[Animal, ShapeStats, Result]= ..
```
I should be able to do something like this

```scala
ApplicativeAggregators(a::b::HNil).andThenPresent{ case (a,b) => implicitly[Semigroup[Result]].combine(a,b)  }

```

algebird-generic/src/main/scala/com/twitter/algebird/generic/JoinedAggregators.scala
```diff
@@ -0,0 +1,174 @@
...
+ *
+ *   val m2: MonoidAggregator[Dog :+: Cat :+: CNil, DogStats :: CatStats :: HNil, Result] =
+ *     JoinedAggregators.oneOf(a :: b :: HNil)
+ *
+ * ...and get a single aggregator that works for dogs or cats (wrapped in a coproduct)!
+ *
+ * We can also say:
+ *
+ *   val m3: MonoidAggregator[Animal, DogStats :: CatStats :: HNil, Result] =
+ *     JoinedAggregators(a :: b :: HNil).composePrepare(Generic[Animal].to)
```
MansurAshraf, line 41: Same comment as above, why cant be implement this as a combinator using ApplicativeAggregators?

MansurAshraf (commented):
@johnynek are these compatible with [algebird-spark](https://github.com/twitter/algebird/blob/develop/algebird-spark/src/main/scala/com/twitter/algebird/spark/AlgebirdRDD.scala#L61)? I wonder if we will run into Serialization issues when HList based aggregators cross execution boundaries

johnynek:
Good point! We also have kryo serializers for HLists (easy to do, but a good idea to add).

We don’t customize our kryo serializers internally as far as I know (much less spark than scalding) but we did set up for scalding.

MansurAshraf:
@johnynek can we get this merged? Would love to use it at my work

algebird-generic/src/main/scala/com/twitter/algebird/generic/JoinedAggregators.scala
johnynek: Can you show me how? Suggest a change?

algebird-generic/src/main/scala/com/twitter/algebird/generic/CombinedAggregators.scala
johnynek: yeah, it's the same concept, it is just implemented directly, which is actually in a way easier due to all the shapeless work.

PR #660, opened 2018-05-15
remove unused imports, fix bug

johnynek:
@ianoc can you review?

### DIFF
--- a/algebird-util/src/main/scala/com/twitter/algebird/util/summer/HeavyHittersCachingSummer.scala
+++ b/algebird-util/src/main/scala/com/twitter/algebird/util/summer/HeavyHittersCachingSummer.scala
@@ -16,13 +16,9 @@ limitations under the License.
 package com.twitter.algebird.util.summer
 
 import com.twitter.algebird._
-import com.twitter.util.{Duration, Future, FuturePool}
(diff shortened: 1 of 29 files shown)


algebird-util/src/main/scala/com/twitter/algebird/util/summer/HeavyHittersCachingSummer.scala
```diff
@@ -107,7 +101,7 @@ class ApproxHHTracker(hhPct: HeavyHittersPercent, updateFreq: UpdateFrequency, r
 
     if (hh.containsKey(item)) {
       val v = hh.get(item)
-      val newItemCount = +1L
+      val newItemCount = v + 1L
```
johnynek, line 104: @ianoc can you double check this? This triggered a warning in scalac (which we ignored), but it looks like we intended to increment the count, but lost the `v`. What do you think?

I guess this means this cache was always pruning...

ianoc-stripe: ouch, yeah this change looks right/looks like a bug.

ianoc-stripe:
lgtm

PR #662, opened 2018-08-02
use builder in SetMonoid

nevillelyh:
Our benchmark shows that `var` of an immutable `Set` is ~50% faster than `mutable.Set` and `.toSet` later.
https://github.com/spotify/scio/issues/1183#issuecomment-410022235

### DIFF
--- a/algebird-core/src/main/scala/com/twitter/algebird/Monoid.scala
+++ b/algebird-core/src/main/scala/com/twitter/algebird/Monoid.scala
@@ -172,11 +172,11 @@ class SetMonoid[T] extends Monoid[Set[T]] {
   override def sumOption(items: TraversableOnce[Set[T]]): Option[Set[T]] =
     if (items.isEmpty) None
     else {
-      val mutable = scala.collection.mutable.Set[T]()
+      val builder = Set.newBuilder[T]
       items.foreach { s =>
-        mutable ++= s
+        builder ++= s
       }
-      Some(mutable.toSet)
+      Some(builder.result())
     }
 }
 


johnynek:
Thanks for the PR.

What's the motivation? Is this faster? What about actually using a Set.newBuilder? I would assume that should be the fastest way to build.

nevillelyh:
We actually tried `Set.newBuilder` too. Performance is about the same but the loop looks clunkier, i.e.
`items.foreach(_.foreach(x => builder += x))`

johnynek:
Since almost no one looks at the code, but it is used as a library in some perf critical roles, I'd rather use the fastest implementation in almost all cases as long as it is tested well and the API is clean.

johnynek:
like `items.foreach(builder ++= _)` might look clean and use the builder, right?

nevillelyh:
Didn't realize `Builder` has `++=`. Fixed.

algebird-core/src/main/scala/com/twitter/algebird/Monoid.scala
```diff
@@ -172,11 +172,11 @@ class SetMonoid[T] extends Monoid[Set[T]] {
   override def sumOption(items: TraversableOnce[Set[T]]): Option[Set[T]] =
     if (items.isEmpty) None
     else {
-      val mutable = scala.collection.mutable.Set[T]()
+      var builder = Set.newBuilder[T]
```
johnynek: minor nit: `val builder`?

johnynek (approved):

johnynek:
thank you!

nevillelyh:
Also just checking if you guys can release this and maybe chill anytime soon? It's been a while for both projects.

PR #665, opened 2018-08-28
Cuckoo filter

cesarcolle:
From the issue #560 
A first iteration of a CuckooFilter.

I use all the "majors" tests from BloomFilter.
```scala 
 property("CuckooFilter  is a Monoid") {
    commutativeMonoidLaws[CF[String]]
  }

  property("++ is the same as plus") {
    forAll { (a: CF[String], b: CF[String]) =>
      Equiv[CF[String]].equiv(a ++ b, cfMonoid.plus(a, b))
    }
  }

  property("+ is the same as adding with create") {
    forAll { (a: CF[String], b: String) =>
      Equiv[CF[String]].equiv(a + b, cfMonoid.plus(a, cfMonoid.create(b)))
    }
  }
  property("a ++ a = a for CF") {
    forAll { (a: CF[String]) =>
      Equiv[CF[String]].equiv(a ++ a, a)
    }
```
you can use like :
```scala
      val cfMonoid = new CuckooFilterMonoid[String](254)
      val cuckoo = cfMonoid.create("Aline", "Aline", "pour", "qu'elle", "revienne" )
      cuckoo.lookup("Aline")
```
I have seen complicate project around the cuckoo-filter but seem like the asymptotic behavior of the cuckoo filter allow to simplify the code.

If it's seems ok for you I can keep adding new features.

### DIFF
--- a/algebird-core/src/main/scala/com/twitter/algebird/CuckooFilter.scala
+++ b/algebird-core/src/main/scala/com/twitter/algebird/CuckooFilter.scala
@@ -21,40 +21,35 @@ import com.googlecode.javaewah.{IntIterator, EWAHCompressedBitmap => CBitSet}
 
 import scala.util.Random
 
-
 /**
-  * Cuckoo filter for Algebird in a Monoid way of life.
-  * referring :
-  *  - https://github.com/twitter/algebird/issues/560#issue-185161455
-  *  - http://www.eecs.harvard.edu/%7Emichaelm/postscripts/cuckoo-conext2014.pdf
-  *  - https://github.com/irfansharif/cfilter
-  * By nature, this filter isn't commutative
-  * From the inital paper, there is no problem to consider || fingerprint|| = ln(N) where N = fingerprintPerBucket * totalBucket
-  * "" as long as we use reasonably sized buckets, the fingerprint size can remain small. "", we'll use a a 32 bits fingerprint.
-  *
-  * We choose to put a fixed size for the fingerprint.
-  *
-  * TODO : Add the metrics for optimal size see  ASYMPTOTIC BEHAVIOR section of the paper.
-  * TODO : Make cuckoo filter works like an aggregator like the BloomFilter.
-  * TODO : Lookup method have to return a Approximate number like the size method (sometimes you can't insert an element).
+ * Cuckoo filter for Algebird in a Monoid way of life.
+ * referring :
+ *  - https://github.com/twitter/algebird/issues/560#issue-185161455
+ *  - http://www.eecs.harvard.edu/%7Emichaelm/postscripts/cuckoo-conext2014.pdf
+ *  - https://github.com/irfansharif/cfilter
+ * By nature, this filter isn't commutative
+ * From the inital paper, there is no problem to consider || fingerprint|| = ln(N) where N = fingerprintPerBucket * totalBucket
+ * "" as long as we use reasonably sized buckets, the fingerprint size can remain small. "", we'll use a a 32 bits fingerprint.
+ *
+ * We choose to put a fixed size for the fingerprint.
+ *
+ * TODO : Add the metrics for optimal size see  ASYMPTOTIC BEHAVIOR section of the paper.
+ * TODO : Make cuckoo filter works like an aggregator like the BloomFilter.
+ * TODO : Lookup method have to return a Approximate number like the size method (sometimes you can't insert an element).
   **/
-
-
 object CuckooFilter {
   // TODO : optimal parameter have to be compute here : see Asymptotic behavior of the original paper.
   def apply[A](bucketSize: Int, fingerPrintSize: LongBitSet, kicks: Int, buckets: Int)(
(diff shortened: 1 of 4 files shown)


johnynek:
Thank you!

I'll post a review in the next day or two.

cesarcolle:
The cuckoo filter can be used as BloomFilter
i.e : 
```scala
      val cfMonoid1 = new CuckooFilterMonoid[String](32, 256)
      val cf1 = cfMonoid1.create("1", "2", "3", "4", "100")
      val lookup = cf1.lookup("1")
      assert(lookup)
```
There is not enough resources around cuckoo filter for : approximation size, optimal value, ... But theoretically, it's better than Bloomfilter. perhaps a mapping between bloom filter estimation and cuckoo filter ?

cesarcolle:
This is now ready for review :) @johnynek 

Thanks !

algebird-core/src/main/scala/com/twitter/algebird/CuckooFilter.scala
```diff
@@ -0,0 +1,435 @@
...
+ * From the inital paper, there is no problem to consider || fingerprint|| = ln(N) where N = fingerprintPerBucket * totalBucket
+ * "" as long as we use reasonably sized buckets, the fingerprint size can remain small. "", we'll use a a 32 bits fingerprint.
+ *
+ * We choose to put a fixed size for the fingerprint.
+ *
+ * TODO : Make cuckoo filter works like an aggregator like the BloomFilter.
+ * TODO : Lookup method have to return a Approximate number like the size method (sometimes you can't insert an element).
+  **/
+object CuckooFilter {
+  def apply[A](fingerprintPerBucket: Int, buckets: Int = 256)(
```
nevillelyh, line 40: The original paper, as well as this Java [impl](https://github.com/bdupras/guava-probably) uses 2/4/8 fingerprints per bucket. Looks like you're using 10 & 50 in tests, any particular reason?

algebird-core/src/main/scala/com/twitter/algebird/CuckooFilter.scala
```diff
@@ -0,0 +1,435 @@
...
+
+/**
+ * Cuckoo filter for Algebird in a Monoid way of life.
+ * referring :
+ *  - https://github.com/twitter/algebird/issues/560#issue-185161455
+ *  - http://www.eecs.harvard.edu/%7Emichaelm/postscripts/cuckoo-conext2014.pdf
+ *  - https://github.com/irfansharif/cfilter
+ * By nature, this filter isn't commutative
+ * From the inital paper, there is no problem to consider || fingerprint|| = ln(N) where N = fingerprintPerBucket * totalBucket
+ * "" as long as we use reasonably sized buckets, the fingerprint size can remain small. "", we'll use a a 32 bits fingerprint.
```
nevillelyh, line 32: Original paper suggests that fingerprint is best below 8 bits. Any reason you're using 32?

algebird-core/src/main/scala/com/twitter/algebird/CuckooFilter.scala
```diff
@@ -0,0 +1,435 @@
...
+
+  def plus(left: CF[A], right: CF[A]): CF[A] = left ++ right
+
+  /**
+   * it's come from the BloomFilter. But I don't think it's a good idea because insertion from scratch
+   * describe by the "set" and "add" functions are not safe. */
+  override def sumOption(iter: TraversableOnce[CF[A]]): Option[CF[A]] =
+    if (iter.isEmpty) None
+    else {
+      val buckets = Array.fill[CBitSet](totalBuckets)(new CBitSet(fingerprintBucket))
```
nevillelyh, line 63: It's more efficient to use 1 bitset of `numBuckets * numEntriesPerBucket * numBitsPerFingerprint`? IIRC JVM pointers are 8 bytes, so there's `totalBuckets * 8` overhead here, plus those from `CBitSet` and memory lookups.

According to the paper, each bucket should have `numEntriesPerBucket` slots, each with `numBitsPerFingerprint`. how does `fingerprintBucket` fit in here?

algebird-core/src/main/scala/com/twitter/algebird/CuckooFilter.scala
```diff
@@ -0,0 +1,435 @@
...
+      var sets = 0
+
+      @inline def setFingerprint(index: Int, fp: Int): Unit = {
+        buckets(index).set(fp)
```
nevillelyh, line 67: This setting a single bit? Looking at `def fingerprint()`, it's in `[0, Int.MaxValue]` and `>>> fingerprintBucket`, so this would overflow?

nevillelyh (commented):
I'm not sure I understand the impl well enough but it seems quite different from the paper and some other impls I've seen. Also not sure if a `Semigroup` approach makes sense given the non-deterministic nature of CF. I managed to get 100M insertions in ~2min with this Java [impl](https://github.com/bdupras/guava-probably) and we're most likely going with that approach in our code.

Issue #632, opened 2017-05-22
Size on saturated bloom filter throws an exception


cacts:
using the following example:

```
import com.twitter.algebird._
import scala.util.Random

implicit val hash = Hash128.stringHash
val bf = BloomFilter(2, 0.05)
val strings = for (_ <- 0 to 9) yield Random.nextString(5)

val filter = bf.create(strings: _*)
println(filter.size.estimate)
```

we get:

> java.lang.IllegalArgumentException: requirement failed
> 	at scala.Predef$.require(test.sc:208)
> 	at com.twitter.algebird.Approximate.<init>(test.sc:73)
> 	at com.twitter.algebird.BloomFilter$.sizeEstimate(test.sc:168)
> 	at com.twitter.algebird.BFInstance.size(test.sc:595)

in previous versions (0.11.0) the Approximate returned by size would be invalid, but now throws an exception due to the assertion.

cesarcolle:
I reproduce : 
for an input like : 
```scala
Array(0, 1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12)
```
From the BloomFilter object, the size method use 
```scala
def sizeEstimate(numBits: Int,
                   numHashes: Int,
                   width: Int,
                   approximationWidth: Double = 0.05): Approximate[Long] = ???
```
return an _Approximate_ object like : 

```scala
Approximate(min = 0, estimate = -1, max = 0, prob =  1.0)
```

This fail with the : 
```scala
require(numeric.lteq(min, estimate) && numeric.lteq(estimate, max))
```
 property of the Approximate object.
This issue can be handle from the BloomFilter or the Approximate object.

I can fix it but I don't know where to start :)

johnynek:
Great...

Step 1: send a PR with the failing test.
Step 2: lets investigate why the invariant is false, having an estimate of -1 when the min is 0 is a problem. We probably have some arithmetic issue here, or some other problem.
Step 3: fix the bug and watch PR go green.

Happy to review the code!

PR #666, opened 2018-08-29
Saturated bloomfilter size

cesarcolle:
I reproduce the error from issue #632 

1. The error from the initial issue is reproduced with the test : 
```scala
    "return his size event if it's saturated" in {
      val bfMonoid = BloomFilterMonoid[String](5, 13)
      val strings = Array(1, 2, 3, 4, 5, 6, 7, 8, 9, 10).map(_.toString)
      val bf = bfMonoid.create(strings: _*)

      assert(bf.size.min > 0)
    }
```
throws : 
```
[info] - should return his size event if it's saturated *** FAILED ***
[info]   java.lang.IllegalArgumentException: requirement failed
```

Due to, I think , https://github.com/twitter/algebird/issues/632#issuecomment-416946292 we discuss before.

2. This is not the only error I found, seem like a saturated bloomfilter also provided a huge  max approximation for a given input. But not throw an exception

Coming from the comment (L.135 to 138 ) in Bloomfilter file :
```scala
/*
   * approximationWidth defines an interval around the maximum-likelihood cardinality
   * estimate. Namely, the approximation returned is of the form
   * (min, estimate, max) =
   *   ((1 - approxWidth) * estimate, estimate, (1 + approxWidth) * estimate)
   */
```

But there is a problem between the result from the : 

```scala
    "return a max approximate size if it's saturated" in {
      val bfMonoid = BloomFilterMonoid[String](5, 13)
      val strings = List(8, 9, 8, 10, 1, 8, 11, 7, 7, 1).map(_.toString)
      val bf = bfMonoid.create(strings: _*)
      // even it's seems big. This assert fail..
      assert(bf.size.max < strings.length * 100)
    }
```
which return : 
```
[info] - should return a max approximate size if it's saturated *** FAILED ***
[info]   2147483647 was not less than 1000 (BloomFilterTest.scala:459)
```
**2147483647** is a particular number because it's the maximum 32 bits integer you can have.

### DIFF
--- a/README.md
+++ b/README.md
@@ -77,7 +77,6 @@ See [Algebird's page on the Scaladex](https://index.scala-lang.org/twitter/algeb
 - [Scio](https://github.com/spotify/scio)
 - [Packetloop](https://www.packetloop.com) (see [this tweet](https://twitter.com/cloudjunky/status/355073917720858626))
 - Ebay uses Algebird for machine learning: [ScalaDays talk](http://www.slideshare.net/VitalyGordon/scalable-and-flexible-machine-learning-with-scala-linkedin)
-- [Apple (FEAR Team)](https://news.ycombinator.com/item?id=16969118)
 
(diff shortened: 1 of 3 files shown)


johnynek (commented):
Thanks, this looks good.

Could we possibly revert the style changes for code you are not touching?

cesarcolle:
Sorry it's Intelij clean code not working with the .scalafmt.conf of the project. I apply algebird's scalafmt by hand

PR #672, opened 2019-02-21
More benchmarks for creating bloom filters

anish749:
In some use cases in [Scio](https://github.com/spotify/scio) we create large Bloom filters with around 5 ~ 50M entries in some batch data pipelines. We were using BloomFilterAggregator API for most of our uses cases, and it felt like a good idea to bench mark the aggregator for comparison.


johnynek:
by the way, I'm happy to merge this, but you never signed the CLA. Could you do that?

anish749:
Hello, I am extremely sorry I missed the CLA 🤦, I initially thought I already did it.
I tried doing it again, the comment got updated now, but the check still seems to be yellow.

regadas:
Yeah I think we add a similar case before where the status got stuck in the Pending phase ... couldn't find the PR where that happened though

PR #674, opened 2019-02-23
Update Ruby version in travis

anish749:
The builds via travis are failing because of an outdated ruby version which is no longer supported.
updating to the latest version passes the builds.


johnynek:
Thank you for fixing the build!

PR #677, opened 2019-09-16
Update TravisCI jdk

regadas:
default jdk is now `openJDK 11`; Not sure if it's worth adding it to the matrix?


johnynek (approved):
Thank you! will merge when 2.11 is green (hit a flaky test)

PR #678, opened 2019-09-16
Update scala to 2.12.10

regadas:
This requires #677

### DIFF
--- a/sbt
+++ b/sbt
@@ -10,7 +10,7 @@ declare -r sbt_release_version="0.13.17"
 declare -r sbt_unreleased_version="0.13.17"
 
 declare -r latest_213="2.13.0-M5"
-declare -r latest_212="2.12.7"
+declare -r latest_212="2.12.10"
 declare -r latest_211="2.11.12"
 declare -r latest_210="2.10.7"
 declare -r latest_29="2.9.3"
--- a/.travis.yml
+++ b/.travis.yml
@@ -12,7 +12,7 @@ matrix:
       after_success:
         - bash <(curl -s https://codecov.io/bash)
 
-    - scala: 2.12.7
+    - scala: 2.12.10
       jdk: oraclejdk8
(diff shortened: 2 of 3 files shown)


johnynek:
after we merge #677, merge in master and maybe CI will go green.

johnynek:
thanks!

johnynek:
looks like the java-api stuff breaks in 2.12.10... man... what happened with 2.12.10? It seems like it is breaking more that you would expect for a stable minor version.

johnynek:
cc @smarter

smarter:
> looks like the java-api stuff breaks in 2.12.10... man... what happened with 2.12.10? It seems like it is breaking more that you would expect for a stable minor version.

This happened: https://github.com/scala/scala/commit/e48cfd26f66c8f128f966ecd100d23171d9429cf which backported this: https://github.com/scala/scala/commit/3aea776ca1aa82c9de44cc6806dcdb242f3b40f8. Basically chill was taking advantage of a compiler bug, these classes were never supposed to be visible from Scala (they can't really be hidden from Java, so they're still visible there)

regadas:
So it seems that the culprit here is `ScalaUnidoc`/`doc`. For 2.12 `javaapi` is already being discarded, so I applied the same thing to the `ScalaUnidoc` scope.

https://github.com/twitter/algebird/blob/4bb8ffb1e10614e99bd19f585e49eacb79da0f1c/build.sbt#L209-L210

The supposed issue with java api and 2.12.10 led me to #679

sbt
```diff
@@ -10,7 +10,7 @@ declare -r sbt_release_version="0.13.17"
 declare -r sbt_unreleased_version="0.13.17"
 
 declare -r latest_213="2.13.0-M5"
-declare -r latest_212="2.12.7"
+declare -r latest_212="2.12.10"
```
nevillelyh, line 13: Unrelated, do we still need this wrapper? It was needed back in the days when sbt wasn't available in CI env?

johnynek: I don’t think so. We used to use it so people didn’t need sbt installed.

PR #681, opened 2019-09-17
Use scalameta scalafmt plugin

regadas:
As I expected the switch to the official and updated scalafmt introduces a "some" changes. `sbt` files are also being checked now.


regadas:
So apparently, I think there's an issue with `mima` and `scala 2.12.10`; The issue is visible in this job and the in the previous commit merged to master. 

https://github.com/lightbend/mima/issues/387

johnynek:
I think we should stay on 2.12.8 or 9 rather than introduce binary incompatibilities.

regadas:
@johnynek  I went through `2.12.{8, 9, 10}` and MiMa reports issues in all of them when comparing against the latest version `0.13.5` with `2.12.7`. AFAICT it's mostly the `DirectMissingMethodProblem` which, and if i understood correctly, is related to what @dwijnand [mentioned](https://github.com/lightbend/mima/pull/386#issuecomment-532536254).

Since this is a new thing being reported and previous kinda have the same issue, maybe it's kind of ok to `exclude` this for now till a new release is out?

dwijnand:
I've opened https://github.com/lightbend/mima/issues/388 that describes the problem and how MiMa could be enhanced to handle it universally.

regadas:
@johnynek rebased with `develop`

johnynek (approved):
Thanks!

PR #682, opened 2019-09-18
Revert to scala 2.12.9 and ignore MiMa false positives

regadas:
See for more context: https://github.com/lightbend/mima/issues/388

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -1,6 +1,6 @@
 import ReleaseTransformations._
 import algebird._
-import com.typesafe.tools.mima.plugin.MimaPlugin.mimaDefaultSettings
+import com.typesafe.tools.mima.core._
 import pl.project13.scala.sbt.JmhPlugin
 
 val algebraVersion = "0.7.0"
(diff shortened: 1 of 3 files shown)


build.sbt
```diff
@@ -151,6 +151,56 @@ lazy val noPublishSettings = Seq(
   publishArtifact := false
 )
 
+// https://github.com/lightbend/mima/issues/388
+lazy val mimaSettings = Def.settings(
```
regadas, line 155: This is not pretty but I guess we will be able to remove this once that issue is fixed

johnynek:
@regadas This seems fine to get things green for a release, but I think we should revert it after the next publish. What do you think?

regadas:
@johnynek sounds good! 👍

regadas:
flaky `tut`?

johnynek:
I'll rerun but :`(run-main-0) java.lang.NoSuchMethodError: scala.reflect.internal.util.ReusableInstance.<init>(Lscala/Function0;)V` doesn't look flakey. It looks like binary incompatibility.

Maybe tut isn't compatible with one of these version bumps?

maybe an sbt caching bug?

regadas:
Oh! I missed that! I'll have a look.

johnynek:
Maybe we should revert the 2.12 bump. Master is red now and it caused all the scala steward bumps to fail.

regadas:
Yup agreed! yesterday I tracked this a bit and I think there's something fishy around `tut` and `2.12.10` repl interpreter that it's instantiated; seems that the classpath is not right.

`2.12.9` is ok I'll push a commit here to double check.

nevillelyh (approved):

PR #705, opened 2019-09-27
Add scalafix

regadas:
Hi @johnynek I was thinking on your last comment and decided to add scalafix, wdyt?

I also enabled `-Ywarn-unused` for scala `2.12` it throws a lot of warns that scalafix can't fix; we will need to slowly fix them manually.


johnynek:
This is great! thanks!

PR #708, opened 2019-09-30
Remove most of the unused warns

regadas:
Hi @johnynek so I tracked down most of them. We still have some unused implicits and case classes that should have non-implicit param list. Both of them break binary compatibility however fixing `case class` should be less problematic.

### DIFF
--- a/algebird-core/src/main/scala/com/twitter/algebird/BloomFilter.scala
+++ b/algebird-core/src/main/scala/com/twitter/algebird/BloomFilter.scala
@@ -455,8 +455,8 @@ case class BFItem[A](item: A, hashes: BFHash[A], width: Int) extends BF[A] {
 
   def ++(other: BF[A]): BF[A] =
     other match {
-      case _ @BFZero(_, _)            => this
-      case _ @BFItem(otherItem, _, _) => toSparse + otherItem
+      case BFZero(_, _)            => this
+      case BFItem(otherItem, _, _) => toSparse + otherItem
       case _                          => other + item
     }
 
@@ -504,8 +504,8 @@ case class BFSparse[A](hashes: BFHash[A], bits: CBitSet, width: Int) extends BF[
     require(this.numHashes == other.numHashes)
 
     other match {
-      case _ @BFZero(_, _)       => this
-      case _ @BFItem(item, _, _) => this + item
+      case BFZero(_, _)       => this
+      case BFItem(item, _, _) => this + item
       case bf @ BFSparse(_, otherBits, _) => {
         // assume same hashes used
 
@@ -576,7 +576,7 @@ case class BFInstance[A](hashes: BFHash[A], bits: BitSet, width: Int) extends BF
       case BFSparse(_, otherBits, _) =>
         // assume same hashes used
         BFInstance(hashes, bits | (new RichCBitSet(otherBits)).toBitSet(width), width)
-      case _ @BFInstance(_, otherBits, _) => {
+      case BFInstance(_, otherBits, _) => {
         // assume same hashes used
         BFInstance(hashes, bits ++ otherBits, width)
       }
--- a/algebird-core/src/main/scala/com/twitter/algebird/BloomFilter.scala
+++ b/algebird-core/src/main/scala/com/twitter/algebird/BloomFilter.scala
@@ -400,9 +400,9 @@ sealed abstract class BF[A] extends java.io.Serializable {
     (this, that) match {
       // Comparing with empty filter should give number
       // of bits in other set
-      case (x: BFZero[A], y: BFZero[A]) => 0
-      case (x: BFZero[A], y: BF[A])     => y.numBits
-      case (x: BF[A], y: BFZero[A])     => x.numBits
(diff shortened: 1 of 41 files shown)


algebird-util/src/main/scala/com/twitter/algebird/util/summer/AsyncSummer.scala
```diff
@@ -28,13 +28,13 @@ trait AsyncSummer[T, +M <: Iterable[T]] { self =>
...
   def cleanup: Future[Unit] = Future.Unit
-  def withCleanup(cleanup: () => Future[Unit]): AsyncSummer[T, M] = {
+  def withCleanup(cleanupFn: () => Future[Unit]): AsyncSummer[T, M] = {
     val oldSelf = self
     new AsyncSummerProxy[T, M] {
       override val self = oldSelf
       override def cleanup =
         oldSelf.cleanup.flatMap { _ =>
-          cleanup
+          cleanupFn()
```
regadas, line 37: This seemed like a potential bug?

nevillelyh (approved):
LGTM. How do we want to deal with the potentially breaking implicit arg changes? Bump version or keep them as is? @johnynek

algebird-core/src/main/scala/com/twitter/algebird/BloomFilter.scala
```diff
@@ -455,9 +455,9 @@ case class BFItem[A](item: A, hashes: BFHash[A], width: Int) extends BF[A] {
 
   def ++(other: BF[A]): BF[A] =
     other match {
-      case bf @ BFZero(_, _)            => this
-      case bf @ BFItem(otherItem, _, _) => toSparse + otherItem
-      case _                            => other + item
+      case _ @BFZero(_, _)            => this
```
johnynek: instead of `_ @ BFZero(_, _)` can we write `BFZero(_, _)`? Similar below for BFItem?

algebird-core/src/main/scala/com/twitter/algebird/BloomFilter.scala
```diff
@@ -504,8 +504,8 @@ case class BFSparse[A](hashes: BFHash[A], bits: CBitSet, width: Int) extends BF[
     require(this.numHashes == other.numHashes)
 
     other match {
-      case bf @ BFZero(_, _)       => this
-      case bf @ BFItem(item, _, _) => this + item
+      case _ @BFZero(_, _)       => this
+      case _ @BFItem(item, _, _) => this + item
```
johnynek: can we remove `_@`?

algebird-core/src/main/scala/com/twitter/algebird/BloomFilter.scala
```diff
@@ -576,7 +576,7 @@ case class BFInstance[A](hashes: BFHash[A], bits: BitSet, width: Int) extends BF
       case BFSparse(_, otherBits, _) =>
         // assume same hashes used
         BFInstance(hashes, bits | (new RichCBitSet(otherBits)).toBitSet(width), width)
-      case bf @ BFInstance(_, otherBits, _) => {
+      case _ @BFInstance(_, otherBits, _) => {
```
johnynek: remove `_ @`

algebird-util/src/main/scala/com/twitter/algebird/util/summer/AsyncSummer.scala
johnynek: yeah it does. This would have been an infinite loop, right? which I guess means it was never used?

johnynek (changes requested):
couple minor requests, this looks good, thanks!

johnynek:
@nevillelyh I didn't see a breaking change anywhere, and if we had one, I think mima should catch it. Can you comment on the line number you think is a breaking change?

nevillelyh:
It's the `Ordering` & `Monoid` implicits for `SketchMapAggregator` but I guess it's binary compatible despite the signature change.
https://github.com/twitter/algebird/pull/708/files#diff-9e23a301d88510f7cf5dae672bc3b9a3L243

algebird-core/src/main/scala/com/twitter/algebird/BloomFilter.scala
regadas: 🤦‍♂ yeah totally 👍

regadas:
@nevillelyh @johnynek yup, there are some implicit arg changes here but they translate to the same thing and should not break binary compatibility; My point was, and sorry if it led to confusion; that if we want to further remove the outstanding warnings we will break it. I can make another PR or create an issue pointing them out.

algebird-util/src/main/scala/com/twitter/algebird/util/summer/AsyncSummer.scala
regadas: indeed.

johnynek:
yeah, I thought those implicit changes were okay because they are in the same order, same binary result.

johnynek (approved):

PR #712, opened 2019-10-01
remove sbt wrapper and 0.13 cache settings

nevillelyh:


regadas (approved):
LGTM. that sbt-extras script is pretty outdated!

PR #714, opened 2019-10-01
Fix: case classes must have a non-implicit parameter list

regadas:
This triggers mima with the same issue we have seen when updating mima to the last version.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -197,6 +197,8 @@ lazy val mimaSettings = Def.settings(
       .exclude[DirectMissingMethodProblem]("com.twitter.algebird.util.summer.HeavyHittersPercent.apply"),
     ProblemFilters.exclude[DirectMissingMethodProblem]("com.twitter.algebird.util.summer.BufferSize.apply"),
     ProblemFilters.exclude[DirectMissingMethodProblem]("com.twitter.algebird.util.summer.Compact.apply"),
+    ProblemFilters.exclude[DirectMissingMethodProblem]("com.twitter.algebird.MinAggregator.ord"),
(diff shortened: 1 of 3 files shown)


build.sbt
```diff
@@ -197,6 +197,8 @@ lazy val mimaSettings = Def.settings(
       .exclude[DirectMissingMethodProblem]("com.twitter.algebird.util.summer.HeavyHittersPercent.apply"),
     ProblemFilters.exclude[DirectMissingMethodProblem]("com.twitter.algebird.util.summer.BufferSize.apply"),
     ProblemFilters.exclude[DirectMissingMethodProblem]("com.twitter.algebird.util.summer.Compact.apply"),
+    ProblemFilters.exclude[DirectMissingMethodProblem]("com.twitter.algebird.MinAggregator.ord"),
```
nevillelyh: I guess it's backwards compat as long as people don't call `aggr.ord`?

nevillelyh (approved):

build.sbt
regadas: yeah right, with the previous syntax that was totally possible;

johnynek: I’m nervous about each of these. I think we should be very reluctant to risk binary incompatibility to clean up warts.

nevillelyh: @johnynek yeah those change were reverted in 2nd commit of the PR. The squashed commit doesn't have any MiMa changes.

johnynek: Ahh. Right. Good point.  Thanks.

regadas: Yup ended up achieving the same thing; @johnynek was the access to `ord` intended by design?

johnynek: I see no reason to hide it, you can reconstruct it from max or min:

```
def lteq(a: A, b: A) = max(a, b) == b
```

PR #717, opened 2019-10-04
Add return types and override modifiers

regadas:
This makes things a little bit more consistent; For now I only did it for `algebird-core`

### DIFF
--- a/algebird-core/src/main/scala/com/twitter/algebird/Ring.scala
+++ b/algebird-core/src/main/scala/com/twitter/algebird/Ring.scala
@@ -53,8 +53,8 @@ import scala.annotation.implicitNotFound
 
 @implicitNotFound(msg = "Cannot find Ring type class for ${T}")
 trait Ring[@specialized(Int, Long, Float, Double) T] extends Group[T] with CommutativeGroup[T] with ARing[T] {
-  def one: T
-  def times(a: T, b: T): T
+  override def one: T
+  override def times(a: T, b: T): T
   override def product(iter: TraversableOnce[T]): T =
     if (iter.isEmpty) one // avoid hitting one as some have abused Ring for Rng
     else iter.reduce(times)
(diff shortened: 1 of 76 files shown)


algebird-core/src/main/scala/com/twitter/algebird/Ring.scala
```diff
@@ -243,19 +243,19 @@ object Ring extends GeneratedRingImplicits with ProductRings with RingImplicits0
...
-  implicit def shortRing: Ring[Short] = ShortRing
-  implicit def jshortRing: Ring[JShort] = JShortRing
-  implicit def longRing: Ring[Long] = LongRing
-  implicit def bigIntRing: Ring[BigInt] = BigIntRing
-  implicit def jlongRing: Ring[JLong] = JLongRing
-  implicit def floatRing: Ring[Float] = FloatRing
-  implicit def jfloatRing: Ring[JFloat] = JFloatRing
-  implicit def doubleRing: Ring[Double] = DoubleRing
-  implicit def jdoubleRing: Ring[JDouble] = JDoubleRing
+  implicit val boolRing: Ring[Boolean] = BooleanRing
```
johnynek: why change these to `val`? `def` is just as good since these are all objects that have already been allocated, no?

johnynek (commented):
this looks great! I had one question.

johnynek:
How did you do this BTW? Was this a scalafix?

johnynek:
can we do this without making mima unhappy?

algebird-core/src/main/scala/com/twitter/algebird/Ring.scala
regadas: Ahh oops! yup in this case it's the same thing; I was doing this initially to propose a consistent way to define implicit's;

I'll revert this!

regadas:
I tried to use `scalafix` ... but it didn't work out well enough. There's a [rule](https://github.com/scalacenter/scalafix/blob/master/scalafix-rules/src/main/scala/scalafix/internal/rule/ExplicitResultTypes.scala) for it, but is a bit buggy; I took the easy way out and used IntelliJ for a bit 😄 

RE: MiMa ... yeah it should be possible

johnynek (approved):
Thanks for this cleanup!

Issue #671, opened 2019-02-14
Scala 2.13.x


regadas:
Hello!

I was wondering if there's anything planned to cross compile to 2.13.
Probably some of the modules will not be possible for quite some time, like `algebird-spark`.

johnynek:
Happy to take PRs!

skonto:
I am wondering about the `algebird-spark` dependency. Spark requires chill and chill requires `algebird-core`. Can we have independent releases per module per scala version?

johnynek:
We can, but it is a pain with SBT. But doable. We have done it in the past before spark has 2.12 support.

ayoub-benali:
Is this issue worked on ?

johnynek:
I don’t know if anyone is working on it now. The other ticket referenced above had some blockers.

regadas:
Hi, just wanted to let you know that I got around to compile `algebird-core` with `2.13` I should be able to PR something soonish ™️

johnynek:
Amazing! That’s really great.

regadas:
@johnynek is it still important to keep support for 2.10?

The reason I'm asking is because some deps needed for 2.13 stopped being published for 2.10. This leads us to situation where we might have cross compilation with different dep versions. I don't think this is ideal and should be avoided.

SethTisue:
continued support for 2.10 is becoming increasingly rare in Scala OSS. dropping it sounds reasonable to me

johnynek:
Yeah, dropping 2.10 is fine.

h-vetinari:
Xref #719, which added 2.13 support for all but
- [ ] algebird-bijection (blocked on twitter/bijection#287)
- [ ] algebird-spark

Also, 2.10 was dropped in #721.

regadas:
`algebird-bijection` support is added in https://github.com/twitter/algebird/pull/776

TBH I don't know the status of `spark` supporting 2.13. I guess it will take awhile.

Should we close this issue?

johnynek:
Thanks, I guess spark support will have to wait until spark 3?

PR #718, opened 2019-10-06
Fix deprecated symbol literals

regadas:
Symbol literals `'example`  will actually invoke `Symbol("example")`. With 2.13 they are deprecated.

Related to #671

### DIFF
--- a/algebird-test/src/main/scala/com/twitter/algebird/BaseProperties.scala
+++ b/algebird-test/src/main/scala/com/twitter/algebird/BaseProperties.scala
@@ -175,7 +175,7 @@ object BaseProperties extends MetricProperties {
     }
 
   def isAssociativeDifferentTypes[T: Semigroup: Equiv, U <: T: Arbitrary]: Prop =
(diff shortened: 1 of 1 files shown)


algebird-test/src/main/scala/com/twitter/algebird/BaseProperties.scala
```diff
@@ -284,7 +284,7 @@ object BaseProperties extends MetricProperties {
     isDistributiveDifferentTypes[T, T]
 
   def timesIsAssociative[T: Ring: Arbitrary: Equiv]: Prop =
-    'timesIsAssociative |: forAll { (a: T, b: T, c: T) =>
+    Symbol("timesIsAssociative") |: forAll { (a: T, b: T, c: T) =>
```
johnynek: I think you can just use strings here actually. No need for the Symbols at all.

regadas: indeed! I wasn't sure if there was a not obvious reason for the saved up allocations.

johnynek (approved):
Thanks!

PR #719, opened 2019-10-09
Add scala 2.13 support

regadas:
This is an initial take; Currently only targeting `algebird-core` and `algebird-test`.

There are some caveats to this:
1) ~I don't think we will be able to keep binary compatibility~
2) ~`algebra` and `scala-collection-compat` don't publish for 2.10; I can backport some of the things in `scala-collection-compat` but we would need to use a different version of `algebra` for 2.10 which is not ideal~
3) still some things to improve here...

* [x] algebird-core
* [x] algebird-test
* [x] algebird-generic
* [x] algebird-util
* [ ] algebird-bijection
* [ ] algebird-spark

~_NOTE_: MiMa is temporarily disabled~


.travis.yml
```diff
@@ -16,15 +16,19 @@ matrix:
 
     - scala: 2.11.12
       jdk: openjdk8
-      script: sbt ++$TRAVIS_SCALA_VERSION clean coverage test coverageReport mimaReportBinaryIssues
+      script: sbt ++$TRAVIS_SCALA_VERSION clean coverage test coverageReport
```
regadas: I disabled MiMA temporarily just to see if CI likes this.

johnynek:
I guess we can cut off 2.10. If you are that far behind you probably don’t need the latest version of algebird.

.travis.yml
```diff
@@ -31,15 +31,21 @@ matrix:
       script: sbt \
         "++$TRAVIS_SCALA_VERSION clean" \
         "++$TRAVIS_SCALA_VERSION test" \
-        "++$TRAVIS_SCALA_VERSION docs/makeMicrosite" \
-        "++$TRAVIS_SCALA_VERSION mimaReportBinaryIssues"
+        "++$TRAVIS_SCALA_VERSION docs/makeMicrosite"
+        # "++$TRAVIS_SCALA_VERSION mimaReportBinaryIssues"
```
johnynek: What fails if we turn this check back on?

algebird-core/src/main/scala/com/twitter/algebird/SketchMap.scala
```diff
@@ -169,7 +168,7 @@ case class SketchMapParams[K](seed: Int, width: Int, depth: Int, heavyHittersCou
    */
   def updatedHeavyHitters[V: Ordering](hitters: Seq[K], table: AdaptiveMatrix[V]): List[K] = {
     val mapping: Map[K, V] =
-      hitters.map(item => (item, frequency(item, table)))(breakOut)
+      hitters.map(item => (item, frequency(item, table))).toMap
```
johnynek: hitters.iterator.map so we don’t allocate and intermediate Seq

algebird-core/src/main/scala/com/twitter/algebird/Aggregator.scala
```diff
@@ -102,12 +102,12 @@ object Aggregator extends java.io.Serializable {
       override def append(l: T, r: F): T = appnd(l, r)
 
       override def appendAll(old: T, items: TraversableOnce[F]): T =
-        if (items.isEmpty) old else reduce(old, agg(items).get)
+        if (items.iterator.isEmpty) old else reduce(old, agg(items).get)
```
sritchie, line 106: Does this help in this case? don't collections usually have efficient implementations of `.isEmpty` that don't modify the collection?

algebird-core/src/main/scala/com/twitter/algebird/AveragedValue.scala
```diff
@@ -147,11 +148,11 @@ object AveragedGroup extends Group[AveragedValue] with CommutativeGroup[Averaged
    * intermediate objects.
    */
   override def sumOption(iter: TraversableOnce[AveragedValue]): Option[AveragedValue] =
-    if (iter.isEmpty) None
+    if (iter.iterator.isEmpty) None
```
sritchie, line 151: Same question here, etc

.travis.yml
regadas: Hi @johnynek, there are some changes around the `CanBuildFrom` that will trigger `IncompatibleSignatureProblem` and `IncompatibleMethTypeProblem`; i.e [here](https://github.com/twitter/algebird/pull/719/files#diff-cbe57544ccef25e3302c4ec7ae13a3b7R444) and [here](https://github.com/twitter/algebird/pull/719/files#diff-798d0508920b1369bf591916e39019a8R110)

We also get `MissingClassProblem` because I had to relocate `com.twitter.algebird.MutableBackedMap`

algebird-core/src/main/scala/com/twitter/algebird/Aggregator.scala
regadas: Hi @sritchie, yeah this one got me by surprise as well; so with 2.13 `TraversableOnce` is deprecated and synonymous to `IterableOnce`; `isEmpty` is deprecated in favor of `.iterator.isEmpty`.

With this there's no collection modification, there is though an extra invocation that jvm inlining should™ take care of. Also AFAIK if a collection has a specific impl of `isEmpty` that is used by the `Iterator`.

.travis.yml
regadas: Actually, regarding the `com.twitter.algebird.MutableBackedMap` we can actually have it in the same namespace ... it will not be a prob.

johnynek: I think I'd rather take the 2.12-/2.13+ strategy used here: https://github.com/typelevel/paiges/pull/189 -- this code doesn't change a ton, so I think it shouldn't be too much of a pain.

I'm a big believer in compatibility.

regadas: Yup I agree ... I like the ` 2.12-/2.13+`, less copying 😄

algebird-core/src/main/scala-2.13+/com/twitter/algebird/compat.scala
```diff
@@ -0,0 +1,38 @@
...
+  /**
+   * Simple Fold that collects elements into a container.
+   */
+  def container[I, C[_]](implicit cbf: Factory[I, C[I]]): Fold[I, C[I]] =
+    Fold.foldMutable[Builder[I, C[I]], I, C[I]]({ case (b, i) => b += i }, { _ =>
+      cbf.newBuilder
+    }, { _.result })
+}
+
+private[algebird] trait CompatDecayedVector {
```
johnynek, line 29: Why can't this be `monoid[F, C[_]]` in 2.13? Can you comment?

Also, I see no reason to take an implicit `Ordering[Double]`. Shouldn't we just use the default one?

algebird-core/src/main/scala/com/twitter/algebird/HyperLogLogSeries.scala
```diff
@@ -92,9 +94,9 @@ case class HLLSeries(bits: Int, rows: Vector[Map[Int, Long]]) {
     else {
       monoid.sum(rows.iterator.zipWithIndex.map {
         case (map, i) =>
-          SparseHLL(bits, map.mapValues { _ =>
+          SparseHLL(bits, map.view.mapValues { _ =>
```
johnynek: maybe just `map.iterator.map { case (k, _) => (k, Max((i + 1).toByte)) }.toMap`?

algebird-core/src/main/scala/com/twitter/algebird/MapAlgebra.scala
```diff
@@ -115,33 +115,23 @@ class ScMapMonoid[K, V](implicit semigroup: Semigroup[V]) extends GenericMapMono
...
-
 /**
  * You can think of this as a Sparse vector group
  */
 class MapGroup[K, V](implicit val group: Group[V]) extends MapMonoid[K, V]()(group) with Group[Map[K, V]] {
-  override def negate(kv: Map[K, V]): Map[K, V] = kv.mapValues { v =>
-    group.negate(v)
-  }
+  override def negate(kv: Map[K, V]): Map[K, V] =
+    kv.mapValues { v =>
```
johnynek: can we just do `kv.iterator.map { ... }.toMap` so people don't have to have the `mapValues` anxiety when they read the code?

algebird-core/src/main/scala/com/twitter/algebird/MapAlgebra.scala
```diff
@@ -115,33 +115,23 @@ class ScMapMonoid[K, V](implicit semigroup: Semigroup[V]) extends GenericMapMono
...
 }
 
 class ScMapGroup[K, V](implicit val group: Group[V])
     extends ScMapMonoid[K, V]()(group)
     with Group[ScMap[K, V]] {
-  override def negate(kv: ScMap[K, V]): ScMap[K, V] = kv.mapValues { v =>
-    group.negate(v)
-  }
+  override def negate(kv: ScMap[K, V]): ScMap[K, V] =
+    kv.mapValues { v =>
```
johnynek: same comment about killing mapValues.

algebird-core/src/main/scala/com/twitter/algebird/MapAlgebra.scala
```diff
@@ -252,10 +242,11 @@ object MapAlgebra {
     Monoid
       .plus(map1.mapValues { v =>
         (List(v), List[W]())
-      }, map2.mapValues { w =>
+      }.toMap, map2.mapValues { w =>
```
johnynek: can we nuke these mapValues?

algebird-core/src/main/scala/com/twitter/algebird/VectorSpace.scala
```diff
@@ -43,7 +43,7 @@ object VectorSpace {
 
   implicit def mapSpace[K, T: Ring] =
     from[T, ({ type x[a] = Map[K, a] })#x] { (s, m) =>
-      m.mapValues(Ring.times(s, _))
+      m.mapValues(Ring.times(s, _)).toMap
```
johnynek: can we remove the mapValues?

algebird-core/src/main/scala/com/twitter/algebird/macros/Cuber.scala
```diff
@@ -68,7 +68,7 @@ object Cuber {
     val cuber = q"""
     new _root_.com.twitter.algebird.macros.Cuber[${T}] {
       type K = $tupleName
-      def apply(in: ${T}): _root_.scala.Seq[K] = {
+      def apply(in: ${T}): _root_.scala.collection.Seq[K] = {
```
johnynek: this is a bummer that it isn't `collection.immutable`...

algebird-core/src/main/scala/com/twitter/algebird/package.scala
```diff
@@ -17,6 +17,8 @@
 package com.twitter
 
 package object algebird {
+  type Seq[+T] = scala.collection.Seq[T]
+  type IndexedSeq[+T] = scala.collection.IndexedSeq[T]
```
johnynek: is there no way we can't use `collection.immutable` here? Why do we need these aliases?

johnynek (commented):
thanks for doing this!

It will be great to publish for 2.13.

algebird-core/src/main/scala/com/twitter/algebird/package.scala
regadas: yeah good point we can, shadowing was my first thought, since predef.Seq in scala 2.13 changed. But we can definitely remove this.

algebird-core/src/main/scala/com/twitter/algebird/HyperLogLogSeries.scala
regadas: 👍

algebird-core/src/main/scala/com/twitter/algebird/MapAlgebra.scala
regadas: 👍

algebird-core/src/main/scala-2.13+/com/twitter/algebird/compat.scala
regadas: hm, `monoid[F, C[_]]` will fail to compile due to ambiguous `VectorSpace`:
```
[error]  both method indexedSeqSpace in trait SeqInstances of type [T](implicit evidence$1: com.twitter.algebird.Ring[T])com.twitter.algebird.VectorSpace[T,IndexedSeq]
[error]  and method mapSpace in trait MapInstances of type [K, T](implicit evidence$2: com.twitter.algebird.Ring[T])com.twitter.algebird.VectorSpace[T,[a]scala.collection.immutable.Map[K,a]]
[error]  match expected type com.twitter.algebird.VectorSpace[Double,C]
[error]     DecayedVector.monoidWithEpsilon(-1.0)
```

https://github.com/twitter/algebird/blob/d17ba8978b9526cdf6c97e3dfae273ea6a501097/algebird-core/src/main/scala/com/twitter/algebird/VectorSpace.scala#L39-L47

With 2.13 we can make it work if we make one of them low prio.

Regarding Ordering, yup we can remove it in scala 2.13 I just carried it over;

regadas: Update: Low prio alternate solution 01cfc4b

algebird-core/src/main/scala/com/twitter/algebird/Fold.scala
```diff
@@ -245,13 +244,13 @@ object Fold {
...
     }
 
-  /**
-   * Simple Fold that collects elements into a container.
-   */
-  def container[I, C[_]](implicit cbf: CanBuildFrom[C[I], I, C[I]]): Fold[I, C[I]] =
-    Fold.foldMutable[Builder[I, C[I]], I, C[I]]({ case (b, i) => b += i }, { _ =>
-      cbf.apply
-    }, { _.result })
+  // /**
```
johnynek: can we remove the commented code?

algebird-core/src/main/scala/com/twitter/algebird/HyperLogLog.scala
```diff
@@ -463,7 +464,7 @@ case class SparseHLL(override val bits: Int, maxRhow: Map[Int, Max[Byte]]) exten
         val newRhoW = reducedMaxRhoW.getOrElse(newJ, 0: Byte)
         reducedMaxRhoW += (newJ -> (newRhoW.max(modifiedRhoW)))
     }
-    SparseHLL(reducedBits, reducedMaxRhoW.toMap.mapValues(Max(_)))
+    SparseHLL(reducedBits, reducedMaxRhoW.toMap.view.mapValues(Max(_)).toMap)
```
johnynek: can we do `reducedMaxRhoW.iterator.map { case (k, v) => (k, Max(v)) }.toMap` here? looks like we have two conversions to map here.

johnynek (commented):
getting very close. If are are sure mima works, I'm happy to merge this.

algebird-core/src/main/scala/com/twitter/algebird/Fold.scala
regadas: Oops

algebird-core/src/main/scala/com/twitter/algebird/HyperLogLog.scala
regadas: ahh I missed this one 🤦‍♂

regadas:
@johnynek well, I believe MiMa is working properly; we are using the latest MiMa and it's a bit more "picky"; I think that if we accidentally broke compatibility MiMa would be complaining.

note: we are only running MiMa for 2.11 and 2.12. We can add for 2.13 once we release;

johnynek (approved):

regadas:
@johnynek Thanks for the review! 👍

PR #720, opened 2019-10-09
Add jdk11 CI jobs

regadas:

PR #726, opened 2019-10-10
Fix CI build

regadas:
Apologies but I kinda broke the cross build in #720; I thought that aggregating the commands like `++SCALA_VERSION ;cmd; cmd` would work but it turns out that it doesn't as expected.

Also, since sbt 1.3.2 uses coursier, we need to add the dependencies path to the cache.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,28 +5,38 @@ matrix:
     - name: checks
       scala: 2.11.12
       jdk: openjdk8
-      script: sbt ++$TRAVIS_SCALA_VERSION \
+      script: sbt \
         "; scalafmtCheckAll; scalafmtSbtCheck" \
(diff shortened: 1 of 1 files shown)


nevillelyh:
Just to clarify, what we've been using in most Travis builds, e.g.:

`sbt ++$TRAVIS_SCALA_VERSION clean compile test`

no longer works and we need to use

`sbt ++$TRAVIS_SCALA_VERSION clean ++$TRAVIS_SCALA_VERSION compile ...`?

regadas:
> sbt ++$TRAVIS_SCALA_VERSION clean ++$TRAVIS_SCALA_VERSION compile ...

Hi @nevillelyh, yup this is the correct way of doing it.

johnynek:
thanks

PR #727, opened 2019-10-10
Make scala 2.12 default

regadas:
2.12 is now the highest common version to all the aggregated projects.

Also, removed the usage of different version of `twitter-util`.


regadas:
@johnynek any thoughts on this?

johnynek:
sorry for the latency.

regadas:
@johnynek no worries! 😄

PR #734, opened 2019-10-22
Update README

regadas:
Seems also that scaladex is outdated; it's still pointing to `0.13.4`; created https://github.com/scalacenter/scaladex/issues/579 but  as an alternative we can use https://maven-badges.herokuapp.com/maven-central/com.twitter/algebird-core_2.12


johnynek (approved):
Thanks a lot!

PR #739, opened 2019-11-06
new algebird abstraction: Scan

jeff-stripe:
A new abstraction, similar to `Aggregator` or `Fold`, that aims to reify the business logic assocaited with `scanLeft` into its own abstraction with its own combinators.

The testing here is pretty comprehensive, although I don't think `.zip` is covered. Moreover, I originally had a bunch of combinators named `.zipWithXXX` that I later renamed to `.joinWithXXX` to make more consistent with the types of `zip` and `join`. I do think it's worth bikeshedding naming on all of this though.

For posterity, my non-work github is @eigenvariable.

### DIFF
--- a/algebird-core/src/main/scala/com/twitter/algebird/Scan.scala
+++ b/algebird-core/src/main/scala/com/twitter/algebird/Scan.scala
@@ -126,7 +126,7 @@ object Scan {
  */
 sealed trait Scan[-I, +O] extends Serializable { self =>
 
-  import Scan.{Aux, from}
+  import Scan.{from, Aux}
 
   /**
    * The computation of any given scan involves keeping track of a hidden state.
@@ -186,41 +186,31 @@ sealed trait Scan[-I, +O] extends Serializable { self =>
 
   // combinators
 
-  def replaceState(newInitialState: => State): Aux[I, State, O] = from(newInitialState)(presentAndNextState(_, _))
+  def replaceState(newInitialState: => State): Aux[I, State, O] =
+    from(newInitialState)(presentAndNextState(_, _))
 
-  def composePrepare[I1](f: I1 => I): Aux[I1, State, O] = from(self.initialState){(i, stateBeforeProcessingI) =>
-      presentAndNextState(f(i), stateBeforeProcessingI)
+  def composePrepare[I1](f: I1 => I): Aux[I1, State, O] = from(initialState) { (i, stateBeforeProcessingI) =>
+    presentAndNextState(f(i), stateBeforeProcessingI)
   }
 
-  def andThenPresent[O1](g: O => O1): Aux[I, State, O1] = new Scan[I, O1] {
-    override type State = self.State
-    override def initialState: State = self.initialState
-
-    override def presentAndNextState(i: I, stateBeforeProcessingI: State): (O1, State) = {
-      val (c, stateAfterProcessingA) = self.presentAndNextState(i, stateBeforeProcessingI)
-      (g(c), stateAfterProcessingA)
-    }
+  def andThenPresent[O1](g: O => O1): Aux[I, State, O1] = from(initialState) { (i, stateBeforeProcessingI) =>
+    val (c, stateAfterProcessingA) = presentAndNextState(i, stateBeforeProcessingI)
+    (g(c), stateAfterProcessingA)
   }
 
   /**
    *
    * @tparam I1
-   * @return If this Scan's `apply` method is given inputs [a_1, ..., a_n] resulting in outputs
+   * @return If this Scan's `apply` method is given inputs `[a_1, ..., a_n]` resulting in outputs
    * of the form [o_1, ..., o_n], then {{joinWithInput}} results in a Scan whose `apply` method
    * returns [(a_1, o_1), ..., (a_n, o_n)] when given the same input.
-   * In other words, {{joinWithInput}} returns a scanner that is semantically identical to
-   * {{this.join(Scan.identity[I1])}}, but where we don't pollute the {{State}} by pairing it
-   * redundantly with {{Unit}}.
+   * In other words, `joinWithInput` returns a scanner that is semantically identical to
+   * `this.join(Scan.identity[I1]`, but where we don't pollute the `State` by pairing it
+   * redundantly with `Unit`.
    */
-  def joinWithInput[I1 <: I]: Aux[I1, State, (I1, O)] = new Scan[I1, (I1, O)] {
-    override type State = self.State
-
-    override def initialState: State = self.initialState
-
-    override def presentAndNextState(i: I1, stateBeforeProcessingI: State): ((I1, O), State) = {
-      val (o, stateAfterProcessingA) = self.presentAndNextState(i, stateBeforeProcessingI)
-      ((i, o), stateAfterProcessingA)
-    }
+  def joinWithInput[I1 <: I]: Aux[I1, State, (I1, O)] = from(initialState) { (i, stateBeforeProcessingI) =>
+    val (o, stateAfterProcessingI) = self.presentAndNextState(i, stateBeforeProcessingI)
+    ((i, o), stateAfterProcessingI)
   }
 
   /**
@@ -230,15 +220,9 @@ sealed trait Scan[-I, +O] extends Serializable { self =>
    * @return A scan that whose apply method, when given inputs [a_1, ..., a_n] will return
    * [(o_1, state_0), ..., (o_n, state_(n-1))].
    */
-  def joinWithPriorState: Aux[I, State, (State, O)] = new Scan[I, (State, O)] {
-    override type State = self.State
-
-    override def initialState: State = self.initialState
-
-    override def presentAndNextState(i: I, stateBeforeProcessingI: State): ((State, O), State) = {
-      val (o, stateAfterProcessingA) = self.presentAndNextState(i, stateBeforeProcessingI)
-      ((stateBeforeProcessingI, o), stateAfterProcessingA)
-    }
+  def joinWithPriorState: Aux[I, State, (State, O)] = from(initialState) { (i, stateBeforeProcessingI) =>
+    val (o, stateAfterProcessingA) = self.presentAndNextState(i, stateBeforeProcessingI)
+    ((stateBeforeProcessingI, o), stateAfterProcessingA)
   }
 
   /**
@@ -248,15 +232,9 @@ sealed trait Scan[-I, +O] extends Serializable { self =>
    * @return A scan that whose apply method, when given inputs [a_1, ..., a_n] will return
    * [(o_1, state_1), ..., (o_n, state_n].
    */
-  def joinWithPosteriorState: Aux[I, State, (O, State)] = new Scan[I, (O, State)] {
-    override type State = self.State
-
-    override def initialState: State = self.initialState
-
-    override def presentAndNextState(i: I, stateBeforeProcessingI: State): ((O, State), State) = {
-      val (c, stateAfterProcessingA) = self.presentAndNextState(i, stateBeforeProcessingI)
-      ((c, stateAfterProcessingA), stateAfterProcessingA)
-    }
+  def joinWithPosteriorState: Aux[I, State, (O, State)] = from(initialState) { (i, stateBeforeProcessingI) =>
+    val (c, stateAfterProcessingA) = self.presentAndNextState(i, stateBeforeProcessingI)
+    ((c, stateAfterProcessingA), stateAfterProcessingA)
   }
 
   /**
@@ -279,22 +257,12 @@ sealed trait Scan[-I, +O] extends Serializable { self =>
    * In other words: `scan.zip(scan2)(foo.zip(bar)) == scan(foo).zip(scan2(bar)) `
    */
   def zip[I2, O2](scan2: Scan[I2, O2]): Aux[(I, I2), (State, scan2.State), (O, O2)] =
-    new Scan[(I, I2), (O, O2)] {
-      override type State = (self.State, scan2.State)
-
-      override def initialState: State = (self.initialState, scan2.initialState)
-
-      override def presentAndNextState(
-          i1i2: (I, I2),
-          stateBeforeProcessingI1I2: State
-      ): ((O, O2), State) = {
-        val (o1, state1AfterProcesingI1) =
-          self.presentAndNextState(i1i2._1, stateBeforeProcessingI1I2._1)
-        val (o2, state2AfterProcesingI2) =
-          scan2.presentAndNextState(i1i2._2, stateBeforeProcessingI1I2._2)
-        ((o1, o2), (state1AfterProcesingI1, state2AfterProcesingI2))
-
-      }
+    from((self.initialState, scan2.initialState)) { (i1i2, stateBeforeProcessingI1I2) =>
+      val (o1, state1AfterProcesingI1) =
+        self.presentAndNextState(i1i2._1, stateBeforeProcessingI1I2._1)
+      val (o2, state2AfterProcesingI2) =
+        scan2.presentAndNextState(i1i2._2, stateBeforeProcessingI1I2._2)
+      ((o1, o2), (state1AfterProcesingI1, state2AfterProcesingI2))
     }
 
   /**
@@ -306,17 +274,12 @@ sealed trait Scan[-I, +O] extends Serializable { self =>
    * `join` will return a scan whose apply method returns [(o_1, p_1), ..., (o_2, p_2)].
    * In other words: `scan.join(scan2)(foo) == scan(foo).zip(scan2(foo)) `
    */
-  def join[I2 <: I, O2](scan2: Scan[I2, O2]): Aux[I2, (State, scan2.State), (O, O2)] = new Scan[I2, (O, O2)] {
-    override type State = (self.State, scan2.State)
-
-    override def initialState: State = (self.initialState, scan2.initialState)
-
-    override def presentAndNextState(i: I2, stateBeforeProcessingI: State): ((O, O2), State) = {
+  def join[I2 <: I, O2](scan2: Scan[I2, O2]): Aux[I2, (State, scan2.State), (O, O2)] =
+    from((self.initialState, scan2.initialState)) { (i, stateBeforeProcessingI) =>
       val (o1, state1AfterProcesingI1) = self.presentAndNextState(i, stateBeforeProcessingI._1)
       val (o2, state2AfterProcesingI2) = scan2.presentAndNextState(i, stateBeforeProcessingI._2)
       ((o1, o2), (state1AfterProcesingI1, state2AfterProcesingI2))
     }
-  }
 
   /**
    * Takes the output of this scan and feeds as input into scan2.
@@ -326,17 +289,12 @@ sealed trait Scan[-I, +O] extends Serializable { self =>
    * the form [o_1, ..., o_n], and scan2.apply([o_1, ..., o_n] = [p_1, ..., p_n] then
    * `compose` will return a scan which returns [p_1, ..., p_n].
    */
-  def compose[P](scan2: Scan[O, P]): Aux[I, (State, scan2.State), P] = new Scan[I, P] {
-    override type State = (self.State, scan2.State)
-
-    override def initialState: State = (self.initialState, scan2.initialState)
-
-    override def presentAndNextState(i: I, stateBeforeProcessingI: State): (P, State) = {
-      val (o, state1AfterProcesingI1) = self.presentAndNextState(i, stateBeforeProcessingI._1)
-      val (p, state2AfterProcesingI2) = scan2.presentAndNextState(o, stateBeforeProcessingI._2)
-      (p, (state1AfterProcesingI1, state2AfterProcesingI2))
+  def compose[P](scan2: Scan[O, P]): Aux[I, (State, scan2.State), P] =
+    from((self.initialState, scan2.initialState)) { (i, stateBeforeProcessingI) =>
+      val (o, state1AfterProcesingI) = presentAndNextState(i, stateBeforeProcessingI._1)
+      val (p, state2AfterProcesingO) = scan2.presentAndNextState(o, stateBeforeProcessingI._2)
+      (p, (state1AfterProcesingI, state2AfterProcesingO))
     }
-  }
 
 }
 
--- a/algebird-core/src/main/scala/com/twitter/algebird/Scan.scala
+++ b/algebird-core/src/main/scala/com/twitter/algebird/Scan.scala
@@ -264,8 +264,12 @@ sealed trait Scan[-I, +O] extends Serializable { self =>
   /**
    *
    * @tparam I1
-   * @return A scanner that is semantically identical to .join(Scan.identity[I1]), but without
-   *         where we don't pollute the {{State}} by pairing it redundantly with {{Unit}}.
+   * @return If this Scan's {{apply}} method is given inputs [a_1, ..., a_n] resulting in outputs
+   * of the form [o_1, ..., o_n], then {{joinWithInput}} results in a Scan whose {{apply}} method
+   * returns [(a_1, o_1), ..., (a_n, o_n)] when given the same input.
+   * In other words, {{joinWithInput}} returns a scanner that is semantically identical to
+   * {{this.join(Scan.identity[I1])}}, but where we don't pollute the {{State}} by pairing it
+   * redundantly with {{Unit}}.
    */
   def joinWithInput[I1 <: I]: Aux[I1, State, (I1, O)] = new Scan[I1, (I1, O)] {
     override type State = self.State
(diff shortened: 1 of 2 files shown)


algebird-core/src/main/scala/com/twitter/algebird/Scan.scala
```diff
@@ -0,0 +1,361 @@
...
+  /**
+    * Most consumers of Scan don't care about the type of the {{State}} type variable. But for those that do,
+    * we make an effort to expose it in all of our combinators.
+    * @tparam I
+    * @tparam S
+    * @tparam O
+    */
+  type Aux[-I, S, +O] = Scan[I, O] { type State = S }
+
+  implicit def applicative[I]: Applicative[({ type L[O] = Scan[I, O] })#L] = new ScanApplicative[I]
```
MansurAshraf, line 16: you can use Kind projector here?

algebird-core/src/main/scala/com/twitter/algebird/Scan.scala
```diff
@@ -0,0 +1,361 @@
+package com.twitter.algebird
+
+import scala.collection.compat._
+import scala.collection.generic.CanBuildFrom
+
+object Scan {
```
MansurAshraf, line 5: How is this different than ` Aggregator[I,M[_]:SemigroupK, O]` . I think the motivation here is that you want `S` type to be a container so that `reduce` on `S` is collecting intermediate state. I believe that can be achieved by existing Aggregators by 1) making S a type constructor 2) making sure that S has an instance of [`SemigroupK`](https://typelevel.org/cats/typeclasses/semigroupk.html)

jeff-stripe: @MansurAshraf First: great to be on a PR with you again! It's been a long time :-).

Second: the idea here is that this transforms streams to streams. whereas aggregators transform Sequences to single values. I would look at the `apply` method (starting on line 229 of `Stream.scala` as of this writing). The idea is that, for each input in the stream, you have one output in the stream. For example, using `directFreeScan` from `ScanTest`, something like `directFreeScan(List('a', 'b', 'c'))` would result in `List("a", "ab", "abc")`.

Does that make sense?

MansurAshraf: hey @jeff-stripe , its been a while :-)

I will take a look at the Stream.scala but cursory look at `def fromAggregator[A, B, C](aggregator: Aggregator[A, B, C], initState: B): Aux[A, B, C]` gives me the impression that

```Scan[A,C] ~= Aggregator[A,S[_],(S[_],C)]```

Scan has a new function called `presentAndNextState` which return both S state type and final output type C. If you aggregator type was `Aggregator[A,S[_],(S[_],C)]`, the `present` function in existing aggregator will return the same value i think.

**Comment by johnynek** [comment]
Thanks jeff.

CI failed:

1. code formatting, I think running `scalafmtAll`
2. the 2.13 build fails. We need to make a minor change to the CanBuildFrom stuff. I don't know the answer, but 2.13 `BuildFrom` which the compat stuff is aliasing is here: https://www.scala-lang.org/api/2.13.1/scala/collection/BuildFrom.html

File: algebird-core/src/main/scala/com/twitter/algebird/Scan.scala (hunk 2 above)
Comment (jeff-stripe) [reply algebird-core/src/main/scala/com/twitter/algebird/Scan.scala:5]: @MansurAshraf In my example, the analogous `Aggregator` would just return `"abc"`, not `List("a", "ab", "abc")`. In other words, `Aggregator` is to the `reduce` method as `Scan` is to the `scanLeft` method.

Comment (johnynek) [reply algebird-core/src/main/scala/com/twitter/algebird/Scan.scala:5]: Mansur, to a first approximation, `Aggregator[A, _, B]` is `Iterator[A] => Option[B]`, `MonoidAggregator[A, _, B]` is `Iterator[A] => B` and `Scan[A, B]` is `Iterator[A] => Iterator[B]` (actually with an added law that the number of items in the result is exactly the same.

You could imagine another idea: `ExpandingScan[A, B]` that is `Iterator[A] => Iterator[B]` but without the law of the iterators being the same size (then we could do things like filter, or concatMap, etc...)

Does that help explain it at all?

Think of the `ExpandingScan` as the most general function you can do in a reducer in Hadoop.

File: algebird-core/src/main/scala/com/twitter/algebird/Scan.scala (hunk 1 above)
Comment (johnynek) [reply algebird-core/src/main/scala/com/twitter/algebird/Scan.scala:16]: I don't think we use that plugin, we could, but I'd rather not add a build dependency for the few cases where we have done this. This is mostly a demonstration anyway, since I expect if you really want a `Applicative[Scan[I, ?]]` you will want a cats implementation, but anyway, this shows you how to make one.

File: algebird-core/src/main/scala/com/twitter/algebird/Scan.scala [hunk 3]
```diff
@@ -0,0 +1,361 @@
...
+    override type State = S
+    override val initialState = initState
+    override def presentAndNextState(i: Any, stateBeforeProcessingI: S): (O, S) =
+      destructor(stateBeforeProcessingI)
+  }
+
+  /**
+    * A Scan that returns the number N for the Nth input (starting from 0)
+    */
+  val index: Aux[Any, Long, Long] = fromStreamLike(0L)(n => (n, n+1))
```
johnynek: this is really cool.

algebird-core/src/main/scala/com/twitter/algebird/Scan.scala
```diff
@@ -0,0 +1,361 @@
...
+    * that represents the rest of the stream. Scans take streams of inputs to streams of outputs, but some scans
+    * have trivial inputs and just produce a stream of outputs.
+    * @param initState The initial state of the scan; think of this as an infinite stream.
+    * @param destructor This function decomposes a stream into the its head-element and tail-stream.
+    * @tparam S The hidden state of the stream that we are turning into a Scan.
+    * @tparam O The type of the elments of the stream that we are turning into a Scan
+    * @return A Scan whose inputs are irrelevant, and whose outputs are those that we would get from implementing
+    *         a stream using the information provided to this method.
+    */
+  def fromStreamLike[S, O](initState: S)(destructor: S => (O, S)): Aux[Any, S, O] = new Scan[Any, O] {
```
johnynek: can we call this `def iterate`? This is very similar to `Iterator.iterate` or `Stream.iterate` in the scala standard library.

This is really awesome.

algebird-core/src/main/scala/com/twitter/algebird/Scan.scala
```diff
@@ -0,0 +1,361 @@
...
+   *
+   * @param aggregator
+   * @param initState
+   * @tparam A
+   * @tparam B
+   * @tparam C
+   * @return  A scan which, when given [a_1, ..., a_n] outputs [c_1, ..., c_n] where
+   *         c_i = aggregator.prepare(a_i) + ... + aggregator.prepare(a_1) + initState
+   */
+  def fromAggregatorReverse[A, B, C](aggregator: Aggregator[A, B, C], initState: B): Aux[A, B, C] =
```
johnynek: to me this seems like it should be a combinator on `Aggregator` or really `Semigroup` and not here. Like `def reverse` method on Aggregator.

algebird-core/src/main/scala/com/twitter/algebird/Scan.scala
```diff
@@ -0,0 +1,361 @@
...
+  /**
+   *
+   * @param monoidAggregator
+   * @tparam A
+   * @tparam B
+   * @tparam C
+   * @return  A scan which, when given [a_1, ..., a_n] outputs [c_1, ..., c_n] where
+   *         c_i = aggregator.prepare(a_i) + ... + aggregator.prepare(a_1) + monoidAggregator.monoid.zero
+   */
+  def fromMonoidAggregatorReverse[A, B, C](
```
johnynek: same comment about adding a `def reverse` method to `MonoidAggregator`

algebird-core/src/main/scala/com/twitter/algebird/Scan.scala
```diff
@@ -0,0 +1,361 @@
...
+
+    override def initialState: State = self.initialState
+
+    override def presentAndNextState(i: I, stateBeforeProcessingI: State): ((O, State), State) = {
+      val (c, stateAfterProcessingA) = self.presentAndNextState(i, stateBeforeProcessingI)
+      ((c, stateAfterProcessingA), stateAfterProcessingA)
+    }
+  }
+
+  def joinWithIndex: Aux[I, (State, Long), (O, Long)] = join(Scan.index)
```
johnynek, line 255: this show the compositional power so well!

algebird-core/src/main/scala/com/twitter/algebird/Scan.scala
```diff
@@ -0,0 +1,361 @@
...
+    override def initialState: State = (self.initialState, scan2.initialState)
+
+    override def presentAndNextState(i: I2, stateBeforeProcessingI: State): ((O, O2), State) = {
+      val (o1, state1AfterProcesingI1) = self.presentAndNextState(i, stateBeforeProcessingI._1)
+      val (o2, state2AfterProcesingI2) = scan2.presentAndNextState(i, stateBeforeProcessingI._2)
+      ((o1, o2), (state1AfterProcesingI1, state2AfterProcesingI2))
+    }
+  }
+
+  def compose[P](scan2: Scan[O, P]): Aux[I, (State, scan2.State), P] = new Scan[I, P] {
```
johnynek: maybe add comments to the methods: `this feeds each output of this into scan2` and join is like `this forks each input and feeds it into both this and scan2`, etc...

jeff-stripe:
Thanks @johnynek for the comments; I'm running out of steam for the day but will pick this up again tomorrow.

johnynek:
a really cool example `Scan[Double, Double]` would be running z-score. You keep a state of `Moments` and you use then joinWithPosteriorState, and then normalize the `Double` against the mean and the std-dev.

... just thinking aloud, that's not a request to add that.

johnynek:
also, there is a (not very good) streaming median estimator where you keep track of the current median, and if the value is greater than that, you move it up by a constant, and if it is less than that, you move it down. Since the true median should have about half above as below, this should be a random walk that biases towards the median (so after a long enough warmup, it shouldn't be too far from the median).

johnynek:
https://codecov.io/gh/twitter/algebird/pull/739/diff?src=pr&el=tree#diff-YWxnZWJpcmQtY29yZS9zcmMvbWFpbi9zY2FsYS9jb20vdHdpdHRlci9hbGdlYmlyZC9TY2FuLnNjYWxh 

looks like `zip` is untested as is `joinWithIndex` Can I talk you into adding tests for those?

algebird-core/src/main/scala/com/twitter/algebird/Scan.scala
jeff-stripe: I'd like to push back on this a little bit. It's not a-priori obvious to me from the signature of this method whether the state is calculated by adding on the right or adding on the left (in essence doing a `foldRight` or a `foldLeft` to calculate the state of each element of the stream). From my perspective, and having both methods available calls out the ambiguity here, forcing the caller to think about which semantics they prefer, assuming they care.

That said, maybe the names `fromAggregator` and `fromAggregatorReverse` show too much bias on my part for the semantics of one side over the other. How would you feel about me renaming them to `fromAggregatorRight` and `fromAggregatorLeft` or something along those lines?

johnynek: My concern is that Aggregator already has order semantics. You may not like that, but it does: order matters on `apply` and `applyOption` on `Aggregator`. So, trying to change that in a related abstraction seems inconsistent to me. I'm concerned that we have already crossed this bridge, and I think it is better to simply be consistent now.

Next: I'm skeptical this API is actually very commonly called in any case. Would a compromise be to just remove it and not make any other changes? Adding two combinators on Semigroup:

```scala
def intercalate[A](sg: Semigroup[A], middle: A): Semigroup[A] =
  new Semigroup[A] {
    def plus(left: A, right: A) = sg.plus(sg.plus(a, middle), b)
  }

def reverse[A](sg: Semigroup[A]): Semigroup[A] =
  new Semigroup[A] {
    def plus(left: A, right: A) = sg.plus(right, left)
  }
```

is something we can do or not (and the implications for Aggregator/MonoidAggregator) in a separate PR.

jeff-stripe: This is a good point, and seems reasonable enough to me.

Irrelevant to this PR, but relevant to the broader discussion: Are you worried about the performance impact of a `reverse` method on `Semigroup`s essentially destroying any custom `SumOption` methods?

johnynek: I am worried about that. `sumOption` composes terribly, but one thing we do on products is materialize 1000 groups of items, then do sumOption after materialization. We could do the same thing with reverse: materialize an group of 1000, call underlying sumOption, repeat.

jeff-stripe: Seems like a reasonable workaround, although now you're materializing 1000 items even for monoids that don't have a custom `sumOption` which might have weird memory impact. But, to your point, `sumOption` composes badly and that's probably the right tradeoff. And, even more to the point, this is all outside the scope of this PR.

In the meantime: I've deleted the `reverse` methods.

jeff-stripe:
@johnynek I _think_ I've addressed all of your comments so far (but welcome more). Also, the example of z-score was totally one original motivation for this abstraction, from way back when I was working on feature generation.

algebird-core/src/main/scala/com/twitter/algebird/Scan.scala
jeff-stripe: I've added a bunch of comments. Let me know if you like the wording/framing of them.

algebird-test/src/test/scala/com/twitter/algebird/ScanTest.scala
```diff
@@ -66,7 +56,7 @@ object ScanTest {
 
 }
 
-class ScanTest extends WordSpec with Matchers with GeneratorDrivenPropertyChecks {
+class ScanTest extends WordSpec with Matchers with ScalaCheckDrivenPropertyChecks {
```
johnynek, line 59: are these tests actually running? the code coverage now things you have zero coverage.

I wonder if there is something about the new scalatest that is broken here.

If you break your code and do `test` locally, does it still pass?

jeff-stripe: @johnynek  You're right that the tests weren't running (embarrassing!), and that they broke on this commit. But it wasn't because I was using the more up-to-date-import, it was worse than that: when I wrote a method for the new properties for the tests on that commit, a missing close-curly accidentally made the entire test suite exist inside of the method, which was then never being executed (even more embarrassing!).

Is there a way to force CI to give a code coverage report on an adhoc basis, or should I just wait until it auto-runs?

jeff-stripe:
Unrelated: when I was running tests locally to reproduce the above issue, I ran into the following unrelated test failure:

```
[info] DecayedVectorProperties:
[info] - DecayedVector[Map[Int, _]] is a monoid *** FAILED ***
[info]     Falsified after 9 successful property evaluations.
[info]       arg0 = DecayedVector(Map(0 -> 1.2018386197212909E-281, -2147483648 -> -3.854156922919691E232, -1447110768 -> 9.408465891492915E-167, 1805048232 -> -6.75705880126906E215, 9346750 -> 1.0723063735502226E-173, -420323548 -> -8.302534164574593E-175, -571797472 -> -2.179474851477724E-175, -470808363 -> 3.3072194967626795E-250, 1 -> 2.1209952984836344E229, 2076284047 -> -6.199396905341707E-123, 365574431 -> 1.5142075318601491E184, -1752737747 -> -7.883369021177454E29, 2050082858 -> -2.299358981334486E-15, -1994605430 -> -7.431824108415331E217, -698948772 -> -2.1479418657592524E-191, -1417251516 -> -1.5572134571317626E-254, 241933112 -> 545733.1130589077, 897298981 -> -7.439164205119313E10, -654320593 -> 3.2544666904527896E-292, 887471898 -> 4.069738732277512E125, 373768514 -> 3.432851429494389E63, -932811734 -> -2.0244449228411652E-74, 783480879 -> -8.136720367399272E-279, -1586891494 -> 2.7376160145598414E141, 2147483647 -> -8.574297221458591E-159, -1 -> 1.2104690607925745E246, -1468444249 -> -1.86707430482823E-307, -1009301793 -> 5.918573113974547E148, -1013689345 -> 1.2825959305194982E-110, -1791275591 -> 5.587605356252549E79, 401213385 -> -4.862900034659543E190, -738881469 -> -5.061788993603661E-14, -1913217183 -> 1.7586395366496495E133, -2087846791 -> 7.189097244947145E115, 630320687 -> -4.305922779804356E-304, -1994015926 -> -3.4956222225235067E103, 262178902 -> -1.1507181102048183E43, -612977120 -> 4.0867877449600855E-129, 1761853732 -> -7.897508692485825E301, -1861380341 -> -1.7935785304794388E-177, -1862146545 -> -4.7832258685347225E-94, -1742740609 -> -1.4002257491338442E151, 584865158 -> -9676137.866416803, 349932431 -> -1.0699110976306946E-14),185.95991254209034),
[info]       arg1 = DecayedVector(Map(0 -> -1.0830365240639644E-300, -2147483648 -> -2.520348300233662E-155, -1913741494 -> -2.8893542841978695E-283, -69242322 -> 1.9351525683506846E90, 1508963855 -> 1.44442040727596E-95, 1113162072 -> 9.523002151023972E42, 1 -> -1.0700938964522293E-234, 749585887 -> 4.4074563848912957E-303, 816327294 -> 7.404952094444072E-272, 1539634748 -> 2.429738132317924E110, 1201648438 -> 1.2318638393727931E41, 803892976 -> -2.1408844799849434E42, 132589461 -> -6.724616862088172E237, -426808141 -> -1.7818692446798573E-199, 2070302569 -> -1.9363934480648376E152, -142612373 -> -5.289781829272728E-173, -743939942 -> 4.5267978059340745E-158, 284912769 -> -1.3334534846758688E-109, 721636973 -> 6.284482522880498E-7, 44527593 -> -7.149354645962933E255, -143598766 -> 8.298912027782988E111, 2147483647 -> 79.72079182014002, -1 -> 3.731777436319693E-18, 1764631933 -> -1.3174668462363633E106, 1763610516 -> 3.381906954428537E-63, -1324222957 -> -9.59155501289049E-30, -1413936665 -> 5.083708421676393E273),71.81075800741212),
[info]       arg2 = DecayedVector(Map(-1086016398 -> 1.0231140798022226E-136, 0 -> 3.01344056575156E-219, -2147483648 -> 1.867681696329119E300, 1659086411 -> 6.737487374650964E-225, 564546781 -> -3.0766401510152493E22, 1927521424 -> 1.0814416316955774E-185, 566194627 -> 5.292057531131149E-53, 1 -> -3.065831822976799E228, 524856393 -> -2.8200395537352725E299, 61220771 -> -2.761004362337464E171, 873269565 -> -4.347660905341913E-96, -461054465 -> -1.1321040259930026E35, -1661146768 -> -2.8889830019201256E28, 1994521382 -> -4.768238397220232E-53, 1341602364 -> 3.734494775277793E96, -1174356579 -> 1.2689849505531487E8, 2147483647 -> 1.6877385989217227E124, -1 -> 1.4503547040657538E117, 349094581 -> 4.239784438300051E231, 1417192610 -> 1.0801072326468062E-274, 386338808 -> -9.199277461295296E212, -1609871611 -> 1.068174325090749E-250, 1732598637 -> -4.140681213039893E101),186.93533307523302)
[info]     )
[info]     Label of failing property:
[info]       isAssociativeEq
```
Should I just file an issue for that one? Or is this some sort of known numerical-precision kind of thing?

algebird-test/src/test/scala/com/twitter/algebird/ScanTest.scala
```diff
@@ -0,0 +1,146 @@
...
+      .joinWithPriorState
+      .joinWithInput
+      .andThenPresent { case (input, (state, ())) => (input :: state).mkString.reverse }
+
+  val joinWithPriorStateFreeScan2: StringScan =
+    directFreeScan
+      .andThenPresent(_ => ())
+      .joinWithPriorState
+      .join(Scan.identity[Char])
+      .andThenPresent { case ((state, ()), input) => (input :: state).mkString.reverse }
```
jeff-stripe, line 55: Looking at this again, `joinWithInput` and `join(Scan.identity)` end up putting the joined values in opposite orders. Is it worth bikeshedding which order should be the right one?

johnynek:
@jeff-stripe sorry about the flake. I think it is numerical precision, but it does come up periodically.

algebird-test/src/test/scala/com/twitter/algebird/ScanTest.scala
johnynek: if you look here you can see how the coverage report is generated:

https://github.com/twitter/algebird/blob/develop/.travis.yml#L14

basically you do something like `sbt coverage clean test coverageReport`, so you are enabling coverage instrumentation with `coverage` then after you run `test` you run `coverageReport` and it emits an xml and html output you can load in your browser (it prints the path IIRC).

algebird-core/src/main/scala/com/twitter/algebird/Scan.scala
```diff
@@ -0,0 +1,320 @@
...
+    override def presentAndNextState(i: Any, stateBeforeProcessingI: S): (O, S) =
+      destructor(stateBeforeProcessingI)
+  }
+
+  /**
+   * A Scan that returns the number N for the Nth input (starting from 0)
+   */
+  val index: Aux[Any, Long, Long] = iterate(0L)(n => (n, n + 1))
+
+  def identity[A] = fromFunction[A, A](x => x)
```
johnynek: can we write the type here? `def identity[A]: Aux[A, Unit, A]` I think

algebird-core/src/main/scala/com/twitter/algebird/Scan.scala
```diff
@@ -0,0 +1,320 @@
...
+   * @tparam C
+   * @return A scan which, when given `[a_1, ..., a_n]` outputs `[c_1, ..., c_n]` where
+   *         `c_i = initState + aggregator.prepare(a_1) + ... + aggregator.prepare(a_i)`
+   */
+  def fromAggregator[A, B, C](aggregator: Aggregator[A, B, C], initState: B): Aux[A, B, C] =
+    from(initState) { (a: A, stateBeforeProcessingI: B) =>
+      // nb: the order of the arguments to semigroup.plus here is what determines the order of the final summation;
+      // this matters because not all semigroups are commutative
+      val stateAfterProcessingA =
+        aggregator.semigroup.plus(stateBeforeProcessingI, aggregator.prepare(a))
```
johnynek: I think `val stateAfterProcessingA = aggregator.append(stateBeforeProcessingI, a)` is the same (and potentially optimized by the implementer of the Aggregator).

algebird-test/src/test/scala/com/twitter/algebird/ScanTest.scala
johnynek: yes. `WithIndex` in scala puts the Index *after* the value `A => (A, Int)`, so, `WithInput` might also put it on the right to be consistent with that approach.

Particularly dangerous when they are the same type!

algebird-core/src/main/scala/com/twitter/algebird/Scan.scala
jeff-stripe: Good catch; the downside of IntelliJ filling in these details for you is that it's easy to not notice that the type isn't actually written out in the code.

johnynek (commented):
yeah, agree with your suggestion to move the order in `joinWithInput` and suggested using `Aggregator.append`

jeff-stripe:
@johnynek Pushed your suggested changes + new tests (I made them fail first) for `replaceState` and `Scan.const`. The only thing left untested is the `Applicative` (my local codecov says that it's covered, but I don't actually believe that).

algebird-core/src/main/scala/com/twitter/algebird/Scan.scala
```diff
@@ -0,0 +1,322 @@
...
+ * the `scanLeft`, then throw that element away in a `map` immediately after the scanLeft is done, then this
+ * abstraction is for you.
+ *
+ * The canonical method to use a scan is `apply`.
+ *
+ *
+ * @tparam I The type of elements that the computation is scanning over.
+ * @tparam O The output type of the scan (typically distinct from the hidden `State` of the scan).
+ */
+sealed trait Scan[-I, +O] extends Serializable {
```
johnynek: actually, sorry for missing this earlier: `sealed abstract class` is slightly better for binary compatibility on 2.11 and earlier (2.12 is okay) because we can add methods without breaking binary compatibility.

johnynek (commented):
one last minor request, and I'll merge

algebird-core/src/main/scala/com/twitter/algebird/Scan.scala
jeff-stripe: Done.

jeff-stripe:
Done. Thanks for the helpful feedback and thorough review! The result is much better than where it started.

PR #764, opened 2019-12-17
Update scalacheck to 1.14.3

nevillelyh:


regadas:
Isn't this a dup of #761 ? 😄

nevillelyh:
@regadas it's a rebase after fixing ruby gem issue.

Issue #716, opened 2019-10-02
use fast immutable bitset for better bloom filter


johnynek:
see: https://github.com/typelevel/cats-collections/blob/master/core/src/main/scala/cats/collections/BitSet.scala

@non and I implemented that to see if we could make a fast, sparse, immutable bitset. We could! In fact, for some operations it is faster than mutable, and for the use case of a Monoid where we have to copy inputs, it is much faster.

We could copy that implementation into Algebird and remove use of the bitset implementation we have now (java ewah), and it would also remove one of the dependencies of core taking us down to 1 (typelevel algebra): https://github.com/twitter/algebird/blob/develop/build.sbt#L258

this would be binary incompatible. We could move the old BloomFilter into `algebird.legacy` and make the javaewah dep provided so people who don't use it don't pick it up in their build.

If we do this, we should add benchmarks to compare, but my guess is this will actually be faster.

cc @anish749 @regadas @nevillelyh

anish749:
I had tried out this implementation from cats and have the code in [this branch](https://github.com/anish749/algebird/blob/cats-no-sparse/algebird-core/src/main/scala/com/twitter/algebird/BloomFilter.scala)
It does not have for persistence or moving the old implementation to `algebird.legacy`

It is indeed much faster.

The benchmark results are [here](https://github.com/twitter/algebird/pull/673#issuecomment-466822290). I did this a few months back, and I would need to check some of the finer details again.

```

// Immutable with Array copy
[info] Benchmark                                               (falsePositiveRate)  (nbrOfElements)   Mode  Cnt    Score     Error  Units
[info] BloomFilterCreateBenchmark.createBloomFilter                          0.001            10000  thrpt    3  171.513 ± 281.008  ops/s
[info] BloomFilterCreateBenchmark.createBloomFilter                          0.001            50000  thrpt    3   11.661 ± 175.592  ops/s
[info] BloomFilterCreateBenchmark.createBloomFilter                           0.01            10000  thrpt    3   36.594 ±  87.057  ops/s
[info] BloomFilterCreateBenchmark.createBloomFilter                           0.01            50000  thrpt    3   25.702 ± 259.042  ops/s
[info] BloomFilterCreateBenchmark.createBloomFilter                            0.1            10000  thrpt    3  409.528 ± 540.620  ops/s
[info] BloomFilterCreateBenchmark.createBloomFilter                            0.1            50000  thrpt    3   76.085 ± 264.660  ops/s
[...]

PR #774, opened 2020-01-19
Replace BloomFilter underlying BitSet with cats-collections immutable BitSet

regadas:
See #716 for context.

Update: https://github.com/twitter/algebird/issues/716#issuecomment-575951635

### DIFF
--- a/algebird-benchmark/src/main/scala/com/twitter/algebird/benchmark/BloomFilterDistanceBenchmark.scala
+++ b/algebird-benchmark/src/main/scala/com/twitter/algebird/benchmark/BloomFilterDistanceBenchmark.scala
@@ -1,26 +1,16 @@
 package com.twitter.algebird
 package benchmark
 
+import com.twitter.algebird.immutable.BitSet
 import org.openjdk.jmh.annotations._
 
 object BloomFilterDistanceBenchmark {
 
-  def toSparse[A](bf: BF[A]): BFSparse[A] = bf match {
-    case BFZero(hashes, width) => BFSparse(hashes, RichCBitSet(), width)
-    case BFItem(item, hashes, width) =>
-      BFSparse(hashes, RichCBitSet.fromArray(hashes(item)), width)
-    case bfs @ BFSparse(_, _, _) => bfs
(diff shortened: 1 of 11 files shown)


johnynek:
If we are going to make a new implementation, I'd like to explore a pattern @non and I have found for parameterized monoids:
```scala
final class BloomFilter[A](hashes: Int, width: Int)(implicit val hasher: Hasher[A]) {
  sealed class BF[A] {
    def mightContain(a: A): Boolean
  }

  final case class BFInstance[A](toValue: A) extends BF[A]
  final case class BFSet(toBitSet: BitSet) extends BF[A]
  // ...

  def apply(a: A): BF[A] = ...

  implicit val monoid: Monoid[BF[A]] = ...
}
```

Then you use it with:

```scala
val bloomFilter: BloomFilter[A] = new BloomFilter(4, 1024)

val itemA1: bloomFilter.BF[A] = bloomFilter(a1)
val itemA2: bloomFilter.BF[A] = bloomFilter(a2)

val a3 = Monoid.plus(itemA1, itemA2)
```
Now, the type system prevents us from confusing bloom filter values from different parameter sets. Also, the serializers can be inside `BloomFilter` which knows the parmeters, so you don't need to serialize the parameters with each value.

johnynek:
what do you think?

regadas:
Sounds good! ... I've been meaning to pick this up again! I'll explore that!

A few notes around this PR:

* I moved the current impl to a legacy package ... I'll probably move it back (keep things as is) and add the new impl in a new package. (it will be opt in)

* I really think the BitSet impl as potential and as I mentioned in the comment, perf wise is on par and slightly better in some cases. When it comes to `createBloomFilterUsingFold` and `createBloomFilterAggregator` from #672 I saw significant improvements as well.

* I think you and @non already did some ground work in https://github.com/typelevel/cats-collections/pull/184 some things in this impl are a mix of that with the impl from master. Not sure what the move is ... if we cleanup that impl and merge it (we would depend on cats-collections) or we copy the impl (this gives us more freedom to tap into private mutable methods when needed to)

regadas (commented):
Closing this in favour of #840

PR #775, opened 2020-01-28
Cleanup around Monoid and Algebra test

regadas:
Some minor things I stumbled upon

### DIFF
--- a/algebird-core/src/main/scala/com/twitter/algebird/Monoid.scala
+++ b/algebird-core/src/main/scala/com/twitter/algebird/Monoid.scala
@@ -48,8 +48,12 @@ trait Monoid[@specialized(Int, Long, Float, Double) T]
       throw new java.lang.IllegalArgumentException("argument should not be zero")
     }
 
-  def nonZeroOption(v: T): Option[T] = Some(v).filter(isNonZero)
-
+  def nonZeroOption(v: T): Option[T] =
+    if (isNonZero(v)) {
+      Some(v)
+    } else {
+      None
+    }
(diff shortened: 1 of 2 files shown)


algebird-core/src/main/scala/com/twitter/algebird/Monoid.scala
```diff
@@ -48,12 +48,8 @@ trait Monoid[@specialized(Int, Long, Float, Double) T]
       throw new java.lang.IllegalArgumentException("argument should not be zero")
     }
 
-  def nonZeroOption(v: T): Option[T] =
-    if (isNonZero(v)) {
-      Some(v)
-    } else {
-      None
-    }
+  def nonZeroOption(v: T): Option[T] = Some(v).filter(isNonZero)
```
johnynek: This requires allocating a closure and virtual dispatch that won’t be inlined. I’d rather not change this since it can almost only reduce performance.

johnynek (changes requested):
I don’t want to do one of the changes but the rest look good.

algebird-core/src/main/scala/com/twitter/algebird/Monoid.scala
regadas: Hi @johnynek thanks for looking at this! oh i see it's a good point; i'll revert this! 👍

johnynek: Ps: I hope that scala 3 can fix this by making Option filter map etc... an in-line def.

regadas: +1 on that

johnynek (approved):
Thanks!

PR #809, opened 2020-04-18
Add DecayingCMS[K] type.

non:
This type represents a count-min sketch whose values decay over time
according to a provided half-life.

It uses a module-style approach, where the half-life, width, depth,
and other parameters are set in a module. The actual CMS values are
path-dependent types on the DecayingCMS[K] module value. This means
that we don't have to carry around or serialize anything besides the
count-min values themselves, and also means we can be sure that CMS
values to be combined are aligned.

It's a slightly different pattern that people may be used to, but I
think it works much better for this kind of thing. (I could imagine
creating a non-decaying variant with a similar design, but that's
outside the scope of this PR.)

cc @johnynek

### DIFF
--- a/algebird-core/src/main/scala/com/twitter/algebird/DecayingCMS.scala
+++ b/algebird-core/src/main/scala/com/twitter/algebird/DecayingCMS.scala
@@ -1,0 +1,689 @@
+package com.twitter.algebird
+
+import java.lang.Double.{compare => cmp}
+import java.lang.Math
+import java.util.Arrays.deepHashCode
+import scala.concurrent.duration.Duration
+import scala.util.Random
+
+/**
+ * DecayingCMS is a module to build count-min sketch instances whose
+ * counts decay exponentially.
+ *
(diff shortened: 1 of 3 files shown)


johnynek:
Thanks a ton!

So great to see this code in OSS!

Two issues:

1. can you run scalafmt? Looks like our fmt check failed.
2. Looks like the tests either flaked of failed on 2.13.1 + java8. Can you check and see if this is a flake due to a overly strict floating point issue, and if so loosen it so it doesn't make the CI fail for future runs?
```
[info] *** 1 TEST FAILED ***
[error] Failed: Total 915, Failed 1, Errors 0, Passed 914
[error] Failed tests:
[error] 	com.twitter.algebird.DecayingCMSProperties
```

I really appreciate you doing this! I forgot how good this code was (the DoubleAt, and all the laws we had to make sure it was sensible.).

johnynek:
This is awesome! thanks for sending the PR!

PR #814, opened 2020-04-29
Added Spark Dataset support

Tomczik76:


johnynek:
This is great! thanks for the PR. I'd like to merge it. you need to accept the contributor agreement above.

Also, can you run `scalafmtAll` at the sbt prompt and commit the results?

algebird-spark/src/main/scala/com/twitter/algebird/spark/implicits/package.scala
```diff
@@ -0,0 +1,29 @@
+package com.twitter.algebird.spark
+
+package object implicits {
+
+import com.twitter.algebird.BF
```
johnynek, line 5: can you move the imports to the top of the file?

algebird-spark/src/main/scala/com/twitter/algebird/spark/package.scala
```diff
@@ -10,16 +10,22 @@ import scala.reflect.ClassTag
...
+import org.apache.spark.sql.Dataset
+
   /**
    * spark exposes an Aggregator type, so this is here to avoid shadowing
    */
   type AlgebirdAggregator[A, B, C] = Aggregator[A, B, C]
   val AlgebirdAggregator = Aggregator
 
-  implicit class ToAlgebird[T](val rdd: RDD[T]) extends AnyVal {
+  implicit class ToAlgebirdRDD[T](val rdd: RDD[T]) extends AnyVal {
```
johnynek, line 21: can we not change the name of this? this will break source and binary compatibility for people.

algebird-spark/src/test/scala/com/twitter/algebird/spark/AlgebirdDatasetTests.scala
```diff
@@ -0,0 +1,117 @@
...
+class AlgebirdDatasetTest extends AnyFunSuite with BeforeAndAfter {
+
+  private var spark: SparkSession = _
+
+  before {
+    spark = SparkSession.builder().master("local").getOrCreate()
+  }
+
+  after {
+    // try spark.stop()
```
regadas, line 31: Seems that we can remove this `after` block?

algebird-spark/src/main/scala/com/twitter/algebird/spark/implicits/package.scala
```diff
@@ -0,0 +1,29 @@
+package com.twitter.algebird.spark
+
+package object implicits {
```
regadas, line 3: I think I would actually create an `EncoderInstances` `trait` and have `package object spark extends EncoderInstances `.

Users will do `import com.twitter.algebird.spark._` and they will get these lower priority `implicits`. I can see that this was to possibly be inline with spark way of things, but I think this is more consistent with the rest of algebird.

algebird-spark/src/main/scala/com/twitter/algebird/spark/implicits/package.scala
```diff
@@ -0,0 +1,29 @@
...
+
+import java.util.PriorityQueue
+
+import com.twitter.algebird.BloomFilterMonoid
+
+import org.apache.spark.sql.Encoder
+
+  import scala.reflect.ClassTag
+  implicit def kryoPriorityQueueEncoder[A](implicit ct: ClassTag[PriorityQueue[A]]): Encoder[PriorityQueue[A]] =
+    org.apache.spark.sql.Encoders.kryo[PriorityQueue[A]](ct)
```
regadas, line 17:
```suggestion
    org.apache.spark.sql.Encoders.kryo[PriorityQueue[A]]
```
iirc `ClassTag` is a context bound so the above should be possible?

algebird-spark/src/test/scala/com/twitter/algebird/spark/AlgebirdDatasetTests.scala
```diff
@@ -0,0 +1,117 @@
+package com.twitter.algebird.spark
+
+import com.twitter.algebird.{MapAlgebra, Monoid, Semigroup}
+import org.scalatest._
+import org.scalatest.funsuite.AnyFunSuite
+import com.twitter.algebird.Min
+import org.apache.spark.sql.SparkSession
+import org.apache.spark.sql.Encoder
+import org.apache.spark.sql.Dataset
+import com.twitter.algebird.BloomFilter
```
regadas, line 10: seems unused?

algebird-spark/src/test/scala/com/twitter/algebird/spark/AlgebirdDatasetTests.scala
```diff
@@ -0,0 +1,117 @@
...
+
+  import com.twitter.algebird.spark.implicits._
+
+  implicit val hash = Hash128.intHash
+  /**
+   * These tests almost always timeout on Travis. Leaving the
+   * above to at least check compilation
+   */
+  test("aggregate") {
+    val sparkImplicits = spark.implicits
```
regadas, line 84: I think I would move these to #L24 to be more evident that these are indeed the `SparkSession` implicitis

algebird-spark/src/main/scala/com/twitter/algebird/spark/implicits/package.scala
Tomczik76: I tried this and it didn't work with `spark.implicits._`. So I'm moving them into the `package object spark` instead.

johnynek:
@Tomczik76 I'd love to include this. Any idea when you might have time to get back around to it?

PR #818, opened 2020-05-07
Fix scalacOptions

regadas:
`-optimize` behaves differently (more agressive) for scala >=2.12 and we were applying it for 2.13 builds.

inline optimization is still kept for >=2.12 but it's only from algebird packages.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -52,22 +52,18 @@ val sharedSettings = Seq(
     "-language:existentials"
   ),
   scalacOptions ++= {
-    if (scalaVersion.value.startsWith("2.11"))
-      Seq("-Ywarn-unused", "-Ywarn-unused-import")
-    else
-      Seq()
-  },
-  scalacOptions ++= {
-    if (scalaVersion.value.startsWith("2.12"))
-      Seq("-Ywarn-unused", "-opt:l:inline", "-opt-inline-from:com.twitter.algebird.**")
-    else
-      Seq("-optimize")
-  },
-  scalacOptions ++= {
-    if (isScala213x(scalaVersion.value)) {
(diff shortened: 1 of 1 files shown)


regadas:
Hi @johnynek, would you have time to make a release and also include #818? Thanks!

johnynek:
oh man... good catch... yeah, I will do a release tomorrow.

regadas:
Thanks @johnynek 🙏

johnynek:
I just spent some time on trying to publish.... after churning for quite a while it failed saying this machine has no GPG key. I thought I had published from this machine before.... now I'll go back into debug mode... I hope to have it solved today.

johnynek:
https://github.com/twitter/algebird/releases/tag/v0.13.7

*seems* to have worked with a machine that had an existing gpg key (I wish the plugin could test that before doing a ton of work). Just did `sbt release` and everything seemed to work.

Thanks to you and @nevillelyh fixing up the sbt release script.

regadas:
Ahah that's good news! 👍 glad you got it working. Don't know if it's of interest but i have a simple Github Actions workflow that allow to publish a release on tag vX.X.X. It only requires adding a few secrets in the repo settings. If this is something you would want to explore let me know 👍 

We can have it publish signed snapshots on merge to master as well ... that way we also now that signing always works.

johnynek:
yeah, that sounds interesting for sure, then anyone who can push to master/merge can effectively publish (so, that would include you guys right?)

It would also be probably worth it to migrate CI wholesale to github actions...

regadas:
> yeah, that sounds interesting for sure, then anyone who can push to master/merge can effectively publish (so, that would include you guys right?)

correct. Ok I'll propose something soon.

PR #829, opened 2020-06-05
group for calculating Pearson's Correlation Coefficient

eigenvariable:
Analogous to the `Moments` group, but for covariance and Pearson's correlation coefficient instead.

In principle, this could probably be extended to Coskewness and Cokurtosis; I don't have a reference for that, so in a sense it would be original research, and hence isn't part of this PR. Another direction for future work would be to add weights and/or exponential decay to the calculation of correlation, both of which are covered in https://dl.acm.org/doi/10.1145/3221269.3223036. 

There was a delicate balance between reusing business logic from `Moments` and re-implementing it. I'm especially open to feedback on re-working some of that.

### DIFF
--- a/algebird-core/src/main/scala/com/twitter/algebird/CorrelationGroup.scala
+++ b/algebird-core/src/main/scala/com/twitter/algebird/CorrelationGroup.scala
@@ -2,7 +2,7 @@ package com.twitter.algebird
 
 object Correlation {
   def apply(x: (Double, Double), weight: Double): Correlation =
-    Correlation(c2 = 0, m2Left = 0, m2Right = 0, m1Left = x._1, m1Right = x._2, weight)
+    Correlation(c2 = 0, m2x = 0, m2y = 0, m1x = x._1, m1y = x._2, weight)
 
   def apply(x: (Double, Double)): Correlation =
     apply(x, 1.0)
@@ -46,34 +46,27 @@ object Correlation {
  * A class to calculate covariance and the first two central moments of a sequence of pairs of Doubles, from which the
  * pearson correlation coeifficient can be calculated.
  *
- * m{i}Left denotes the ith central moment of the first projection of the pair.
- * m{i}Right denotes the ith central moment of the second projection of the pair.
- * c2 the covariance equivalent of the second central moment, i.e. c2 = Sum_(l,r) (l - m1Left)*(r - m1Right).
+ * m{i}x denotes the ith central moment of the first projection of the pair.
+ * m{i}y denotes the ith central moment of the second projection of the pair.
+ * c2 the covariance equivalent of the second central moment, i.e. c2 = Sum_(x,y) (x - m1x)*(y - m1y).
  *
  */
-case class Correlation(
-    c2: Double,
-    m2Left: Double,
-    m2Right: Double,
-    m1Left: Double,
-    m1Right: Double,
-    m0: Double
-) {
+case class Correlation(c2: Double, m2x: Double, m2y: Double, m1x: Double, m1y: Double, m0: Double) {
   def totalWeight: Double = m0
 
-  def meanLeft: Double = m1Left
+  def meanX: Double = m1x
 
-  def meanRight: Double = m1Right
+  def meanY: Double = m1y
 
   // variance, stddev, and covariance are for the population, not a sample
 
-  def varianceLeft: Double = m2Left / m0
+  def varianceX: Double = m2x / m0
 
-  def varianceRight: Double = m2Right / m0
+  def varianceY: Double = m2y / m0
 
-  def stddevLeft: Double = Math.sqrt(varianceLeft)
+  def stddevX: Double = Math.sqrt(varianceX)
 
-  def stddevRight: Double = Math.sqrt(varianceRight)
+  def stddevY: Double = Math.sqrt(varianceY)
 
   def covariance: Double = c2 / totalWeight
 
@@ -85,7 +78,7 @@ case class Correlation(
     // correlation is defined as: covariance / (varianceLeft * varianceRight)
     // however, dividing by "count" cancels out, and leaves us with the following formula, which relies on fewer
     // divisions
-    c2 / (Math.sqrt(m2Left * m2Right))
+    c2 / (Math.sqrt(m2x * m2y))
 
   /**
    * Assume this instance of Correlation came from summing together Correlation.apply((x_i, y_i)) for i in 1...n.
@@ -93,13 +86,13 @@ case class Correlation(
    *         See, e.g. https://mathworld.wolfram.com/LeastSquaresFitting.html.
    */
   def linearLeastSquares: (Double, Double) = {
-    val m = c2 / m2Left
-    val b = meanRight - m * meanLeft
+    val m = c2 / m2x
+    val b = meanY - m * meanX
     (m, b)
   }
 
   def swap: Correlation =
-    Correlation(c2 = c2, m2Left = m2Right, m2Right = m2Left, m1Left = m1Right, m1Right = m1Left, m0 = m0)
+    Correlation(c2 = c2, m2x = m2y, m2y = m2x, m1x = m1y, m1y = m1x, m0 = m0)
 }
 
 object CorrelationGroup extends Group[Correlation] {
@@ -122,18 +115,18 @@ object CorrelationGroup extends Group[Correlation] {
       zero
     else {
       val prodSumRatio = a.totalWeight * b.totalWeight / count
-      val m1Left = Correlation.getCombinedMean(a.totalWeight, a.m1Left, b.totalWeight, b.m1Left)
-      val m1Right = Correlation.getCombinedMean(a.totalWeight, a.m1Right, b.totalWeight, b.m1Right)
-      val deltaLeft = b.m1Left - a.m1Left
-      val deltaRight = b.m1Right - a.m1Right
+      val m1x = Correlation.getCombinedMean(a.totalWeight, a.m1x, b.totalWeight, b.m1x)
+      val m1y = Correlation.getCombinedMean(a.totalWeight, a.m1y, b.totalWeight, b.m1y)
+      val deltaX = b.m1x - a.m1x
+      val deltaY = b.m1y - a.m1y
 
-      val m2Left = a.m2Left + b.m2Left + math.pow(deltaLeft, 2) * prodSumRatio
-      val m2Right =
-        a.m2Right + b.m2Right + math.pow(deltaRight, 2) * prodSumRatio
+      val m2x = a.m2x + b.m2x + math.pow(deltaX, 2) * prodSumRatio
+      val m2y =
+        a.m2y + b.m2y + math.pow(deltaY, 2) * prodSumRatio
 
-      val c2 = a.c2 + b.c2 + deltaLeft * deltaRight * prodSumRatio
+      val c2 = a.c2 + b.c2 + deltaX * deltaY * prodSumRatio
 
-      Correlation(c2 = c2, m2Left = m2Left, m2Right = m2Right, m1Left = m1Left, m1Right = m1Right, m0 = count)
+      Correlation(c2 = c2, m2x = m2x, m2y = m2y, m1x = m1x, m1y = m1y, m0 = count)
     }
   }
 
@@ -142,10 +135,10 @@ object CorrelationGroup extends Group[Correlation] {
   override def negate(v: Correlation): Correlation =
     Correlation(
       c2 = -v.c2,
-      m2Left = -v.m2Left,
-      m2Right = -v.m2Right,
-      m1Left = v.m1Left,
-      m1Right = v.m1Right,
+      m2x = -v.m2x,
+      m2y = -v.m2y,
+      m1x = v.m1x,
+      m1y = v.m1y,
       m0 = -v.m0
     )
 
--- a/algebird-core/src/main/scala/com/twitter/algebird/CorrelationGroup.scala
+++ b/algebird-core/src/main/scala/com/twitter/algebird/CorrelationGroup.scala
@@ -7,7 +7,14 @@ object Correlation {
   implicit val group: Group[Correlation] = CorrelationGroup
 }
 
-case class Correlation(c2: Double, m2Left: Double, m2Right: Double, m1Left: Double, m1Right: Double, m0: Long) {
+case class Correlation(
+    c2: Double,
+    m2Left: Double,
+    m2Right: Double,
+    m1Left: Double,
+    m1Right: Double,
+    m0: Long
+) {
   def count: Long = m0
 
   def meanLeft: Double = m1Left
@@ -52,7 +59,8 @@ object CorrelationGroup extends Group[Correlation] {
       val deltaRight = b.m1Right - a.m1Right
 
       val m2Left = a.m2Left + b.m2Left + math.pow(deltaLeft, 2) * prodSumRatio // a.count * b.count / count
-      val m2Right = a.m2Right + b.m2Right + math.pow(deltaRight, 2) * prodSumRatio // a.count * b.count / count
+      val m2Right =
+        a.m2Right + b.m2Right + math.pow(deltaRight, 2) * prodSumRatio // a.count * b.count / count
 
       val c2 = a.c2 + b.c2 + deltaLeft * deltaRight * prodSumRatio // a.count * b.count / count
 
@@ -63,7 +71,14 @@ object CorrelationGroup extends Group[Correlation] {
   override val zero = Correlation(0, 0, 0, 0, 0, 0)
 
   override def negate(v: Correlation): Correlation =
-    Correlation(c2 = -v.c2, m2Left = -v.m2Left, m2Right = -v.m2Right, m1Left = v.m1Left, m1Right = v.m1Right, m0 = -v.m0)
+    Correlation(
+      c2 = -v.c2,
+      m2Left = -v.m2Left,
+      m2Right = -v.m2Right,
+      m1Left = v.m1Left,
+      m1Right = v.m1Right,
+      m0 = -v.m0
+    )
 
 }
 
--- a/algebird-core/src/main/scala/com/twitter/algebird/CorrelationGroup.scala
+++ b/algebird-core/src/main/scala/com/twitter/algebird/CorrelationGroup.scala
@@ -119,23 +119,19 @@ object CorrelationGroup extends Group[Correlation] {
    */
   override def plus(a: Correlation, b: Correlation): Correlation = {
     val count = a.totalWeight + b.totalWeight
-    if (count == 0)
-      zero
-    else {
-      val prodSumRatio = a.totalWeight * b.totalWeight / count
-      val m1x = Correlation.getCombinedMean(a.totalWeight, a.m1x, b.totalWeight, b.m1x)
-      val m1y = Correlation.getCombinedMean(a.totalWeight, a.m1y, b.totalWeight, b.m1y)
-      val deltaX = b.m1x - a.m1x
-      val deltaY = b.m1y - a.m1y
+    val prodSumRatio = a.totalWeight * b.totalWeight / count
+    val m1x = Correlation.getCombinedMean(a.totalWeight, a.m1x, b.totalWeight, b.m1x)
+    val m1y = Correlation.getCombinedMean(a.totalWeight, a.m1y, b.totalWeight, b.m1y)
+    val deltaX = b.m1x - a.m1x
+    val deltaY = b.m1y - a.m1y
 
-      val m2x = a.m2x + b.m2x + math.pow(deltaX, 2) * prodSumRatio
-      val m2y =
-        a.m2y + b.m2y + math.pow(deltaY, 2) * prodSumRatio
+    val m2x = a.m2x + b.m2x + math.pow(deltaX, 2) * prodSumRatio
+    val m2y =
+      a.m2y + b.m2y + math.pow(deltaY, 2) * prodSumRatio
 
-      val c2 = a.c2 + b.c2 + deltaX * deltaY * prodSumRatio
+    val c2 = a.c2 + b.c2 + deltaX * deltaY * prodSumRatio
 
-      Correlation(c2 = c2, m2x = m2x, m2y = m2y, m1x = m1x, m1y = m1y, m0 = count)
-    }
+    Correlation(c2 = c2, m2x = m2x, m2y = m2y, m1x = m1x, m1y = m1y, m0 = count)
   }
 
   override val zero = Correlation(0, 0, 0, 0, 0, 0)
--- a/algebird-core/src/main/scala/com/twitter/algebird/CorrelationGroup.scala
+++ b/algebird-core/src/main/scala/com/twitter/algebird/CorrelationGroup.scala
@@ -7,6 +7,15 @@ object Correlation {
   implicit val group: Group[Correlation] = CorrelationGroup
 }
 
+/**
+ * A class to calculate covariance and the first two central moments of a sequence of pairs of Doubles, from which the
+ * pearson correlation coeifficient can be calculated.
+ *
+ * m{i}Left denotes the ith central moment of the first projection of the pair.
+ * m{i}Right denotes the ith central moment of the second projection of the pair.
+ * c2 the covariance equivalent of the second central moment, i.e. c2 = Sum_(l,r) (l - m1Left)*(r - m1Right).
+ *
+ */
 case class Correlation(
     c2: Double,
     m2Left: Double,
@@ -26,7 +35,7 @@ case class Correlation(
 
   private[algebird] def rightMoment: Moments = Moments(m0 = m0, m1 = m1Right, m2 = m2Right, 0, 0)
 
-  // variance, stddev, covariance, and correlation are for the population, not a sample
+  // variance, stddev, and covariance are for the population, not a sample
   def varianceLeft: Double = leftMoment.variance
 
   def varianceRight: Double = rightMoment.variance
@@ -37,8 +46,10 @@ case class Correlation(
 
   def covariance: Double = c2 / count
 
-  // pearson's correlation coefficient
-
+  /**
+   *
+   * @return Pearson's correlation coefficient
+   */
   def correlation: Double =
     // correlation is defined as: covariance / (varianceLeft * varianceRight)
     // however, dividing by "count" cancels out, and leaves us with the following formula, which relies on fewer
@@ -47,6 +58,13 @@ case class Correlation(
 }
 
 object CorrelationGroup extends Group[Correlation] {
+
+  /**
+   * The algorithm for combining the correlation calculations from two partitions of pairs of numbers. Comes from
+   * Pébay, Philippe (2008), "Formulas for Robust, One-Pass Parallel Computation of Covariances and Arbitrary-Order Statistical Moments",
+   *   Technical Report SAND2008-6212, Sandia National Laboratories
+   * https://prod-ng.sandia.gov/techlib-noauth/access-control.cgi/2008/086212.pdf
+   */
   override def plus(a: Correlation, b: Correlation): Correlation = {
     val count = a.count + b.count
     if (count == 0)
@@ -58,11 +76,11 @@ object CorrelationGroup extends Group[Correlation] {
       val deltaLeft = b.m1Left - a.m1Left
       val deltaRight = b.m1Right - a.m1Right
 
-      val m2Left = a.m2Left + b.m2Left + math.pow(deltaLeft, 2) * prodSumRatio // a.count * b.count / count
+      val m2Left = a.m2Left + b.m2Left + math.pow(deltaLeft, 2) * prodSumRatio
       val m2Right =
-        a.m2Right + b.m2Right + math.pow(deltaRight, 2) * prodSumRatio // a.count * b.count / count
+        a.m2Right + b.m2Right + math.pow(deltaRight, 2) * prodSumRatio
 
-      val c2 = a.c2 + b.c2 + deltaLeft * deltaRight * prodSumRatio // a.count * b.count / count
+      val c2 = a.c2 + b.c2 + deltaLeft * deltaRight * prodSumRatio
 
       Correlation(c2 = c2, m2Left = m2Left, m2Right = m2Right, m1Left = m1Left, m1Right = m1Right, m0 = count)
     }
--- a/algebird-core/src/main/scala/com/twitter/algebird/CorrelationGroup.scala
+++ b/algebird-core/src/main/scala/com/twitter/algebird/CorrelationGroup.scala
@@ -1,10 +1,45 @@
 package com.twitter.algebird
 
 object Correlation {
+  def apply(x: (Double, Double), weight: Double): Correlation =
+    Correlation(c2 = 0, m2Left = 0, m2Right = 0, m1Left = x._1, m1Right = x._2, weight)
+
   def apply(x: (Double, Double)): Correlation =
-    Correlation(c2 = 0, m2Left = 0, m2Right = 0, m1Left = x._1, m1Right = x._2, 1L)
+    apply(x, 1.0)
 
   implicit val group: Group[Correlation] = CorrelationGroup
+
+  /**
+   * When combining averages, if the counts sizes are too close we
+   * should use a different algorithm.  This constant defines how
+   * close the ratio of the smaller to the total count can be:
+   */
+  private val STABILITY_CONSTANT = 0.1
+
+  /**
+   * Given two streams of doubles (weightN, an) and (weightK, ak) of form (weighted count,
+   * mean), calculates the mean of the combined stream.
+   *
+   * Uses a more stable online algorithm which should be suitable for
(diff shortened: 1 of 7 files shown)


algebird-core/src/main/scala/com/twitter/algebird/CorrelationGroup.scala
```diff
@@ -0,0 +1,107 @@
...
+ * A class to calculate covariance and the first two central moments of a sequence of pairs of Doubles, from which the
+ * pearson correlation coeifficient can be calculated.
+ *
+ * m{i}Left denotes the ith central moment of the first projection of the pair.
+ * m{i}Right denotes the ith central moment of the second projection of the pair.
+ * c2 the covariance equivalent of the second central moment, i.e. c2 = Sum_(l,r) (l - m1Left)*(r - m1Right).
+ *
+ */
+case class Correlation(
+    c2: Double,
```
eigenvariable: Unlike `Moments`, I listed the arguments here in descending order of degree. My reasoning was that `c2` is the most important quantity in this calculation, and should therefore be listed first, but I'm open to other opinions.

algebird-core/src/main/scala/com/twitter/algebird/CorrelationGroup.scala
```diff
@@ -0,0 +1,107 @@
...
+
+  /**
+   * The algorithm for combining the correlation calculations from two partitions of pairs of numbers. Comes from
+   * Pébay, Philippe (2008), "Formulas for Robust, One-Pass Parallel Computation of Covariances and Arbitrary-Order Statistical Moments",
+   *   Technical Report SAND2008-6212, Sandia National Laboratories
+   * https://prod-ng.sandia.gov/techlib-noauth/access-control.cgi/2008/086212.pdf
+   */
+  override def plus(a: Correlation, b: Correlation): Correlation = {
+    val count = a.count + b.count
+    if (count == 0)
```
eigenvariable: As with `Moments`, this is necessary to make this into a `Group` instead of just a `Monoid`. It seems a tiny bit jankny to me, but hey, negation is a nice property so why not?

algebird-core/src/main/scala/com/twitter/algebird/CorrelationGroup.scala
```diff
@@ -0,0 +1,107 @@
...
+      zero
+    else {
+      val prodSumRatio = a.count.toDouble * b.count / count
+      val m1Left = MomentsGroup.getCombinedMean(a.count, a.m1Left, b.count, b.m1Left)
+      val m1Right = MomentsGroup.getCombinedMean(a.count, a.m1Right, b.count, b.m1Right)
+      val deltaLeft = b.m1Left - a.m1Left
+      val deltaRight = b.m1Right - a.m1Right
+
+      val m2Left = a.m2Left + b.m2Left + math.pow(deltaLeft, 2) * prodSumRatio
```
eigenvariable: The calculation of the first central moment in `Moments` was nuanced enough that there's a specific method for it, so I reused it above. However, the calculation of the second central moment is an expression with a lot of variables, each of which is used once, so I didn't refactor it into its own method for the purposes of DRY. I think that was the right call but am open to feedback.

I also could have defined `m1Left`, `m2Left`, `m1Right` and `m2Right` in terms of the private methods `leftMoment` and `rightMoment`. However, I noticed that the calculation factored out into `prodSumRatio` was used in several places here, and that expanding these calculations out reduces the number of arithmetic operations. Let me know if you don't think the tradeoff is worth it.

algebird-core/src/main/scala/com/twitter/algebird/CorrelationGroup.scala
```diff
@@ -0,0 +1,107 @@
...
+    Correlation(c2 = 0, m2Left = 0, m2Right = 0, m1Left = x._1, m1Right = x._2, 1L)
+
+  implicit val group: Group[Correlation] = CorrelationGroup
+}
+
+/**
+ * A class to calculate covariance and the first two central moments of a sequence of pairs of Doubles, from which the
+ * pearson correlation coeifficient can be calculated.
+ *
+ * m{i}Left denotes the ith central moment of the first projection of the pair.
```
eigenvariable: `Left` and `Right` don't seem like great naming conventions for the first and second elements of a pair, but when you start having first and second moments of each of these, using numbers gets confusing fast. Any suggestions for improvements in naming convention would be welcome.

algebird-core/src/main/scala/com/twitter/algebird/CorrelationGroup.scala
```diff
@@ -0,0 +1,107 @@
...
+ * m{i}Right denotes the ith central moment of the second projection of the pair.
+ * c2 the covariance equivalent of the second central moment, i.e. c2 = Sum_(l,r) (l - m1Left)*(r - m1Right).
+ *
+ */
+case class Correlation(
+    c2: Double,
+    m2Left: Double,
```
eigenvariable: At some point, I experimented with having two full copies of `Moments` for each element here, but the fact that the 0th moment (count) has to be the same in both cases made that representation harder to work with (and also less space efficient).

algebird-core/src/main/scala/com/twitter/algebird/CorrelationGroup.scala
```diff
@@ -0,0 +1,107 @@
...
+    m2Right: Double,
+    m1Left: Double,
+    m1Right: Double,
+    m0: Long
```
johnynek: Can we make this a `Double` actually? If we make this a double, we can actually make Correlation a `VectorSpace` which allows us to do decayed values on Correlation (which is a pretty cool idea)....

I've been using decayed 2nd moments in a project and it seems to be a useful feature transformation since it can give you decayed z-scores.

algebird-core/src/main/scala/com/twitter/algebird/CorrelationGroup.scala
```diff
@@ -0,0 +1,107 @@
...
+) {
+  def count: Long = m0
+
+  def meanLeft: Double = m1Left
+
+  def meanRight: Double = m1Right
+
+  // useful in calculating variance and stddev; private because we don't have skewness or kurtosis info
```
johnynek: I'd rather move this code to the `Moment` object so we don't have to allocate a `Moments` class to get the result. The JIT *may* be able to allocate on the stack using escape analysis, but I'm nervous.... but if you don't have time, I understand (that said, my request to use a Double for the m0 may mean we can't use this method).

algebird-core/src/main/scala/com/twitter/algebird/CorrelationGroup.scala
```diff
@@ -0,0 +1,107 @@
...
+      m2Left = -v.m2Left,
+      m2Right = -v.m2Right,
+      m1Left = v.m1Left,
+      m1Right = v.m1Right,
+      m0 = -v.m0
+    )
+
+}
+
+object CorrelationAggregator extends MonoidAggregator[(Double, Double), Correlation, Correlation] {
```
johnynek: seems like you might want the `.andThenPresent(_.correlation)` a lot.

algebird-test/src/main/scala/com/twitter/algebird/scalacheck/Gen.scala
```diff
@@ -69,4 +69,20 @@ object gen extends ExpHistGen with IntervalGen {
   def genSSManySpaceSaver: Gen[SpaceSaver[String]] =
     Gen.nonEmptyListOf(genFixedSSOneSpaceSaver).flatMap(l => l.reduce(_ ++ _))
 
+  def genCorrelation: Gen[Correlation] =
```
johnynek: this never generates 0, right? (a count of zero)? A way that I like that might actually be simpler is to generate according to how this is used:
```scala
lazy val genCorrelation: Gen[Correlation] = {
  val recur = Gen.lzy(genCorrelation)
  
  // we can start with any pair of numbers:
  val genClose: Gen[Correlation] = for {
    x <- choose(-1000, 1000)
    delta <- choose(-100.0, 100.0)
   } yield Correlation((x, x + delta))

  val genUncorr: Gen[Correlation] = for {
    x <- choose(-1e20, 1e20)
    y <- choose(-1e20, 1e20)
   } yield Correlation((x, y))

  val genSum = Gen.zip(recur, recur).map { case (a, b) => a + b }
  val genDiff = Gen.zip(recur, recur).map { case (a, b) => a - b }
  val genNeg = recur.map(_.negate)
  
  // now return with a low probability of choosing the branching cases:
  Gen.frequency(
    (5, genClose),
    (5, genUncorr),
    (2, genNeg),
    (1, Correlation.zero),
    (1, genSum),
    (1, genDiff))
}
```
In this style, you are generating values that are built up from formulas of using the data.

algebird-core/src/main/scala/com/twitter/algebird/CorrelationGroup.scala
```diff
@@ -0,0 +1,107 @@
...
+  def varianceLeft: Double = leftMoment.variance
+
+  def varianceRight: Double = rightMoment.variance
+
+  def stddevLeft: Double = leftMoment.stddev
+
+  def stddevRight: Double = rightMoment.stddev
+
+  def covariance: Double = c2 / count
+
```
johnynek: nice to have:

we could also include the formula for a least squares fit right? So, fit `left = m * right + b` and return `(m, b)`? I think the formula is:

`m = (c2 - c0 * meanLeft * meanRight) / m2Right` and `b = meanLeft - m * meanRight` or something...

algebird-core/src/main/scala/com/twitter/algebird/CorrelationGroup.scala
```diff
@@ -0,0 +1,107 @@
...
+  /**
+   *
+   * @return Pearson's correlation coefficient
+   */
+  def correlation: Double =
+    // correlation is defined as: covariance / (varianceLeft * varianceRight)
+    // however, dividing by "count" cancels out, and leaves us with the following formula, which relies on fewer
+    // divisions
+    c2 / (Math.sqrt(m2Left * m2Right))
+}
```
johnynek: can we add `def swap: Correlation` which swaps the left with the right? So we expect `x.swap.swap == x` and `Correlation(a, b).swap == Correlation(b, a)` etc...

algebird-test/src/test/scala/com/twitter/algebird/CorrelationLaws.scala
```diff
@@ -0,0 +1,60 @@
...
+
+class CorrelationTest extends WordSpec with Matchers {
+  def aggregateFunction(f: Double => Double): Aggregator[Double, Correlation, Double] =
+    CorrelationAggregator
+      .composePrepare[Double](x => (x, f(x)))
+      .andThenPresent(_.correlation)
+
+  val testList = Range.inclusive(-10, 10).map(_.toDouble).toList
+
+  "correlation with y = x should be 1" in {
```
johnynek: I think this is also true for `y = m * x` for all m > 0.0 right? similarly below?

johnynek (commented):
I'm really sorry I let this sit so long. It's very interesting and I actually have a project where I expect I will be using this very soon.

This is really cool (especially if we make count a Double so we can decay it and have decayed correlation, so we don't have to have infinite memory).

johnynek:
Also, if you don't have the interest or time in making these changes, just let me know, I can merge it and implement them. I don't want to block great work just because I want to add a few things.

algebird-core/src/main/scala/com/twitter/algebird/CorrelationGroup.scala
eigenvariable: Taking a closer look at section 3.1 of the paper I linked to in the PR description, it looks like the calculation for weights really is as simple as making this a `Double` (the choice of notation in the two papers is annoyingly different though). Given how easy this is, it seems like a no-brainer to implement. Good call-out here.

Section 4.1 talks about weighted correlation being used as a metric (seems like an easy enough method to add) and 4.2 exponentially weighted moving correlation (which seems like something that should live outside of this particular abstraction--do you agree?).

algebird-core/src/main/scala/com/twitter/algebird/CorrelationGroup.scala
eigenvariable: Agreed, but sometimes you might want `.andThenPresent(_.covariance)` from time-to-time as well. Maybe I can add a convenience method to this object?

algebird-test/src/main/scala/com/twitter/algebird/scalacheck/Gen.scala
eigenvariable: That's true, although it _does_ get verified when validating `Group` laws. Although my guess is that you prefer this for tests written by consumers of this library, right?

algebird-core/src/main/scala/com/twitter/algebird/CorrelationGroup.scala
eigenvariable: @johnynek  I agree re:weights, and it's not very deep business logic, so I that re-implementing it here is no big deal. However, what should be the equivalent of this line? Should `> 1` still be the threshold in the presence of weights? 

https://github.com/twitter/algebird/blob/cf913f69992bf4345c3d38a46698c3c7278c6d3f/algebird-core/src/main/scala/com/twitter/algebird/MomentsGroup.scala#L35

johnynek: no, I think it should probably be `count != 0.0` I suppose. But even then, maybe we should just remove the branch and always return the value and let the caller be careful what it means if count <= 0.0?

algebird-core/src/main/scala/com/twitter/algebird/CorrelationGroup.scala
johnynek: good point. Yeah, two convenience methods on this would be good. `def correlation` and `def covariance` would be good.

algebird-core/src/main/scala/com/twitter/algebird/CorrelationGroup.scala
eigenvariable: Maybe instead of `left` and `right` I could refer to them as `x` and `y` (like the coordinates of the plane). Your thoughts, @johnynek?

algebird-core/src/main/scala/com/twitter/algebird/CorrelationGroup.scala
eigenvariable: Maybe instead of left and right I could refer to them as x and y (like the coordinates of the plane). Your thoughts, @johnynek?

algebird-core/src/main/scala/com/twitter/algebird/CorrelationGroup.scala
eigenvariable: I opted for the latter, but am open to changing it again.

algebird-test/src/main/scala/com/twitter/algebird/scalacheck/Gen.scala
eigenvariable: I ended up needing to make exponent of the range of Doubles be `10` instead of `20`. Otherwise, it was too easy for variance to become `Infinity` which often ends up making `c2` be `NaN`, which in turn makes the resulting `Correlation` instance unequal to itself.

algebird-core/src/main/scala/com/twitter/algebird/CorrelationGroup.scala
eigenvariable: Nice! I hadn't realized you could do linear least squares regression with just this info, although it makes sense. The formula for `m` turned out to be slightly different, but I was able to verify the correctness of the implementation with a nice scalacheck test.

algebird-test/src/test/scala/com/twitter/algebird/CorrelationLaws.scala
eigenvariable: It is (for any `y = mx + b`, with `m != 0`, as it turns out). I turned this into a scalacheck test, although I needed to restrict the choices of `m` and `b` to `Int`s, otherwise scalacheck was too good at finding examples where you end up losing a fair bit of precision (that said, I also conducted that experiment _before_ restricting the size of the generated `Doubles` to be `1e20`, which may have helped).

eigenvariable:
I _think_ I addressed all of the review feedback (if not, this was an unintentional oversight).

PTAL @johnynek 
(at your convenience)

algebird-core/src/main/scala/com/twitter/algebird/CorrelationGroup.scala
```diff
@@ -0,0 +1,165 @@
...
+   * mean), calculates the mean of the combined stream.
+   *
+   * Uses a more stable online algorithm which should be suitable for
+   * large numbers of records similar to:
+   * http://en.wikipedia.org/wiki/Algorithms_for_calculating_variance#Parallel_algorithm
+   *
+   * This differs from the implementation in MomentsGroup.scala only in that here, the counts are weighted, and are
+   * thus doubles instead of longs
+   */
+  def getCombinedMean(weightN: Double, an: Double, weightK: Double, ak: Double): Double =
```
eigenvariable: I don't love that this is copy-pasta. We could in principle refactor the method in `Moments` to take a `Numeric` or something like that, but I'm not sure if that would force you to bump a semantic version of Algebird, so I opted for a little bit of copying. Happy to take another approach if you prefer, @johnynek.

algebird-core/src/main/scala/com/twitter/algebird/CorrelationGroup.scala
johnynek: yeah, I like that. I'd be +1 on using x and y (and if we did emit a m, b for a linear fit, such that y = m x + b).

eigenvariable: Done. I ended up going with `m1x` instead of `m1X` because it seemed easier to read and type, but did  go with the technicall correct caml-casing with things like `meanX`. Happy to change conventions if you have a preference.

algebird-core/src/main/scala/com/twitter/algebird/CorrelationGroup.scala
```diff
@@ -0,0 +1,165 @@
...
+   * Assume this instance of Correlation came from summing together Correlation.apply((x_i, y_i)) for i in 1...n.
+   * @return (m, b) where y = mx + b is the line with the least squares fit of the points (x_i, y_i).
+   *         See, e.g. https://mathworld.wolfram.com/LeastSquaresFitting.html.
+   */
+  def linearLeastSquares: (Double, Double) = {
+    val m = c2 / m2Left
+    val b = meanRight - m * meanLeft
+    (m, b)
+  }
+
```
johnynek: now that we have m0 as a Double, can we add `def scale(z: Double): Correlation` which scales the entire correlation (which is used for making exponential decays)?

I think the mean isn't touched but the totalWeight and the second moments are changed.

The laws we want: `x.scale(0.0) == Monoid.zero`, `x.scale(a).scale(b) == x.scale(a * b)`, `x.scale(1.0) == x`. I *think* those laws are enough...

(obviously with the correct approximately equals situation).

algebird-core/src/main/scala/com/twitter/algebird/CorrelationGroup.scala
```diff
@@ -0,0 +1,158 @@
...
+   * http://en.wikipedia.org/wiki/Algorithms_for_calculating_variance#Parallel_algorithm
+   *
+   * This differs from the implementation in MomentsGroup.scala only in that here, the counts are weighted, and are
+   * thus doubles instead of longs
+   */
+  def getCombinedMean(weightN: Double, an: Double, weightK: Double, ak: Double): Double =
+    if (weightN < weightK) getCombinedMean(weightK, ak, weightN, an)
+    else
+      (weightN + weightK) match {
+        case 0L                              => 0.0
```
johnynek: shouldn't this be `0.0` not `0L`?

algebird-core/src/main/scala/com/twitter/algebird/CorrelationGroup.scala
```diff
@@ -0,0 +1,158 @@
...
+        case newCount if newCount == weightN => an
+        case newCount =>
+          val scaling = weightK.toDouble / newCount
```
johnynek: isn't `weightK` already a `Double` so no need `.toDouble` here?

johnynek (commented):
This is so great! I can't wait to use it.

Added just one more request (about scale) if you have time.

algebird-core/src/main/scala/com/twitter/algebird/CorrelationGroup.scala
eigenvariable: The linked paper seems to agree with you, but those laws are also satisfied by either a) making `scale` a no-op or b) only scaling `m0` (which it took me a while to convince myself was the wrong thing). Should there also be an invariant about the average-like quantities (`meanX`, `meanY`, `varianceX`, `varianceY`) being unchanged after scaling? That would at least rule out option b.

johnynek: I think no-op scale fails `x.scale(0.0) == zero`. So, I don't think it can be a no-op.

eigenvariable: Ok, that works. Just pushed.

algebird-core/src/main/scala/com/twitter/algebird/CorrelationGroup.scala
eigenvariable: Good catch. This was lazy copy-pasting on my part.

algebird-core/src/main/scala/com/twitter/algebird/CorrelationGroup.scala
eigenvariable: Good catch. This was lazy copy-pasting on my part.

eigenvariable:
It's worth noting that the `GroupLaws` tests are flaky for me locally now that I've updated the `Generator` to be more realistic. I think it doesn't help that `Double`s don't really satisfy those laws.

Also, I added `distanceMetric`, since that was mentioned in `Numerically stable parallel computation of (Co-)Variance` and was trivial to implement. I'm not sure what tests it would need. I'm also happy to leave it out if you think it's too obsucre.

PTAL @johnynek

johnynek (approved):
This is great! I'll merge this when CI is green.

Thanks for pushing this all the way!

eigenvariable:
It was my pleasure! Thanks for the thorough and patient review.

eigenvariable:
CI is failing because the group laws are, as I said, flaky. I tried changing the generator back to what it was before locally, and am now running into an error that doesn't appear to be a flake (TBH I'm pretty surprised that the current version of the generator doesn't uncover it). Basically, if you have an ordinary `Correlation` instance and scale it by a negative number, the sign of the value associated with `.correlation` flips (right now the assertion is written for it to be invariant). This yields two questions:

1.  Is the bug with the implementation of scale (i.e. should c2 be scaled by `math.abs(z)`?) or with the invariant in the test (i.e. should the sign of `.correlation` be expected to flip when scaling by a negative value)? I've verified that both versions of this fix pass.
2. What should we do about the generator?
a. Getting it to find the above bug is easy: just add something like `genRandom = // original implementation of genCorrelation` as one of the arguments to `Gen.frequency`.
b. Getting it to not flake on the group laws seems pretty difficult here. Even when I drop the `EPS` constant to `0.1` I still get failures like this:
```
  Falsified after 3 successful property evaluations.
  Location: (CheckProperties.scala:12)
  Occurred when passed generated values (
    arg0 = Correlation(3.4781449464340951E18,-1.2058550120859232E19,-1.00322942551780416E18,3.147562977424351E9,-1.5649316090685885E9,-6.903287220248554E9),
    arg1 = Correlation(0.0,0.0,0.0,-3.109131087800439E9,-2.583738651855649E9,1.0),
    arg2 = Correlation(-0.0,-0.0,-0.0,-786.0,-853.3705897627212,-1.0)
  )

  Label of failing property:
    isAssociativeEq
```
where the two sides  of associativity are
```
Correlation(3.4781449464340951E18,-1.2058550120859232E19,-1.00322942551780416E18,3.1475629774243507E9,-1.5649316090685885E9,-6.903287220248554E9)
Correlation(1.4778228264716098E19,1.7180512861538128E19,-2.4142699042477665E18,3.147562977874735E9,-1.5649316086943119E9,-6.903287220248554E9)
```
It's plausible that the code I wrote is somehow leaking numerical precision in a way that I haven't realized. It's also plausible to me that the magnitude of numerical precision losses get magnified as you go up the chain of moments, and that making weights be arbitrary `Double`s doesn't help the situation.

Edit: the fact that there's a sign flip for `m2x` here makes me think something deeper is going on.

Edit edit: I see something similar going on with `Moments`:

```scala
val m0 = Moments(-6903287220,3.147562977424351E9,-1.2058550120859232E19,0.0,0.0)
val m1 = Moments(1,-3.109131087800439E9,0.0,0.0,0.0)
val m2 = Moments(-1,-786.0,-0.0,0.0,0.0)

scala> val left = semi.plus(m0, semi.plus(m1, m2))                                
left: com.twitter.algebird.Moments = Moments(-6903287220,3.1475629774243507E9,-1.2058550120859232E19,0.0,0.0)

scala> val right = semi.plus(semi.plus(m0, m1), m2)
right: com.twitter.algebird.Moments = Moments(-6903287220,3.147562977874735E9,1.7180512861538128E19,-2.1374251588738178E29,1.4342748170710145E39)
```
It's worth noting that my original implementation of the generator for `Correlation` was cargo-culted from the one for `Moments`, meaning that the `Moments` tests never test with negated elements, things summed together, etc. I think it's plausible that neither one of these is actually a group.

johnynek:
@eigenvariable why don't we lower `EPS=1e-5` or something, and then generate doubles in the range of like 2e9, 2e-9, rather than go above, and see if that works.

The example you have above looks pretty bad though... 

I agree that something seems off here... I'm confused how `left` in your example wound up with 0 skew and 0 kurtosis.... I bet these is some bug with negative counts that was never tested.

johnynek:
okay, I think the issue is if `count0 + count1 == 0` we destroy the other values by returning zero....

So, that is breaking associativity I think.

It seems that rule has to go... what if you remove that, what laws break then?

Oh, and I thought of more laws we need on scale:

```scala
(x + y).scale(z) == (x.scale(z) + y.scale(z))
x.scale(y + z) == (x.scale(y) + x.scale(z))
```

Thanks for your careful testing that uncovered this issue!

johnynek:
actually... I'm not so sure anymore.... I guess we need to take the limit of `delta -> 0` with `(count + delta) - count` and see what the values are... I actually don't see why the rule of `count == 0` implies `zero` is wrong on second glance.

johnynek:
one question:

if we consider two moments the same if they both have `count == 0` OR all the approximately equal on all the fields, does that solve it? Is this just due to having bad limits as count goes to zero and forcing us to return zero which breaks associativity?

eigenvariable:
IIRC, setting everything to `0` came from some fields ending up being `NaN` under certain circumstances, which breaks reflexivity. I updated the code to just compare fields with `NaN` in them as equal, and, combined with dropping the `count == 0` comparison _seems_ to be working, even with decently strict numerical precision. I'll push the commit shortly.

eigenvariable:
Ok, I've pushed a commit that works, where there's now an assertion that scaling by a negative factor will have a negative effect on correlation. I'm still not sure if that's the right fix.

Also, I still need to implement the invariants WRT addition. Doing that now.

eigenvariable:
Ooh, one of them (`(x + y).scale(z) == (x.scale(z) + y.scale(z))`) fails :

```scala
GeneratorDrivenPropertyCheckFailedException was thrown during property evaluation.
 (CheckProperties.scala:12)
  Falsified after 0 successful property evaluations.
  Location: (CheckProperties.scala:12)
  Occurred when passed generated values (
    arg0 = Correlation(Infinity,Infinity,Infinity,6.353346554766811E9,-3.987013780614332E9,1.0),
    arg1 = 1, // 23 shrinks
    arg2 = -1 // 31 shrinks
  )
```
I don't think the `Infinity`s are the issue here, I think it's that `arg1 + arg2 == 0`. Scaling ends up multiplying everything but the means by 0, zeroing them out, but no longer forcing a count of 0 to zero out the other moments during addition breaks this property.

Edit: the other addition property fails with scaling by 0

algebird-core/src/main/scala/com/twitter/algebird/CorrelationGroup.scala
```diff
@@ -119,23 +119,19 @@ object CorrelationGroup extends Group[Correlation] {
...
-
-      val m2x = a.m2x + b.m2x + math.pow(deltaX, 2) * prodSumRatio
-      val m2y =
-        a.m2y + b.m2y + math.pow(deltaY, 2) * prodSumRatio
-
-      val c2 = a.c2 + b.c2 + deltaX * deltaY * prodSumRatio
-
-      Correlation(c2 = c2, m2x = m2x, m2y = m2y, m1x = m1x, m1y = m1y, m0 = count)
-    }
+    val prodSumRatio = a.totalWeight * b.totalWeight / count
```
johnynek: so this is `xy/(x+y)`. if x == y, this should be x/2, so if `x = y = 0` seems like this should be zero.

if y -> 0, but x != 0, we should have 0 also.

finally, if x = -y but both > 0, then we should have Infinity here, right?

algebird-core/src/main/scala/com/twitter/algebird/CorrelationGroup.scala
snoble: It looks like I wrote this piece of jank originally in Moments. As you two to have figured out here, it certainly seems to not be associative. I can submit a PR to remove it from Moments if you like.

eigenvariable: Wait, @snoble, do you get notified for all algebird PRs or did you just happen to stumble across this?

eigenvariable: Actually, it looks like it's necessary even for it to be a `Monoid` instead of a `Semigroup` (otherwise zero + zero ends up with `NaN` in it). By the time I pushed this review, I'd forgotten which property it was needed for.

eigenvariable:
OK, @johnynek and I talked offline about this, and we couldn't find a fix to make this into a group instead of just a monoid, so for now it remains just a monoid. As it turns out, I'd forgotten that the zero check in `plus` (and also `scale`) was necessary to get a `Monoid` (and not a `Group`), so I put that back in just in case people really want to have a zero.

AFAICT, everything passes now, but it's probably worth taking another look at this to make sure that none of the changes we made while exploring the above inadvertently ended up getting committed.

johnynek:
this is green except for formatting:
```
/home/travis/build/twitter/algebird/algebird-core/src/main/scala/com/twitter/algebird/CorrelationMonoid.scala isn't formatted properly!
```
Can you rerun `scalafmtAll`?

eigenvariable:
I ran `scalafmtAll` for the most recent commit, and running it again locally is a no-op. Moreover, running
```
 sbt \ "; scalafmtCheckAll; scalafmtSbtCheck" \ "; scalafixEnable; scalafix --check; test:scalafix --check"
```
(which seems to be the command that is failing in CI) succeeds for me locally. Any idea what the problem could be?

johnynek:
okay, here is the deal, since you made the PR, develop has moved on and got a later version of scalafmt. That version wants this diff:
```diff
diff --git a/algebird-core/src/main/scala/com/twitter/algebird/CorrelationMonoid.scala b/algebird-core/src/main/scala/com/twitter/algebird/CorrelationMonoid.scala
index daa4dc0f..fb2eae56 100644
--- a/algebird-core/src/main/scala/com/twitter/algebird/CorrelationMonoid.scala
+++ b/algebird-core/src/main/scala/com/twitter/algebird/CorrelationMonoid.scala
@@ -49,7 +49,6 @@ object Correlation {
  * m{i}x denotes the ith central moment of the first projection of the pair.
  * m{i}y denotes the ith central moment of the second projection of the pair.
  * c2 the covariance equivalent of the second central moment, i.e. c2 = Sum_(x,y) (x - m1x)*(y - m1y).
- *
  */
 case class Correlation(c2: Double, m2x: Double, m2y: Double, m1x: Double, m1y: Double, m0: Double) {
   def totalWeight: Double = m0
@@ -71,7 +70,6 @@ case class Correlation(c2: Double, m2x: Double, m2y: Double, m1x: Double, m1y: D
   def covariance: Double = c2 / totalWeight

   /**
-   *
    * @return Pearson's correlation coefficient
    */
   def correlation: Double =
```

So, if you do `git fetch origin` and `git merge develop`, then run `scalafmtAll` at the sbt prompt, or `sbt scalafmtAll` I think it should resolve it. Alternatively, you can manually remove the two lines above it wants removed.

eigenvariable:
I'm embarrassed that rebasing against develop didn't occur to me; I'm rusty. I just force-pushed after rebasing and rerunning `scalafmtAll`.

johnynek:
This is great. Thank you so much.

PR #840, opened 2020-07-11
Experimental path dependent bloom filter

regadas:
Hi @johnynek long overdue but here it is. I think this bloom filter is inline with what you mentioned previously and it's backed by a slightly modified version of the `cats-collections` `BitSet`. Since this is very much experimental it lives under the `experimental` package.

To be honest this has been sitting in a branch for awhile now and I've been meaning to look more into it to try to squeeze more pref (it should be possible). Putting out here to fish for extra opinions.

Let me know what you think.

```
[info] BloomFilterCreateBenchmark.createBloomFilter                                      0.001            10000  avgt    3     3.810 ±    0.607  ms/op
[info] BloomFilterCreateBenchmark.createBloomFilter                                      0.001            50000  avgt    3    19.622 ±    1.380  ms/op
[info] BloomFilterCreateBenchmark.createBloomFilter                                       0.01            10000  avgt    3     2.647 ±    0.096  ms/op
[info] BloomFilterCreateBenchmark.createBloomFilter                                       0.01            50000  avgt    3    12.683 ±    1.980  ms/op
[info] BloomFilterCreateBenchmark.createBloomFilterExperimental                          0.001            10000  avgt    3     4.717 ±    0.216  ms/op
[info] BloomFilterCreateBenchmark.createBloomFilterExperimental                          0.001            50000  avgt    3    21.364 ±    0.259  ms/op
[info] BloomFilterCreateBenchmark.createBloomFilterExperimental                           0.01            10000  avgt    3     3.341 ±    0.453  ms/op
[info] BloomFilterCreateBenchmark.createBloomFilterExperimental                           0.01            50000  avgt    3    16.450 ±    0.345  ms/op
```

```
[info] BloomFilterQueryBenchmark.queryBloomFilter                            0.001              100  avgt    3  134.870 ± 12.429  ns/op
[info] BloomFilterQueryBenchmark.queryBloomFilter                            0.001             1000  avgt    3  135.747 ± 13.614  ns/op
[info] BloomFilterQueryBenchmark.queryBloomFilter                            0.001            10000  avgt    3  136.107 ±  3.723  ns/op
[info] BloomFilterQueryBenchmark.queryBloomFilter                             0.01              100  avgt    3   96.572 ±  5.242  ns/op
[info] BloomFilterQueryBenchmark.queryBloomFilter                             0.01             1000  avgt    3   97.131 ±  8.785  ns/op
[info] BloomFilterQueryBenchmark.queryBloomFilter                             0.01            10000  avgt    3   98.331 ±  6.771  ns/op
[info] BloomFilterQueryBenchmark.queryBloomFilterExperimental                0.001              100  avgt    3  142.514 ±  6.100  ns/op
[info] BloomFilterQueryBenchmark.queryBloomFilterExperimental                0.001             1000  avgt    3  143.342 ±  4.156  ns/op
[info] BloomFilterQueryBenchmark.queryBloomFilterExperimental                0.001            10000  avgt    3  146.318 ±  6.112  ns/op
[info] BloomFilterQueryBenchmark.queryBloomFilterExperimental                 0.01              100  avgt    3  101.955 ±  3.377  ns/op
[info] BloomFilterQueryBenchmark.queryBloomFilterExperimental                 0.01             1000  avgt    3  111.415 ±  8.235  ns/op
[info] BloomFilterQueryBenchmark.queryBloomFilterExperimental                 0.01            10000  avgt    3  106.131 ±  3.394  ns/op
```

```
[info] BloomFilterCreateBenchmark.createBloomFilterAggregator                            0.001            10000  avgt    3   163.437 ±   20.909  ms/op
[info] BloomFilterCreateBenchmark.createBloomFilterAggregator                            0.001            50000  avgt    3  4822.596 ± 1001.072  ms/op
[info] BloomFilterCreateBenchmark.createBloomFilterAggregator                             0.01            10000  avgt    3    91.614 ±   16.930  ms/op
[info] BloomFilterCreateBenchmark.createBloomFilterAggregator                             0.01            50000  avgt    3  2423.146 ±  354.286  ms/op
[info] BloomFilterCreateBenchmark.createBloomFilterAggregatorExperimental                0.001            10000  avgt    3    32.485 ±    1.291  ms/op
[info] BloomFilterCreateBenchmark.createBloomFilterAggregatorExperimental                0.001            50000  avgt    3   587.019 ±  796.350  ms/op
[info] BloomFilterCreateBenchmark.createBloomFilterAggregatorExperimental                 0.01            10000  avgt    3    22.232 ±    1.424  ms/op
[info] BloomFilterCreateBenchmark.createBloomFilterAggregatorExperimental                 0.01            50000  avgt    3   370.341 ±   10.683  ms/op
[info] BloomFilterCreateBenchmark.createBloomFilterUsingFold                             0.001            10000  avgt    3   161.405 ±    5.454  ms/op
[info] BloomFilterCreateBenchmark.createBloomFilterUsingFold                             0.001            50000  avgt    3  4746.247 ±  333.122  ms/op
[info] BloomFilterCreateBenchmark.createBloomFilterUsingFold                              0.01            10000  avgt    3    94.377 ±    8.656  ms/op
[info] BloomFilterCreateBenchmark.createBloomFilterUsingFold                              0.01            50000  avgt    3  2265.540 ±   86.962  ms/op
[info] BloomFilterCreateBenchmark.createBloomFilterUsingFoldExperimental                 0.001            10000  avgt    3    32.108 ±    0.386  ms/op
[info] BloomFilterCreateBenchmark.createBloomFilterUsingFoldExperimental                 0.001            50000  avgt    3   542.427 ±   97.306  ms/op
[info] BloomFilterCreateBenchmark.createBloomFilterUsingFoldExperimental                  0.01            10000  avgt    3    22.257 ±    0.485  ms/op
[info] BloomFilterCreateBenchmark.createBloomFilterUsingFoldExperimental                  0.01            50000  avgt    3   400.538 ±  989.902  ms/op
```


algebird-core/src/main/scala/com/twitter/algebird/experimental/BloomFilter.scala
```diff
@@ -0,0 +1,402 @@
...
+        target(hashIndex) = buffer(thisBidx) % w
+        nextHash(valueToHash, hashIndex + 1, buffer, thisBidx + 1, target)
+      }
+  }
+
+  /**
+   * Bloom Filter data structure
+   */
+  sealed abstract class BF extends Serializable {
+    val numHashes: Int = n
```
johnynek: can't this be a def an save some space?

algebird-core/src/main/scala/com/twitter/algebird/experimental/BloomFilter.scala
```diff
@@ -0,0 +1,402 @@
...
+
+    val width: Int = w
```
johnynek: def?

algebird-core/src/main/scala/com/twitter/algebird/experimental/BloomFilter.scala
```diff
@@ -0,0 +1,402 @@
...
+ * Bloom Filter - a probabilistic data structure to test presence of an element.
+ *
+ * Operations
+ *   1) insert: hash the value k times, updating the bitfield at the index equal to each hashed value
+ *   2) query: hash the value k times.  If there are k collisions, then return true; otherwise false.
+ *
+ * http://en.wikipedia.org/wiki/Bloom_filter
+ */
+final case class BloomFilter[A](n: Int, w: Int)(implicit hash: Hash128[A]) { self =>
+  case object Hash {
```
johnynek: why a `case` object? why not just `object Hash`?

algebird-core/src/main/scala/com/twitter/algebird/experimental/BloomFilter.scala
```diff
@@ -0,0 +1,402 @@
...
+    }
+
+    @annotation.tailrec
+    private def nextHash(
+        valueToHash: A,
+        hashIndex: Int,
+        buffer: Array[Int],
+        bidx: Int,
+        target: Array[Int]
+    ): Array[Int] =
```
johnynek: should this maybe return `Unit` to make it more clear that is it a side effecting function? Seems that might send a clearer message.

algebird-core/src/main/scala/com/twitter/algebird/experimental/BloomFilter.scala
```diff
@@ -0,0 +1,402 @@
...
+        case (x: BF, _: Zero.type)        => x.numBits
+
+        // Otherwise compare as bit sets
+        case (_, _) => (this.toBitSet ^ that.toBitSet).size.toInt
+      }
+
+  }
+
+  case object Zero extends BF {
+    override def toBitSet: BitSet = BitSet()
```
johnynek: nit: is there no `BitSet.empty`?

algebird-core/src/main/scala/com/twitter/algebird/experimental/BloomFilter.scala
```diff
@@ -0,0 +1,402 @@
...
+    /**
+     * The number of bits set to true
+     */
+    override def numBits: Int = bits.size.toInt
+
+    override val toBitSet: BitSet = bits
+
+    override def ++(other: BF): BF =
+      other match {
+        case _: Zero.type        => this
```
johnynek: why `case _: Zero.type` and not `case Zero =>`?

algebird-core/src/main/scala/com/twitter/algebird/experimental/BloomFilter.scala
```diff
@@ -0,0 +1,402 @@
...
+      if (as.iterator.isEmpty) None
+      else {
+        var bfItem: Item = null
+        val iter = as.iterator
+        var counter = 0
+        var bs = BitSet.newEmpty(0)
+
+        while (iter.hasNext) {
+          iter.next() match {
+            case _: Zero.type => ()
```
johnynek: `case Zero =>`?

algebird-core/src/main/scala/com/twitter/algebird/experimental/BloomFilter.scala
```diff
@@ -0,0 +1,402 @@
...
+          }
+        }
+
+        if (counter == 0) Some(zero)
+        else if (counter == n && (bfItem != null)) Some(bfItem)
+        else Some(Instance(bs))
+      }
+  }
+
+  implicit val aggregator = new MonoidAggregator[A, BF, BF] {
```
johnynek: are aggregators resolved implicitly generally? I don't think so. Maybe we don't need this? Can we also add a type annotation.

algebird-core/src/main/scala/com/twitter/algebird/experimental/BloomFilter.scala
```diff
@@ -0,0 +1,402 @@
...
+    override val monoid: Monoid[BF] = self.monoid
+
+    override def prepare(value: A): BF = Item(value)
+
+    override def present(bf: BF): BF = bf
+  }
+
+  implicit val equiv = new Equiv[BF] {
```
johnynek: will you add a type annotation to `equiv`

algebird-core/src/main/scala/com/twitter/algebird/immutable/BitSet.scala
```diff
@@ -0,0 +1,1172 @@
...
+ *
+ * Bitset treats the values it stores as 32-bit unsigned values, which
+ * is relevant to the internal addressing methods as well as the order
+ * used by `iterator`.
+ *
+ * The benchmarks suggest this bitset is MUCH faster than Scala's
+ * built-in bitset for cases where you may need many modifications and
+ * merges, (for example in a BloomFilter).
+ */
+sealed abstract class BitSet { lhs =>
```
johnynek, line 55: can we add the tests for this code in case people modify it later?

algebird-core/src/main/scala/com/twitter/algebird/experimental/BloomFilter.scala
```diff
@@ -0,0 +1,402 @@
...
+
+        target(hashIndex) = buffer(thisBidx) % w
+        nextHash(valueToHash, hashIndex + 1, buffer, thisBidx + 1, target)
+      }
+  }
+
+  /**
+   * Bloom Filter data structure
+   */
+  sealed abstract class BF extends Serializable {
```
johnynek: we have the opportunity to rename this, but that might complicate people adopting it. Should we have another name? Image, Hash, Sketch?

algebird-core/src/main/scala/com/twitter/algebird/experimental/BloomFilter.scala
```diff
@@ -0,0 +1,402 @@
...
+/**
+ * Bloom Filter - a probabilistic data structure to test presence of an element.
+ *
+ * Operations
+ *   1) insert: hash the value k times, updating the bitfield at the index equal to each hashed value
+ *   2) query: hash the value k times.  If there are k collisions, then return true; otherwise false.
+ *
+ * http://en.wikipedia.org/wiki/Bloom_filter
+ */
+final case class BloomFilter[A](n: Int, w: Int)(implicit hash: Hash128[A]) { self =>
```
johnynek: I think making `val hash` is a fine idea here, btw.

johnynek (commented):
Thanks for doing this.

So, is it right that the TL;DR is that it may be a bit slower but very close for all operations except folding/aggregating where it is considerably faster?

Would you agree this is the right take?

algebird-core/src/main/scala/com/twitter/algebird/experimental/BloomFilter.scala
regadas: Yup, i was trying to see if this moved some pref needles when JIT kicks in ... bit i would need to do further digging.

algebird-core/src/main/scala/com/twitter/algebird/experimental/BloomFilter.scala
regadas: `Hash` needs to Serializable but tbh is better if we just extend

algebird-core/src/main/scala/com/twitter/algebird/experimental/BloomFilter.scala
regadas: yup 👍

algebird-core/src/main/scala/com/twitter/algebird/experimental/BloomFilter.scala
regadas: There is 🤦

algebird-core/src/main/scala/com/twitter/algebird/experimental/BloomFilter.scala
regadas: Yikes ... one of those days you know?

algebird-core/src/main/scala/com/twitter/algebird/experimental/BloomFilter.scala
regadas: yeah good point!

algebird-core/src/main/scala/com/twitter/algebird/immutable/BitSet.scala
regadas: makes sense, I'll add them

algebird-core/src/main/scala/com/twitter/algebird/experimental/BloomFilter.scala
regadas: I'll add it, in this cases I tend to be ok with not having since it's type is clear... We should probably add this rule to scalafix to check for explicit result types.

algebird-core/src/main/scala/com/twitter/algebird/experimental/BloomFilter.scala
regadas: Still thinking on this one

regadas:
@johnynek yup that's the right take. If I'm not missing anything obvious we need dig a little bit more into this BitSet.

algebird-core/src/main/scala/com/twitter/algebird/experimental/BloomFilter.scala
johnynek: On the adoption question: you can always do an aliased import so if we are fixing warts I'm for fixing any that are significant. BF is a pretty bad name.

algebird-core/src/main/scala/com/twitter/algebird/experimental/BloomFilter.scala
```diff
@@ -0,0 +1,403 @@
...
+ * Bloom Filter - a probabilistic data structure to test presence of an element.
+ *
+ * Operations
+ *   1) insert: hash the value k times, updating the bitfield at the index equal to each hashed value
+ *   2) query: hash the value k times.  If there are k collisions, then return true; otherwise false.
+ *
+ * http://en.wikipedia.org/wiki/Bloom_filter
+ */
+final case class BloomFilter[A](n: Int, w: Int)(implicit val hash: Hash128[A]) { self =>
+  object Hash extends Serializable {
```
johnynek: Why do we have this object at all? Why not a method like hashToArray and not nested inside an object?

algebird-core/src/main/scala/com/twitter/algebird/experimental/BloomFilter.scala
```diff
@@ -0,0 +1,403 @@
...
+     * The number of bits set to true in the bloom filter
+     */
+    def numBits: Int
+
+    /**
+     * Proportion of bits that are set to true.
+     */
+    final val density: Double = numBits.toDouble / width
+
+    def ++(other: BF): BF
```
johnynek: BTW: def |(that: BF): BF might be a nice API to add. It could be an alias for ++ but it is closer to what the Monoid does.

regadas:
New numbers:

```
[info] BloomFilterCreateBenchmark.createBloomFilterUsingFoldExperimental                0.001            10000  avgt    3   14.570 ±  2.050  ms/op
[info] BloomFilterCreateBenchmark.createBloomFilterUsingFoldExperimental                0.001            50000  avgt    3  100.728 ± 10.937  ms/op
[info] BloomFilterCreateBenchmark.createBloomFilterUsingFoldExperimental                 0.01            10000  avgt    3   11.131 ±  1.336  ms/op
[info] BloomFilterCreateBenchmark.createBloomFilterUsingFoldExperimental                 0.01            50000  avgt    3   71.354 ±  1.632  ms/op
```

previous:
```
[info] BloomFilterCreateBenchmark.createBloomFilterUsingFoldExperimental                 0.001            10000  avgt    3    32.108 ±    0.386  ms/op
[info] BloomFilterCreateBenchmark.createBloomFilterUsingFoldExperimental                 0.001            50000  avgt    3   542.427 ±   97.306  ms/op
[info] BloomFilterCreateBenchmark.createBloomFilterUsingFoldExperimental                  0.01            10000  avgt    3    22.257 ±    0.485  ms/op
[info] BloomFilterCreateBenchmark.createBloomFilterUsingFoldExperimental                  0.01            50000  avgt    3   400.538 ±  989.902  ms/op
```

algebird-core/src/main/scala/com/twitter/algebird/experimental/BloomFilter.scala
regadas: I renamed it to `Hash`

algebird-core/src/main/scala/com/twitter/algebird/experimental/BloomFilter.scala
regadas: yeah good point! moved it to `hashToArray` 👍

algebird-core/src/main/scala/com/twitter/algebird/experimental/BloomFilter.scala
```diff
@@ -0,0 +1,413 @@
...
+
+    override def maybeContains(x: A): Boolean = item == x
+
+    override def size: Approximate[Long] = Approximate.exact[Long](1)
+  }
+
+  /*
+   * Bloom filter with multiple values
+   */
+  case class Instance(bits: BitSet) extends Hash {
```
johnynek: I think this is the only subclass we need here. The point of BitSet is that it is very fast. We don't need to allocate for an empty one.

We could have two cases, Item and Instance, and use Item to defer hashing until hopefully we have a number of them to combine at once.

on the other hand, there is a nice invariant here that this is always two or more items, we could even check that law in tests... though I guess if we merge an item with itself we do get a single element bitset.

maybe leave ignore all this and just update the comment to explain this has 1 or more bit set, and maybe we can check that property.

algebird-core/src/main/scala/com/twitter/algebird/experimental/BloomFilter.scala
```diff
@@ -0,0 +1,413 @@
...
+    override val monoid: Monoid[Hash] = self.monoid
+
+    override def prepare(value: A): Hash = Item(value)
+
+    override def present(bf: Hash): Hash = bf
+  }
+
+  implicit val equiv: Equiv[Hash] = new Equiv[Hash] {
+    override def equiv(a: Hash, b: Hash): Boolean =
+      (a eq b) || ((a.numHashes == b.numHashes) &&
```
johnynek: we don't need to check numHashes or width right? the path dependent type verifies that statically, no?

algebird-core/src/main/scala/com/twitter/algebird/experimental/BloomFilter.scala
```diff
@@ -0,0 +1,413 @@
...
+ * Bloom Filter - a probabilistic data structure to test presence of an element.
+ *
+ * Operations
+ *   1) insert: hash the value k times, updating the bitfield at the index equal to each hashed value
+ *   2) query: hash the value k times.  If there are k collisions, then return true; otherwise false.
+ *
+ * http://en.wikipedia.org/wiki/Bloom_filter
+ */
+final case class BloomFilter[A](n: Int, w: Int)(implicit val hash: Hash128[A]) { self =>
+  def hashToArray(s: A): Array[Int] = {
```
johnynek: lets add a comment here as to what is going on: what is the size of the result, etc...

algebird-core/src/main/scala/com/twitter/algebird/experimental/BloomFilter.scala
```diff
@@ -0,0 +1,413 @@
...
+  /**
+   * Create a bloom filter with multiple items.
+   */
+  def create(data: A*): Hash = create(data.iterator)
+
+  /**
+   * Create a bloom filter with multiple items from an iterator
+   */
+  def create(data: Iterator[A]): Hash = monoid.sum(data.map(Item))
+
```
johnynek: can we add a `def fromBitSet(bs: BitSet): Try[Hash]` and comment that this is how you do serialization: convert to BitSet, serialize that, and then on the other end, use this method?

The Try here verifies that the bitset maximum entry does not go beyond the size we would expect for this particular bloomfilter. Also, if it is empty, we just replace it with Empty.

johnynek (commented):
I'm sorry this slipped again.

A couple of minor points.

Also, I think we should just move this into algebird.immutable (not experimental). We have pretty good evidence this is going to be as good or better for virtually all users.

After we merge this, I can see making a separate algebird-legacy package that has things we want to remove, but at the same class name. So users who need it will need to update their build, but not their source code.

I'd like to remove the mutable bitset we are using from the build so we could have a scalajs build of algebird.

algebird-core/src/main/scala/com/twitter/algebird/experimental/BloomFilter.scala
```diff
@@ -0,0 +1,413 @@
...
+          val iter = t.iterator
+          var counter = 0
+          var bs = BitSet.newEmpty(0)
+
+          while (iter.hasNext) {
+            iter.next() match {
+              case Empty => ()
+              case bi @ Item(item) =>
+                bfItem = bi
+                val array = hashToArray(item)
```
johnynek: if we had a variant of hashToArray that did not allocate the result array, we could reuse the array. Like hashIntoArray(item, ary). That would likely make a real difference here since we can allocate one array at the beginning.

algebird-core/src/main/scala/com/twitter/algebird/experimental/BloomFilter.scala
```diff
@@ -0,0 +1,413 @@
...
+                val iter = bitset.iterator
+                while (iter.hasNext) {
+                  counter += 1
+                  bs = bs.mutableAdd(iter.next())
+                }
+            }
+          }
+
+          if (counter == 0) empty
+          else if (counter == n && (bfItem != null)) bfItem
```
johnynek: why bother with this since we've already hashed the item and done the work? Why not return `Instance(bs)` unconditionally? Also we could avoid keeping bfItem in that case so it is simpler.

algebird-core/src/main/scala/com/twitter/algebird/experimental/BloomFilter.scala
johnynek: also, maybe worth it to consider a private variant that does zero allocation so we can use that one in sumOption. I think that will be a significant win.

algebird-core/src/main/scala/com/twitter/algebird/experimental/BloomFilter.scala
regadas: Yup you are right we can simplify this.

algebird-core/src/main/scala/com/twitter/algebird/experimental/BloomFilter.scala
regadas: right! missed this

algebird-core/src/main/scala/com/twitter/algebird/experimental/BloomFilter.scala
regadas: Hi @johnynek Thanks for getting back to me! Here wouldn't we need to validate the `numHashes` as well?

johnynek (approved):
Thanks for making the changes.

algebird-core/src/main/scala/com/twitter/algebird/experimental/BloomFilter.scala
johnynek: We statically know the numHashes, don't we? So, the only dynamic thing we have when we are converting from a BitSet is the maximum set value in the BitSet. So, as long as that isn't invalid, we assume all is good, I guess...

algebird-core/src/main/scala/com/twitter/algebird/immutable/BloomFilter.scala
```diff
@@ -0,0 +1,432 @@
...
+   * if the BitSet has a maximum entry behond the BloomFilter expected size.
+   *
+   * This method will be helpfull on BloomFilter desirialization. Serialization is achieved
+   * through the serialization of the underlying [[BitSet]].
+   */
+  def fromBitSet(bitSet: BitSet): Try[Hash] =
+    if (bitSet.isEmpty) {
+      Success(empty)
+    } else {
+      if (bitSet.toSet.max > width)
```
johnynek: This is going to be pretty slow. Can we use reverseIterator.next() from bitSet instead?

johnynek (changes requested):

algebird-core/src/main/scala/com/twitter/algebird/immutable/BloomFilter.scala
regadas: You are right! we totally can! 🤦‍♂️ give me 1 sec

regadas: I overlooked that... that was embarrassing lol.

regadas:
@johnynek Addressed the issue before. After this I'll go ahead and move the other BF impl to a `algebird-legacy` as you proposed.

johnynek (approved):

johnynek:
Thanks for pushing this through!

PR #844, opened 2020-07-17
Remove implicit from MomentsGroup and other cleanups

johnynek:
While working on #829 @eigenvariable discovered that our tests were too weak for Moments and it is not really a group.

This PR is binary compatible, but removes the implicit label (which only breaks source compatibility and only for users unsafely using Moments as a Group).

Note: the code is not *totally* wrong and were were testing the laws, but we were not hitting the corner case that violates the law. The fact that it is nearly lawful (e.g. you don't ever have negative counts, it seems to be lawful) and the fact that we don't want to do a binary break is the reason why we are leaving the code.

I did two other kinds of changes:

1. implement sumOption for Moments and Correlation to speed up aggregations of big lists
2. remove some warnings that I could without breaking binary compatibility

@regadas and @eigenvariable could you take a look?

### DIFF
--- a/algebird-benchmark/src/main/scala/com/twitter/algebird/benchmark/HLLBenchmark.scala
+++ b/algebird-benchmark/src/main/scala/com/twitter/algebird/benchmark/HLLBenchmark.scala
@@ -2,7 +2,6 @@ package com.twitter.algebird
 package benchmark
 
 import scala.util.Random
-import com.twitter.bijection._
 
 import org.openjdk.jmh.annotations._
 import org.openjdk.jmh.infra.Blackhole
@@ -47,10 +46,9 @@ object HllBenchmark {
 
       val rng = new Random(3)
 
-      val byteEncoder = implicitly[Injection[Long, Array[Byte]]]
       def setSize = rng.nextInt(10) + 1 // 1 -> 10
       def hll(elements: Set[Long]): HLL =
-        hllMonoid.batchCreate(elements)(byteEncoder)
+        hllMonoid.sum(elements.map(hllMonoid.toHLL[Long](_)))
 
       val inputIntermediate = (0L until numElements).map { _ =>
         val setElements = (0 until setSize).map(_ => rng.nextInt(1000).toLong).toSet
--- a/algebird-benchmark/src/main/scala/com/twitter/algebird/benchmark/HLLPresentBenchmark.scala
+++ b/algebird-benchmark/src/main/scala/com/twitter/algebird/benchmark/HLLPresentBenchmark.scala
@@ -1,13 +1,10 @@
 package com.twitter.algebird.benchmark
 
 import com.twitter.algebird.{DenseHLL, HLL, HyperLogLogMonoid, SparseHLL}
-import com.twitter.bijection._
 import org.openjdk.jmh.annotations._
 import org.openjdk.jmh.infra.Blackhole
 
 object HLLPresentBenchmark {
-  implicit val byteEncoder = implicitly[Injection[Long, Array[Byte]]]
-
   @State(Scope.Benchmark)
   class HLLPresentState {
 
@@ -27,7 +24,7 @@ object HLLPresentBenchmark {
       val r = new scala.util.Random(12345L)
       data = (0 until numHLL).map { _ =>
         val input = (0 until max).map(_ => r.nextLong).toSet
-        hllMonoid.batchCreate(input)(byteEncoder.toFunction)
+        hllMonoid.sum(input.map(hllMonoid.toHLL(_)))
       }.toIndexedSeq
     }
   }
--- a/algebird-core/src/main/scala/com/twitter/algebird/CorrelationMonoid.scala
+++ b/algebird-core/src/main/scala/com/twitter/algebird/CorrelationMonoid.scala
@@ -120,7 +120,7 @@ object CorrelationMonoid extends Monoid[Correlation] {
    */
   override def plus(a: Correlation, b: Correlation): Correlation = {
     val count = a.totalWeight + b.totalWeight
-    if (count == 0)
+    if (count == 0.0)
       CorrelationMonoid.zero
     else {
       val prodSumRatio = a.totalWeight * b.totalWeight / count
@@ -141,11 +141,67 @@ object CorrelationMonoid extends Monoid[Correlation] {
   }
 
   override val zero = Correlation(0, 0, 0, 0, 0, 0)
+
+  override def sumOption(cs: TraversableOnce[Correlation]): Option[Correlation] =
+    if (cs.isEmpty) None
+    else {
+      val iter = cs.toIterator
+      val item = iter.next()
+
+      var m0 = item.m0
+      var m1y = item.m1y
+      var m1x = item.m1x
+      var m2y = item.m2y
+      var m2x = item.m2x
+      var c2 = item.c2
+
+      while (iter.hasNext) {
+
+        /**
+         * This is tested by monoidLaws to match plus
+         * we do this loop here to avoid allocating
+         * between each pair of Correlations
+         */
+        val b = iter.next()
+        val m0New = m0 + b.m0
+
+        if (m0New == 0.0) {
+          m1y = 0.0
+          m1x = 0.0
+          m2y = 0.0
+          m2x = 0.0
+          c2 = 0.0
+        } else {
+          val prodSumRatio = m0 * b.m0 / m0New
+
+          val m1xNew = Correlation.getCombinedMean(m0, m1x, b.m0, b.m1x)
+          val m1yNew = Correlation.getCombinedMean(m0, m1y, b.m0, b.m1y)
+          val deltaX = b.m1x - m1x
+          val deltaY = b.m1y - m1y
+
+          val m2xNew = m2x + b.m2x + math.pow(deltaX, 2) * prodSumRatio
+          val m2yNew =
+            m2y + b.m2y + math.pow(deltaY, 2) * prodSumRatio
+
+          val c2New = c2 + b.c2 + deltaX * deltaY * prodSumRatio
+
+          m1y = m1yNew
+          m1x = m1xNew
+          m2y = m2yNew
+          m2x = m2xNew
+          c2 = c2New
+        }
+        m0 = m0New
+      }
+
+      if (m0 == 0.0) Some(zero)
+      else Some(Correlation(c2 = c2, m2x = m2x, m2y = m2y, m1x = m1x, m1y = m1y, m0 = m0))
+    }
 }
 
 object CorrelationAggregator extends MonoidAggregator[(Double, Double), Correlation, Correlation] {
   override def prepare(a: (Double, Double)): Correlation = Correlation(a)
-  override val monoid = CorrelationMonoid
+  override def monoid: Monoid[Correlation] = CorrelationMonoid
   override def present(c: Correlation): Correlation = c
 
   def correlation: MonoidAggregator[(Double, Double), Correlation, Double] =
--- a/algebird-core/src/main/scala/com/twitter/algebird/CorrelationMonoid.scala
+++ b/algebird-core/src/main/scala/com/twitter/algebird/CorrelationMonoid.scala
@@ -157,7 +157,7 @@ object CorrelationMonoid extends Monoid[Correlation] {
 
       while (iter.hasNext) {
 
-        /**
+        /*
          * This is tested by monoidLaws to match plus
          * we do this loop here to avoid allocating
          * between each pair of Correlations
--- a/algebird-core/src/main/scala/com/twitter/algebird/HyperLogLog.scala
+++ b/algebird-core/src/main/scala/com/twitter/algebird/HyperLogLog.scala
@@ -737,7 +737,8 @@ abstract class SetSizeAggregatorBase[A](hllBits: Int, maxSetSize: Int)
 
 case class SetSizeAggregator[A](hllBits: Int, maxSetSize: Int = 10)(implicit toBytes: A => Array[Byte])
     extends SetSizeAggregatorBase[A](hllBits, maxSetSize) {
-  override def convert(set: Set[A]): HLL = leftSemigroup.batchCreate(set.map(toBytes))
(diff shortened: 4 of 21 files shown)


algebird-benchmark/src/main/scala/com/twitter/algebird/benchmark/HLLBenchmark.scala
```diff
@@ -47,10 +46,9 @@ object HllBenchmark {
 
       val rng = new Random(3)
 
-      val byteEncoder = implicitly[Injection[Long, Array[Byte]]]
       def setSize = rng.nextInt(10) + 1 // 1 -> 10
       def hll(elements: Set[Long]): HLL =
-        hllMonoid.batchCreate(elements)(byteEncoder)
+        hllMonoid.sum(elements.map(hllMonoid.toHLL[Long](_)))
```
johnynek, line 51: this is to remove the deprecation warning on using batchCreate

algebird-benchmark/src/main/scala/com/twitter/algebird/benchmark/HLLPresentBenchmark.scala
```diff
@@ -27,7 +24,7 @@ object HLLPresentBenchmark {
       val r = new scala.util.Random(12345L)
       data = (0 until numHLL).map { _ =>
         val input = (0 until max).map(_ => r.nextLong).toSet
-        hllMonoid.batchCreate(input)(byteEncoder.toFunction)
+        hllMonoid.sum(input.map(hllMonoid.toHLL(_)))
```
johnynek, line 27: remove deprecation warning on batchCreate

algebird-core/src/main/scala/com/twitter/algebird/HyperLogLog.scala
```diff
@@ -737,7 +737,8 @@ abstract class SetSizeAggregatorBase[A](hllBits: Int, maxSetSize: Int)
 
 case class SetSizeAggregator[A](hllBits: Int, maxSetSize: Int = 10)(implicit toBytes: A => Array[Byte])
     extends SetSizeAggregatorBase[A](hllBits, maxSetSize) {
-  override def convert(set: Set[A]): HLL = leftSemigroup.batchCreate(set.map(toBytes))
+  override def convert(set: Set[A]): HLL =
+    leftSemigroup.sum(set.iterator.map(a => leftSemigroup.toHLL(toBytes(a))))
```
johnynek, line 741: remove deprecation warning on batchCreate

algebird-core/src/main/scala/com/twitter/algebird/MomentsGroup.scala
```diff
@@ -57,8 +57,13 @@ case class Moments(m0: Long, m1: Double, m2: Double, m3: Double, m4: Double) {
 }
 
 object Moments {
-  implicit val group: Group[Moments] with CommutativeGroup[Moments] =
+  @deprecated("use monoid[Moments], this isn't lawful for negate", "0.13.8")
+  def group: Group[Moments] with CommutativeGroup[Moments] =
```
johnynek, line 61: we are removing `implicit` but that is not a problem for binary compatibility (and this is checked by mima)

algebird-core/src/main/scala/com/twitter/algebird/MomentsGroup.scala
```diff
@@ -108,35 +110,108 @@ object MomentsGroup extends Group[Moments] with CommutativeGroup[Moments] {
...
-    Moments(-a.count, a.m1, -a.m2, -a.m3, -a.m4)
-
   // Combines the moment calculations from two streams.
   // See http://en.wikipedia.org/wiki/Algorithms_for_calculating_variance#Higher-order_statistics
   // for more information on the formulas used to update the moments.
   override def plus(a: Moments, b: Moments): Moments = {
-    val delta = b.mean - a.mean
     val countCombined = a.count + b.count
-    if (countCombined == 0)
-      return zero
```
johnynek, line 121: remove the use of return and move `delta` below the else in the case we don't need it.

algebird-core/src/main/scala/com/twitter/algebird/Operators.scala
```diff
@@ -16,23 +16,36 @@ limitations under the License.
...
+  def toMinus[T: Group](t: T): MinusOp[T] = new MinusOp(t)
+  @deprecated("use Operators.Ops", "0.13.8")
+  def toTimes[T: Ring](t: T): TimesOp[T] = new TimesOp(t)
+
   implicit def toRichTraversableFromIterator[T](t: Iterator[T]): RichTraversable[T] =
     new RichTraversable(t)
   implicit def toRichTraversable[T](t: Traversable[T]): RichTraversable[T] =
     new RichTraversable(t)
+
+  implicit class Ops[A](private val a: A) extends AnyVal {
```
johnynek, line 31: this pattern is better because it doesn't have to allocate. The `PlusOp` etc pattern has to allocate because it captures the typeclass in the constructor vs the method callsite.

algebird-test/src/test/scala/com/twitter/algebird/CorrelationLaws.scala
```diff
@@ -75,7 +75,9 @@ class CorrelationLaws extends CheckProperties {
   property("the swap method on moments works as you'd think") {
     forAll { l: List[(Double, Double)] =>
       val swapped = CorrelationAggregator(l).swap
-      val reversedInput = CorrelationAggregator.composePrepare[(Double, Double)] { case (x, y) => (y, x) }(l)
+      val fn: ((Double, Double)) => (Double, Double) = { tup => tup.swap }
```
johnynek, line 78: just wanted to use `Tuple2.swap` here for which the `Correlation` method is named and show that they are the same. The previous code correctly inlined the implementation of swap, but I thought highlighting the method name equivalence was nice.

algebird-test/src/test/scala/com/twitter/algebird/HyperLogLogTest.scala
```diff
@@ -362,7 +362,10 @@ class HyperLogLogTest extends AnyWordSpec with Matchers {
       val partialSums = data.foldLeft(IndexedSeq(mon.zero)) { (seq, value) =>
         seq :+ (seq.last + mon.create(value))
       }
-      (1 to 200).map(n => assert(partialSums(n) == mon.batchCreate(data.slice(0, n))))
```
johnynek, line 365: remove deprecation warning on batchCreate

algebird-test/src/test/scala/com/twitter/algebird/MomentsLaws.scala
```diff
@@ -1,24 +1,48 @@
...
+    val init: Gen[Moments] = genA.map(Moments(_))
+
+    val recur = Gen.lzy(opBasedGen[A](genA))
+    val pair = Gen.zip(recur, recur)
+
+    import Operators.Ops
+
+    Gen.frequency(
+      (10, init),
+      (1, pair.map { case (a, b) => a + b })
```
johnynek, line 30: this is testing the API as the user would use it. It avoids making highly unusual Moments which might hide laws passing (as happened with Group tests previously).

There are really only two things you can do: create a Moments from a value, and combine it. This is generating both kinds.

algebird-test/src/test/scala/com/twitter/algebird/MomentsLaws.scala
```diff
@@ -1,24 +1,48 @@
...
-      Equiv.fromFunction { (ml, mr) =>
-        (ml.m0 == mr.m0) &&
-        approxEq(EPS)(ml.m1, mr.m1) &&
-        approxEq(EPS)(ml.m2, mr.m2) &&
-        approxEq(EPS)(ml.m3, mr.m3) &&
-        approxEq(EPS)(ml.m4, mr.m4)
-      }
+    import com.twitter.algebird.scalacheck.arbitrary.momentsArb
+    implicit val group: Group[Moments] = MomentsGroup
     groupLaws[Moments]
```
johnynek, line 37: these laws continue to (usually) pass when we use the old generator. If you use a generator that includes generating pairs and subtracting them it almost always fails.

build.sbt
```diff
@@ -203,12 +203,12 @@ lazy val mimaSettings = Def.settings(
...
  * the current.
  */
-val noBinaryCompatCheck = Set[String]("benchmark", "caliper", "generic", "spark")
+val noBinaryCompatCheck = Set[String]("benchmark", "caliper", "spark")
 
 def previousVersion(subProj: String) =
   Some(subProj)
     .filterNot(noBinaryCompatCheck.contains)
-    .map(s => "com.twitter" %% ("algebird-" + s) % "0.13.5")
+    .map(s => "com.twitter" %% ("algebird-" + s) % "0.13.7")
```
johnynek, line 211: we forgot to bump this a couple of times...

algebird-core/src/main/scala/com/twitter/algebird/CorrelationMonoid.scala
```diff
@@ -141,11 +141,67 @@ object CorrelationMonoid extends Monoid[Correlation] {
...
+
+      if (m0 == 0.0) Some(zero)
+      else Some(Correlation(c2 = c2, m2x = m2x, m2y = m2y, m1x = m1x, m1y = m1y, m0 = m0))
+    }
 }
 
 object CorrelationAggregator extends MonoidAggregator[(Double, Double), Correlation, Correlation] {
   override def prepare(a: (Double, Double)): Correlation = Correlation(a)
-  override val monoid = CorrelationMonoid
+  override def monoid: Monoid[Correlation] = CorrelationMonoid
```
eigenvariable, line 204: OOC, why keep this a `def` when you're not going to extend an `object`? Serialization?

algebird-core/src/main/scala/com/twitter/algebird/MomentsGroup.scala
```diff
@@ -108,35 +110,108 @@ object MomentsGroup extends Group[Moments] with CommutativeGroup[Moments] {
...
+          math.pow(b.count, 2) * a.m2) / math.pow(countCombined, 2) +
+        4 * delta * (a.count * b.m3 - b.count * a.m3) / countCombined
+
+      Moments(countCombined, meanCombined, m2, m3, m4)
+    }
+  }
 
-    val m2 = a.m2 + b.m2 +
-      math.pow(delta, 2) * a.count * b.count / countCombined
+  override def sumOption(items: TraversableOnce[Moments]): Option[Moments] =
```
eigenvariable, line 141: Is there any downside to `plus` being defined in terms of `sumOption`?

eigenvariable (approved):
This all looks peaceful to me, including the stuff that makes things more imperative for efficiency reasons. I'm going to trust that those all have good test coverage already.

Also, it's interesting to note that removing `implicit` maintains binary compatibility but not source-compatible. I'd never thought about that before, and will have to update my mental model (previously I'd always thought of binary compatibility as being a strictly stronger criterion than source compatibility).

algebird-core/src/main/scala/com/twitter/algebird/CorrelationMonoid.scala
johnynek: val always makes a field on the object. But here we don't need a field just a method to point to the object.

algebird-core/src/main/scala/com/twitter/algebird/MomentsGroup.scala
johnynek: We would need to allocate a wrapper for the pair. My philosophy with FP has changed a bit: make pure and immutable APIs that are easy to reason about but underneath use any (sound) trick you can find to improve performance.

Maybe it's often unneeded but seeing massive differences in performance and having my code run at huge scale made me look for this more.

build.sbt
regadas: nice catch! Now mima is complaining about the addition of the overrides to Double/FloatRing (type was previously erased). It should be ok to just add 2 exclusion rules here.

regadas (approved):
LGTM

build.sbt
regadas: I think it might be possible to update this versions dynamically

algebird-core/src/main/scala/com/twitter/algebird/Ring.scala
```diff
@@ -143,6 +165,19 @@ object DoubleRing extends Ring[Double] {
   override def plus(l: Double, r: Double): Double = l + r
   override def minus(l: Double, r: Double): Double = l - r
   override def times(l: Double, r: Double): Double = l * r
+  override def sum(t: TraversableOnce[Double]): Double = {
```
johnynek: this is a binary incompatible change it seems.

algebird-core/src/main/scala/com/twitter/algebird/Ring.scala
```diff
@@ -134,6 +143,19 @@ object FloatRing extends Ring[Float] {
   override def plus(l: Float, r: Float): Float = l + r
   override def minus(l: Float, r: Float): Float = l - r
   override def times(l: Float, r: Float): Float = l * r
+  override def sum(t: TraversableOnce[Float]): Float = {
```
johnynek: looks like this is binary incompatible

algebird-core/src/main/scala/com/twitter/algebird/Ring.scala
```diff
@@ -143,19 +143,19 @@ object FloatRing extends Ring[Float] {
   override def plus(l: Float, r: Float): Float = l + r
   override def minus(l: Float, r: Float): Float = l - r
   override def times(l: Float, r: Float): Float = l * r
-  override def sum(t: TraversableOnce[Float]): Float = {
```
regadas: Type being erasured is a bummer and this is probably for the best ... however i reckon that at some we should nail these and release a major version. It feels that the API is becoming a bit incoherent in come cases.

johnynek: yeah, maybe we should have an issue to track the wins... that said, I still think minimizing source breaks to removing real bugs is a good goal.

PR #846, opened 2020-07-20
Add explicit return type rule

regadas:


algebird-core/src/main/scala/com/twitter/algebird/Eventually.scala
```diff
@@ -130,7 +130,7 @@ class EventuallyMonoid[E, O](convert: O => E)(mustConvert: O => Boolean)(implici
 ) extends EventuallySemigroup[E, O](convert)(mustConvert)
     with Monoid[Either[E, O]] {
 
-  override def zero = Right(Monoid.zero[O])
+  override def zero = Right(Monoid.zero[O]) // scalafix:ok
```
regadas: some of these need to be ignored ... inferred type is `Right` and it will break bin-compat

algebird-core/src/main/scala/com/twitter/algebird/DecayedVector.scala
```diff
@@ -51,13 +53,16 @@ object DecayedVector extends CompatDecayedVector {
...
+      vs: VectorSpace[Double, ({ type x[a] = Map[K, a] })#x],
+      metric: Metric[Map[K, Double]]
+  ): Monoid[DecayedVector[({ type x[a] = Map[K, a] })#x]] =
     monoidWithEpsilon[({ type x[a] = Map[K, a] })#x](eps)
 
   implicit def mapMonoid[K](implicit
       vs: VectorSpace[Double, ({ type x[a] = Map[K, a] })#x],
       metric: Metric[Map[K, Double]]
-  ) =
+  ): Monoid[DecayedVector[({ type x[a] = Map[K, a] })#x]] =
```
regadas, line 65: On q follow up PR maybe we can bring kind-projector to make these a little nice.

algebird-core/src/main/scala/com/twitter/algebird/Aggregator.scala
```diff
@@ -495,7 +495,7 @@ trait Aggregator[-A, B, +C] extends java.io.Serializable { self =>
     new MonoidAggregator[A, Option[B], Option[C]] {
       override def prepare(input: A): Option[B] = Some(self.prepare(input))
       override def present(reduction: Option[B]): Option[C] = reduction.map(self.present)
-      override def monoid = new OptionMonoid[B]()(self.semigroup)
+      override def monoid: OptionMonoid[B] = new OptionMonoid[B]()(self.semigroup)
```
johnynek: can we relax this to `Monoid[Option[B]]`? I hate to make the types needlessly tight.

algebird-core/src/main/scala/com/twitter/algebird/Aggregator.scala
```diff
@@ -471,7 +471,7 @@ trait Aggregator[-A, B, +C] extends java.io.Serializable { self =>
     val ag1 = this
     new Aggregator[(A, A2), (B, B2), (C, C2)] {
       override def prepare(a: (A, A2)): (B, B2) = (ag1.prepare(a._1), ag2.prepare(a._2))
-      override val semigroup = new Tuple2Semigroup()(ag1.semigroup, ag2.semigroup)
+      override val semigroup: Tuple2Semigroup[B, B2] = new Tuple2Semigroup()(ag1.semigroup, ag2.semigroup)
```
johnynek: can we relax to `Semigroup[(B, B2)]` I had to make these too tight.

algebird-core/src/main/scala/com/twitter/algebird/Aggregator.scala
```diff
@@ -572,7 +572,7 @@ trait MonoidAggregator[-A, B, +C] extends Aggregator[A, B, C] { self =>
         case Left(a)   => (self.prepare(a), that.monoid.zero)
         case Right(a2) => (self.monoid.zero, that.prepare(a2))
       }
-      override val monoid = new Tuple2Monoid[B, B2]()(self.monoid, that.monoid)
+      override val monoid: Tuple2Monoid[B, B2] = new Tuple2Monoid[B, B2]()(self.monoid, that.monoid)
```
johnynek: loosen?

algebird-core/src/main/scala/com/twitter/algebird/Aggregator.scala
```diff
@@ -620,7 +620,7 @@ trait MonoidAggregator[-A, B, +C] extends Aggregator[A, B, C] { self =>
     val ag1 = self
     new MonoidAggregator[(A, A2), (B, B2), (C, C2)] {
       override def prepare(a: (A, A2)): (B, B2) = (ag1.prepare(a._1), ag2.prepare(a._2))
-      override val monoid = new Tuple2Monoid[B, B2]()(ag1.monoid, ag2.monoid)
+      override val monoid: Tuple2Monoid[B, B2] = new Tuple2Monoid[B, B2]()(ag1.monoid, ag2.monoid)
```
johnynek: loosen?

algebird-core/src/main/scala/com/twitter/algebird/BloomFilter.scala
```diff
@@ -293,10 +293,10 @@ object BF {
                       uniq(src, dst, cur, spos + 1, dpos + 1)
                     }
                   }
-                val uniqVs = new Array[Int](hashvalues.length)
+                val uniqVs: Array[Int] = new Array[Int](hashvalues.length)
                 val len: Int = uniq(hashvalues, uniqVs, -1, 0, 0)
 
-                var pos = 0
+                var pos: Int = 0
```
johnynek: this is a bummer. Can we make a change that doesn't require the annotations on private types? This cannot be seen from the outside since we are returning an IntIterator

algebird-core/src/main/scala/com/twitter/algebird/DecayedVector.scala
johnynek: be happy to, especially now that the syntax is set for scala 3

algebird-core/src/main/scala/com/twitter/algebird/Eventually.scala
johnynek: wait, why not? why can't we write the type inferred? has to be either `Right[E, O]` or `Either[E, O]`. Both are fine types for zero if you ask me.

algebird-core/src/main/scala/com/twitter/algebird/Eventually.scala
```diff
@@ -162,7 +162,7 @@ class EventuallyRing[E, O](convert: O => E)(mustConvert: O => Boolean)(implicit
 ) extends EventuallyGroup[E, O](convert)(mustConvert)
     with Ring[Either[E, O]] {
 
-  override def one = Right(Ring.one[O])
+  override def one = Right(Ring.one[O]) // scalafix:ok
```
johnynek: same comment as above.

algebird-core/src/main/scala/com/twitter/algebird/Eventually.scala
```diff
@@ -181,7 +181,7 @@ class EventuallyRing[E, O](convert: O => E)(mustConvert: O => Boolean)(implicit
 }
 
 trait AbstractEventuallyAggregator[A, E, O, C] extends Aggregator[A, Either[E, O], C] {
-  override def prepare(a: A) = Right(rightAggregator.prepare(a))
+  override def prepare(a: A) = Right(rightAggregator.prepare(a)) // scalafix:ok
```
johnynek: same comment as above.

algebird-core/src/main/scala/com/twitter/algebird/Eventually.scala
```diff
@@ -199,7 +199,7 @@ trait AbstractEventuallyAggregator[A, E, O, C] extends Aggregator[A, Either[E, O
 trait EventuallyAggregator[A, E, O, C] extends AbstractEventuallyAggregator[A, E, O, C] {
 
   //avoid init order issues and cyclical references
-  @transient override lazy val semigroup =
+  @transient override lazy val semigroup: EventuallySemigroup[E, O] =
```
johnynek, line 202: can we loosen this to `Semigroup[Either[E, O]]`?

algebird-core/src/main/scala/com/twitter/algebird/Eventually.scala
```diff
@@ -209,6 +209,6 @@ trait EventuallyMonoidAggregator[A, E, O, C]
 
   override def rightAggregator: MonoidAggregator[A, O, C]
 
-  @transient override lazy val monoid =
+  @transient override lazy val monoid: EventuallyMonoid[E, O] =
```
johnynek, line 212: can we loosen to `Monoid[Either[E, O]]`?

johnynek (commented):
thanks for doing this!

algebird-core/src/main/scala/com/twitter/algebird/Eventually.scala
regadas: ahh sorry i  probably explained myself badly ... yeah we can make it `override def one: Right[E, O] = Right(Ring.one[O])` but not `override def one: Either[E, O] = Right(Ring.one[O])`. I felt it was to tight ... but on a second thought maybe it's ok for these.

algebird-core/src/main/scala/com/twitter/algebird/Eventually.scala
regadas: yup yup

regadas:
@johnynek gonna do a snd round of coz I'm seeing that i missed a lot of them and loosen a few more 👍

algebird-core/src/main/scala/com/twitter/algebird/BloomFilter.scala
regadas: good question ... maybe? I'll have a look

algebird-core/src/main/scala/com/twitter/algebird/Eventually.scala
johnynek: actually, I think promising the zero/one are Right is actually good (and true).

regadas:
@johnynek Tweaked a little bit scalafix and managed to cover ~most~ some of the unwanted cases. Regarding the types that we would want to loosen up ... that will not fly ... mima will report that it's not bin-compat.

johnynek:
Thanks for doing this!

PR #849, opened 2020-07-22
Add kind-projector

regadas:


johnynek (approved):
Thanks!

regadas:
@johnynek thanks for taking a look! 

PS: I'll revise CI here as well

johnynek:
looks like running tut to build the documentation failed, I think. Maybe the plugins somehow didn't get passed through to something there?

Side note: if we are cleaning things up, using mdoc would be nice at some point in the future.

johnynek:
This is great, thanks!

The next release is shaping up nicely.

PR #850, opened 2020-07-23
Add scale method to Moments

johnynek:
This is a follow up to #829 which adds the scale method to Moments.

I had thought we would have to wait to break binary compatibility, but by writing the case class methods by hand, we can actually make the internal state a double and add the scale method to Moments.

This offers the ability to do exponentially decayed moments, which can be useful to aggregate over long time scales but not have very old values dominate the results.

It's a bit ugly to write this code out by hand, but binary compatibility is worth it, IMO.

PTAL @eigenvariable @regadas


algebird-core/src/main/scala/com/twitter/algebird/MomentsGroup.scala
```diff
@@ -24,36 +24,84 @@ import algebra.{CommutativeGroup, CommutativeMonoid}
...
   // Population standard deviation, not sample standard deviation.
   def stddev: Double = math.sqrt(variance)
 
   def skewness: Double =
-    if (count > 2)
-      math.sqrt(count) * m3 / math.pow(m2, 1.5)
-    else
-      /* don't return junk when the moment is not defined */
-      Double.NaN
+    math.sqrt(m0D) * m3 / math.pow(m2, 1.5)
```
eigenvariable, line 56: No longer filtering on `count > 2` is a semantic change that's probably worth at least calling out in the release notes or something.

eigenvariable (commented):
Another approach that might be worth exploring: declare a new version of moments where `count` is a `Double` instead of a `Long`, with a trait that implements all of the shared business logic between both versions in terms of a `Numeric` (or whatever the best mechanism is for making this generic). Overriding methods like `equals` is just kind of a lot here.

johnynek:
Yeah. I could just copy all the code and leave the old one for binary compatibility but I don't really want to have to two copies in the repository. The mima check shows we needed to make all these changes to support binary compatibility.

It really wasn't that much work to implement everything.

algebird-core/src/main/scala/com/twitter/algebird/MomentsGroup.scala
johnynek: indeed it is. I will call it out, but since we returned NaN before, we were already in a "silent-ish" failure more.

eigenvariable:
I don't doubt that this was straightforward to write, but it's pretty difficult to read and I suspect maintain. Ultimately, algebird is your baby more so than anyone else's, and IMO you've earned the right to make these sorts of aesthetic decisions for the library. But my 2¢ is that this would all be much cleaner by just defining a `ScalableMoments` class and sharing as much code as possible between the two.

algebird-core/src/main/scala/com/twitter/algebird/MomentsGroup.scala
```diff
@@ -24,36 +24,84 @@ import algebra.{CommutativeGroup, CommutativeMonoid}
...
+ * This code manually inlines code to make it look like a case class. This is done
+ * because we changed the count from a Long to a Double to enable the scale method,
+ * which allows exponential decays of moments, but we didn't want to break backwards
+ * binary compatibility.
  */
-case class Moments(m0: Long, m1: Double, m2: Double, m3: Double, m4: Double) {
+sealed class Moments(val m0D: Double, val m1: Double, val m2: Double, val m3: Double, val m4: Double)
+    extends Product
+    with Serializable {
+  def this(m0: Long, m1: Double, m2: Double, m3: Double, m4: Double) {
```
regadas:
```suggestion
  def this(m0: Long, m1: Double, m2: Double, m3: Double, m4: Double) = {
```

algebird-core/src/main/scala/com/twitter/algebird/MomentsGroup.scala
```diff
@@ -94,6 +188,8 @@ class MomentsMonoid extends Monoid[Moments] with CommutativeMonoid[Moments] {
    * Uses a more stable online algorithm which should be suitable for
    * large numbers of records similar to:
    * http://en.wikipedia.org/wiki/Algorithms_for_calculating_variance#Parallel_algorithm
+   *
+   * we no longer use this, but we can't remove it due to binary compatibility
    */
   def getCombinedMean(n: Long, an: Double, k: Long, ak: Double): Double =
```
regadas, line 195: should we add a `@deprecated` as well?

regadas, line 195:
```suggestion
  @deprecated("Use Moments.getCombinedMeanDouble instead", since = "0.13.8")
  def getCombinedMean(n: Long, an: Double, k: Long, ak: Double): Double =
```

regadas: @johnynek wdyt?

johnynek:
@regadas merged your suggestion, sorry for being slow. I merged develop and lets make sure CI is green, then it would be nice to merge.

regadas:
@johnynek awesome! yeah don't worry about it!

regadas (approved):

PR #867, opened 2020-09-18
Setup automated releases from GitHub Actions CI

olafurpg:
Previously, releases were done from a local computer with the
sbt-release plugin. Now, releases happen automatically from GitHub
Actions. Every merge into master should trigger a SNAPSHOT release
and every git tag push should trigger a stable Sonatype release.

For more details, see https://github.com/olafurpg/sbt-ci-release

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -1,0 +1,19 @@
+name: Release
+on:
+  push:
+    branches: [develop]
+    tags: ["*"]
+jobs:
+  publish:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v1
+      - uses: olafurpg/setup-scala@v7
+      - uses: olafurpg/setup-gpg@v2
+      - name: Publish ${{ github.ref }}
+        run: sbt ci-release
+        env:
+          PGP_PASSPHRASE: ${{ secrets.PGP_PASSPHRASE }}
+          PGP_SECRET: ${{ secrets.PGP_SECRET }}
+          SONATYPE_PASSWORD: ${{ secrets.SONATYPE_PASSWORD }}
+          SONATYPE_USERNAME: ${{ secrets.SONATYPE_USERNAME }}
--- a/version.sbt
+++ b/version.sbt
@@ -1,1 +1,0 @@
-version in ThisBuild := "0.13.8-SNAPSHOT"
--- a/build.sbt
+++ b/build.sbt
@@ -1,4 +1,3 @@
-import ReleaseTransformations._
 import algebird._
 import com.typesafe.tools.mima.core._
 import pl.project13.scala.sbt.JmhPlugin
@@ -34,6 +33,9 @@ def scalaBinaryVersion(scalaVersion: String) = scalaVersion match {
 
 def isScala212x(scalaVersion: String) = scalaBinaryVersion(scalaVersion) == "2.12"
 def isScala213x(scalaVersion: String) = scalaBinaryVersion(scalaVersion) == "2.13"
+
+noPublishSettings
+crossScalaVersions := Nil
 
 val sharedSettings = Seq(
   organization := "com.twitter",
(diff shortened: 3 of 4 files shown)


regadas (approved):
This is super nice! 🎉 I guess this just speeds up the process of moving completely to GHA as we did in chill.

Now @posco just needs to add a few secrets to the repo! 🚀 (I don't have enough perms)

olafurpg:
I just received admin access for this repo and configured the Sonatype+PGP secrets. It may take several attempts after this is merged to validate that everything is working as expected.

> I guess this just speeds up the process of moving completely to GHA as we did in chill.

Yes, I think it would be a nice followup PR to move all the CI to GitHub Actions.

olafurpg:
I'm planning to followup with similar changes in the chill and bijection repos.

version.sbt
```diff
@@ -1 +0,0 @@
-version in ThisBuild := "0.13.8-SNAPSHOT"
```
olafurpg, line 1: This changes the local value of `version` to be formatted like this 0.13.7+34-2d333018-SNAPSHOT. Is that an acceptable change? I think it's possible to get configure the version to not include the git describe information you prefer to keep the old behavior.

regadas: imo i think it's ok.

regadas:
Yeah following up with a PR would be nice... I've been meaning to do this for a long time lol

.github/workflows/release.yml
```diff
@@ -0,0 +1,19 @@
...
+on:
+  push:
+    branches: [develop]
+    tags: ["*"]
+jobs:
+  publish:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v1
+      - uses: olafurpg/setup-scala@v7
```
regadas, line 11: One i thing i realised is that we can add some artifact cache here

```
      - name: cache SBT
        uses: coursier/cache-action@v3
```

olafurpg: This job only runs the releases, which I think is desirable to run with a clean machine. My experience is that GitHub Actions is usually fast enough for most projects without caching

regadas: right, anecdotally I've seen it fail a few times ¯\_(ツ)_/¯ I guess we can always re-trigger?

olafurpg: You can always rerun the job or force push the git tag

regadas: right that's what I meant. I understand that this is probably a preference thing but i don't feel that fetching artifacts every time offer a lot of guarantees. If a build is green on `develop` for a given hash, running the release job (`develop` or `tag`) should mean that it will equally build and not be subjective to transient errors. Publishing to sonatype should be the only thing that can fail (which fails quite often unfortunately)

johnynek (approved):
This is great and super exciting!

One thing we probably want to address: the CI is a bit flaky on two kinds of tests: 1. approximate laws related to floating point numbers, 2. tests checking probabilistic bounds (this is true with Prob > y, which we run many trials and see if the result is highly improbable given that).

We should probably tune it so that those rarely happen.

olafurpg:
Merging to see what happens on master. Small heads up, it usually takes a few failed attempts before it works correctly

olafurpg:
The SNAPSHOT release worked, you can download the artifacts with
```
cs fetch com.twitter:algebird-core_2.11:0.13.7+37-fd6454ad-SNAPSHOT -r sonatype:snapshots
```
The next step is to try a stable release by pushing a git tag. Any objections if I push a v0.13.8-M1 tag to see if everything works?

FWIW, I have a `git-release` script in my `$PATH` 
```
❯ cat `which git-release`
set -eux
version=$1
git tag -af "v$version" -m "v$version" && git push -f origin v$version
```

I normally cut releases by running `git release 0.13.8` from the terminal. I usually wait until the CI release job has completed before publishing a release on GitHub.

olafurpg:
> One thing we probably want to address: the CI is a bit flaky on two kinds of tests: 1. approximate laws related to floating point numbers, 2. tests checking probabilistic bounds (this is true with Prob > y, which we run many trials and see if the result is highly improbable given that).

The CI release currently runs the publish step without testing so flaky failures should not block the release. It's possible to add custom steps before `+publishSigned` in case you prefer that, but when I tried this in the past I normally ended up spending more time dealing with incomplete releases than catching actual bugs before a release (the master branch has presumably gone through PR code review).

regadas:
>  v0.13.8-M1

we can go with a v0.13.8 since there are dependency updates, #844, #829 and some other minor things.

> It's possible to add  custom steps before +publishSigned

Yeah we can always just add `sbt +test` if needed but +1 on leaving it as is for now.

regadas:
Let's include #850 as well

regadas:
I think we are good for a new release `v0.13.8`. 

who should do the honours? I can prepare the release notes if you want /cc @johnynek @olafurpg

johnynek:
One argument is the honors should go to @olafurpg who once again did the important work of making releases easier, and helping remove friction from the ecosystem. My second nomination would be @regadas who has really stepped up to do a lot of work managing these libraries. For the first time you will be able to release since you don't need the twitter sonatype credentials.

I'm happy to do it as well... :)

olafurpg:
I pushed a tag for v0.13.8-M1 and it seems to have worked
```
❯ cs resolve com.twitter:algebird-core_2.12:0.13.8-M1
com.googlecode.javaewah:JavaEWAH:1.1.7:default
com.twitter:algebird-core_2.12:0.13.8-M1:default
org.scala-lang:scala-library:2.12.12:default
org.scala-lang:scala-reflect:2.12.12:default
org.scala-lang.modules:scala-collection-compat_2.12:2.2.0:default
org.typelevel:algebra_2.12:2.0.0:default
org.typelevel:cats-kernel_2.12:2.0.0:default
```

I opened #871 to make it easier to write release notes. I'll follow up with similar changes in the chill and bijection repos, although it may take a few days until I get around to it.

regadas:
it worked! 🎉 however i thought you were releasing `v0.13.8`

regadas:
@johnynek thx for the acknowledgement 🙏

olafurpg:
You can do the non-milestone release, I just wanted to validate the pipeline was working :)

olafurpg:
Followup PRs

* https://github.com/twitter/chill/pull/507
* https://github.com/twitter/bijection/pull/460

PR #880, opened 2020-10-30
Use Kahan summation for Float and Double

johnynek:
see: https://en.wikipedia.org/wiki/Kahan_summation_algorithm

@sritchie reminded me of this recently, which I had seen years ago but forgotten when we implemented this code.

This improves the accuracy of long sums, which will make the results different, but floating point is already not associative or commutative, so I think this is an acceptable risk.

I had to change the tests a bit because scalas `.sum` does not use this kind of approach and it results in larger error.

Thanks Sam!

### DIFF
--- a/algebird-core/src/main/scala/com/twitter/algebird/Ring.scala
+++ b/algebird-core/src/main/scala/com/twitter/algebird/Ring.scala
@@ -151,16 +151,16 @@ object FloatRing extends Ring[Float] {
     else
       Some {
         val iter = t.toIterator
-        var sum = iter.next().toDouble
-        var c = 0.0
+        var sum = iter.next()
+        var c = 0.0f
         while (iter.hasNext) {
-          val y = iter.next().toDouble - c
+          val y = iter.next() - c
           val t = sum + y
           c = (t - sum) - y
           sum = t
         }
 
-        sum.toFloat
+        sum
       }
 }
 
--- a/algebird-core/src/main/scala/com/twitter/algebird/Ring.scala
+++ b/algebird-core/src/main/scala/com/twitter/algebird/Ring.scala
@@ -151,16 +151,16 @@ object FloatRing extends Ring[Float] {
     else
       Some {
         val iter = t.toIterator
-        var sum = iter.next()
-        var c = 0.0f
+        var sum = iter.next().toDouble
+        var c = 0.0
         while (iter.hasNext) {
-          val y = iter.next() - c
+          val y = iter.next().toDouble - c
           val t = sum + y
           c = (t - sum) - y
           sum = t
         }
 
-        sum
+        sum.toFloat
       }
 }
 
--- a/algebird-core/src/main/scala/com/twitter/algebird/Ring.scala
+++ b/algebird-core/src/main/scala/com/twitter/algebird/Ring.scala
@@ -144,14 +144,20 @@ object FloatRing extends Ring[Float] {
   override def minus(l: Float, r: Float): Float = l - r
   override def times(l: Float, r: Float): Float = l * r
 
+  // see: https://en.wikipedia.org/wiki/Kahan_summation_algorithm
+  // for this algorithm
   override def sumOption(t: TraversableOnce[Float]): Option[Float] =
     if (t.isEmpty) None
     else
       Some {
-        var sum = 0.0
         val iter = t.toIterator
+        var sum = iter.next().toDouble
+        var c = 0.0
         while (iter.hasNext) {
-          sum += iter.next().toDouble
+          val y = iter.next().toDouble - c
+          val t = sum + y
+          c = (t - sum) - y
+          sum = t
         }
 
         sum.toFloat
@@ -166,14 +172,20 @@ object DoubleRing extends Ring[Double] {
   override def minus(l: Double, r: Double): Double = l - r
   override def times(l: Double, r: Double): Double = l * r
 
+  // see: https://en.wikipedia.org/wiki/Kahan_summation_algorithm
+  // for this algorithm
   override def sumOption(t: TraversableOnce[Double]): Option[Double] =
     if (t.isEmpty) None
     else
       Some {
-        var sum = 0.0
         val iter = t.toIterator
+        var sum = iter.next()
+        var c = 0.0
         while (iter.hasNext) {
-          sum += iter.next()
+          val y = iter.next() - c
+          val t = sum + y
+          c = (t - sum) - y
+          sum = t
         }
 
         sum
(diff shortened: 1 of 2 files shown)


algebird-core/src/main/scala/com/twitter/algebird/Ring.scala
```diff
@@ -144,14 +144,20 @@ object FloatRing extends Ring[Float] {
...
 
+  // see: https://en.wikipedia.org/wiki/Kahan_summation_algorithm
+  // for this algorithm
   override def sumOption(t: TraversableOnce[Float]): Option[Float] =
     if (t.isEmpty) None
     else
       Some {
-        var sum = 0.0
         val iter = t.toIterator
+        var sum = iter.next().toDouble
```
sritchie, line 154: Any reason not to keep these as floats, and use `var c = 0.0f` for the aggregator?

algebird-test/src/test/scala/com/twitter/algebird/AggregatorLaws.scala
```diff
@@ -89,7 +89,11 @@ class AggregatorLaws extends CheckProperties {
   def checkNumericSum[T: Arbitrary](implicit num: Numeric[T]): Prop =
     forAll { in: List[T] =>
       val aggregator = Aggregator.numericSum[T]
-      aggregator(in) == in.map(num.toDouble).sum
```
sritchie, line 92: I added a few extra tests and I think I've uncovered... what MIGHT be a bigger problem here that I'll submit to your enormous pulsing brain for consideration.

Here are a few tests that check:

- that `input.sum` accumulates errors, as we expect
- that Kahan summation is able to control these errors

I did this by accumulating 1000 `0.01`s.

```scala
import org.scalatest.funsuite.AnyFunSuite

/**
  Unit tests to highlight specific examples of the properties we guarantee.
  */
class AggregatorTests extends AnyFunSuite {
  test("Kahan summation mitigates error accumulation in Double streams") {

    val input = Stream.continually(0.01).take(1000)

    assert(9.999999999999831 == input.sum, "naive summation accumulates errors")
    assert(Some(10.0) == DoubleRing.sumOption(input), "Kahan summation controls error accumulation")

    val aggregator = Aggregator.numericSum[Double]
    assert(10.0 == aggregator(input), "aggregators control error accumulation")
  }

  test("Kahan summation mitigates error accumulation in Float streams") {
    val input = Stream.continually(0.01f).take(1000)

    assert(10.0001335f == input.sum, "naive summation accumulates errors")
    assert(Some(10.0f) == FloatRing.sumOption(input), "Kahan summation controls error accumulation")

    val aggregator = Aggregator.numericSum[Float]
    assert(10.0f == aggregator(input), "agg control error accumulation")
  }
}
```

Here are the results:

```scala
[info] AggregatorTests:
[info] - Kahan summation mitigates error accumulation in Double streams
[info] - Kahan summation mitigates error accumulation in Float streams *** FAILED ***
[info]   10.0 did not equal 9.999999776482582 agg control error accumulation (AggregatorLaws.scala:46)
```

`FloatRing.sumOption` and `DoubleRing.sumOption` are both doing the right thing here, but `Aggregator.numericSum[Float]` is NOT doing the right thing, and accumulating its own errors such that the sum == `9.999999776482582` .

EDIT: I figured out later that called `.toFloat` on this value does indeed get you `10.0f` back. What's actually going on is that (as you know) `Aggregator.numericSum[Float]` is aggregating doubles, using the Kahan-enabled `DoubleRing`. I show in the tests I submitted in #881 that this is of course doing the right thing.

The initial `prepare` pass converts float to double, introducing what looks like error:

```scala
scala> val input = Stream.continually(0.01f).take(1000).map(implicitly[Numeric[Float]].toDouble(_))
input: scala.collection.immutable.Stream[Double] = Stream(0.009999999776482582, ?)

scala> input.map(implicitly[Numeric[Float]].toDouble(_)).take(5).toList
res3: List[Double] = List(0.009999999776482582, 0.009999999776482582, 0.009999999776482582, 0.009999999776482582, 0.009999999776482582)
```

Here's where the double conversion lives:

```scala
def numericSum[T](implicit num: Numeric[T]): MonoidAggregator[T, Double, Double] =
    Preparer[T].map(num.toDouble).monoidAggregate(Aggregator.fromMonoid)
```

But again this makes sense if you are aggregating floats into a double. Calling `toFloat` at the end fixes things up. See #881.

sritchie: Ah, here's the problem:

```scala
scala> val input = Stream.continually(0.01f).take(1000).map(implicitly[Numeric[Float]].toDouble(_))
input: scala.collection.immutable.Stream[Double] = Stream(0.009999999776482582, ?)

scala> input.map(implicitly[Numeric[Float]].toDouble(_)).take(5).toList
res3: List[Double] = List(0.009999999776482582, 0.009999999776482582, 0.009999999776482582, 0.009999999776482582, 0.009999999776482582)
```

So the double conversion introduces trouble here.

```scala
def numericSum[T](implicit num: Numeric[T]): MonoidAggregator[T, Double, Double] =
    Preparer[T].map(num.toDouble).monoidAggregate(Aggregator.fromMonoid)
```

Here's a better set of tests from `Aggregator.fromMonoid` that pass:

```scala
import org.scalatest.funsuite.AnyFunSuite

/**
  Unit tests to highlight specific examples of the properties we guarantee.
  */
class AggregatorTests extends AnyFunSuite {
  test("Kahan summation mitigates Double error accumulation") {
    val input = Stream.continually(0.01).take(1000)

    assert(9.999999999999831 == input.sum, "naive summation accumulates errors")
    assert(Some(10.0) == DoubleRing.sumOption(input), "Kahan summation controls error accumulation")
    assert(10.0 == Aggregator.numericSum[Double].apply(input))
  }

  test("Kahan summation mitigates Float error accumulation") {
    val input = Stream.continually(0.01f).take(1000)

    assert(10.0001335f == input.sum, "naive summation accumulates errors")
    assert(Some(10.0f) == FloatRing.sumOption(input), "Kahan summation controls error accumulation")
    assert(10.0f == Aggregator.fromMonoid[Float].apply(input))
  }
}
```

BTW, with `import Ring.numericRing` the float tests fail, since `NumericRing[Float]` doesn't get access to the Kahan summation trick.

The docs advise

```scala
* Note that if you instead wanted to aggregate Numeric values of a type T into the same type T
   * (e.g. if you want MonoidAggregator[T, T, T] for some Numeric type T), you can directly use
   * Aggregator.fromMonoid[T] after importing the numericRing implicit:
   *
   *   > import com.twitter.algebird.Ring.numericRing
   *   > def numericAggregator[T: Numeric]: MonoidAggregator[T, T, T] = Aggregator.fromMonoid[T]
```

Which is maybe not the correct advice.

sritchie (changes requested):
Hey, my comments are a little stream-of-consciousness as I was investigating, so here's the summary:

- I wrote some extra unit tests (See #881) that show how this trick prevents error accumulation. Would be awesome if you'd add those in.

Nice work!

algebird-core/src/main/scala/com/twitter/algebird/Ring.scala
sritchie: I made this change in #881, take it or leave it!

sritchie:
:shipit: looks good to me

PR #881, opened 2020-10-31
Tests for #880, PR review suggestions

sritchie:
My suggested tests for the Kahan summation PR @johnynek wrote up in #848.

@johnynek, I also backed off the `toDouble`, `toFloat`. I would have expected, based on this, that the tests would fail by turning all of the intermediates into doubles!

```scala
scala> val input = Stream.continually(0.01f).take(1000).map(implicitly[Numeric[Float]].toDouble(_))
input: scala.collection.immutable.Stream[Double] = Stream(0.009999999776482582, ?)

scala> input.map(implicitly[Numeric[Float]].toDouble(_)).take(5).toList
res3: List[Double] = List(0.009999999776482582, 0.009999999776482582, 0.009999999776482582, 0.009999999776482582, 0.009999999776482582)
```

But it turns out that the `toFloat` at the end "fixes" these errors. I added a third test to make this more clear.

### DIFF
--- a/algebird-core/src/main/scala/com/twitter/algebird/Ring.scala
+++ b/algebird-core/src/main/scala/com/twitter/algebird/Ring.scala
@@ -151,16 +151,16 @@ object FloatRing extends Ring[Float] {
     else
       Some {
         val iter = t.toIterator
-        var sum = iter.next().toDouble
-        var c = 0.0
+        var sum = iter.next()
+        var c = 0.0f
         while (iter.hasNext) {
-          val y = iter.next().toDouble - c
+          val y = iter.next() - c
           val t = sum + y
           c = (t - sum) - y
           sum = t
         }
 
-        sum.toFloat
+        sum
       }
 }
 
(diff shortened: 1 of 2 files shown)


algebird-core/src/main/scala/com/twitter/algebird/Ring.scala
```diff
@@ -151,16 +151,16 @@ object FloatRing extends Ring[Float] {
     else
       Some {
         val iter = t.toIterator
-        var sum = iter.next().toDouble
-        var c = 0.0
```
johnynek: Why not use Double here? CPUs aren't any faster at float than they are double for basic operations are they? Can't it in principle avoid error when you have overflow in floats that wouldn't overflow a Double?

Also note: the code did this originally, so this would be a change from the current branch not just my PR.

johnynek (commented):
Thanks a ton for sending these tests!

I did have one question about why to remove the widening to Double.

algebird-core/src/main/scala/com/twitter/algebird/Ring.scala
sritchie: Oh, I just did the change to see if it mattered - I was more tossing this in to accompany my question to you about if / why it mattered :) that is a good call too about avoiding overflow in principle. I’ll revert this now so you can merge cleanly

sritchie:
@johnynek okay, reverted! This is ready to go.

johnynek:
Thanks a ton!

PR #913, opened 2021-01-16
Update scala 2.12 and 2.13

regadas:


regadas:
build fails due to `scoverage` https://github.com/scoverage/sbt-scoverage/issues/319

johnynek:
Huh, the micro site is failing somehow. I haven't checked the logs.

regadas:
Right, the error is related to `tut` which is removed in #920

johnynek:
so, if we merge #920 this should be good to go?

.github/workflows/ci.yml
```diff
@@ -60,7 +60,7 @@ jobs:
           - 11
         scala:
           - 2.11.12
-          - 2.12.12
+          - 2.12.13
```
johnynek, line 63: shouldn't we add 2.13 to this list?

regadas: @johnynek sorry for the late reply, hard couple of days. Indeed! I guess we forgot to add it when we release 2.13.x.

johnynek (approved):
This is great.

Issue #326, opened 2014-07-08
Some HyperLogLog tests are not testing what we think.


johnynek:
https://github.com/twitter/algebird/blob/develop/algebird-test/src/test/scala/com/twitter/algebird/HyperLogLogTest.scala#L100

The code is probably right, but these tests are misleading. They are making very peculiar input sets (either complete or very spares in two cases). This is unusual. We should consider drawing sets according to a subsample of a powerlaw distribution, both on membership probability as well as set size.

PR #923, opened 2021-02-10
Issue #326 - Some HyperLogLog tests are not testing what we think. 

enricoap:
Issue #326 
### Proposed Changes for the jRhoW test

- the size of the Array[Byte] in input in input of the jRhoW function is now dependent on the number of bits.


johnynek:
Thanks for the PR.

Can you run scalafmt at the sbt prompt?

I'm happy to merge this. I guess previously we just never hit a case that had to go very far into the array.

enricoap:
Sorry, I just saw this comment now. Will do!

johnynek (approved):

johnynek:
Thank you!

PR #1008, opened 2021-09-22
expose maximum number of items and ordering in PQ

johnynek:
If you have a PriorityQueue monoid, there is no reason to hide the ordering and the maximum number of items. Both of which could be discovered from the values.

### DIFF
--- a/algebird-core/src/main/scala/com/twitter/algebird/mutable/PriorityQueueMonoid.scala
+++ b/algebird-core/src/main/scala/com/twitter/algebird/mutable/PriorityQueueMonoid.scala
@@ -26,17 +26,23 @@ import java.util.PriorityQueue
  */
 class PriorityQueueMonoid[K](max: Int)(implicit ord: Ordering[K]) extends Monoid[PriorityQueue[K]] {
 
-  require(max > 0, "PriorityQueueMonoid requires keeping at least 1 item")
+  def maximumItems: Int = max
+  def ordering: Ordering[K] = ord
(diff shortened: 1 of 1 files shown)


johnynek:
it would be nice to see the formatting diff in the error. I made this PR on the web (since this machine can't easily check out etc...), but I can't see or guess what is wrong with the formatting.

cc @regadas

algebird-core/src/main/scala/com/twitter/algebird/mutable/PriorityQueueMonoid.scala
```diff
@@ -26,28 +26,39 @@ import java.util.PriorityQueue
...
+  def maximumItems: Int = max
+  def ordering: Ordering[K] = ord
+  private[this] val revOrd: Ordering[K] = ord.reverse
+
+  require(max > 0, s"PriorityQueueMonoid requires keeping at least 1 item, invalid max=$max")
   // Java throws if you try to make a queue size 0
   protected val MINQUEUESIZE = 1
   def build(k: K): PriorityQueue[K] = {
-    val q = new PriorityQueue[K](1, ord.reverse);
+    val q = new PriorityQueue[K](1, revOrd);
```
regadas, line 37:
```suggestion
    val q = new PriorityQueue[K](MINQUEUESIZE, revOrd);
```

regadas (approved):

tlazaro:
LGTM, will help us on Scalding work.

PR #1036, opened 2021-12-10
Update scalafmt with runner.dialect

regadas:


johnynek:
great!

PS: man it is annoying how often scalafmt has been changing and generally requiring updates to adopt.

PR #1049, opened 2022-01-11
More efficient computation in MomentsMonoid

sritchie:
I was staring at this code and noticed that we could make it faster by avoiding a few divisions. So here we go!

Also, I wasn't sure where to put this, but this - and the mutable equivalent - are going to be faster than the provided `Aggregator` in the case of a scan. Any interest?

```scala
  def plusDouble(a: Moments, b: Double): Moments = {
    val countCombined = a.m0D + 1
    val delta = b - a.mean
    val delta_n = delta / countCombined
    val delta_n2 = delta_n * delta_n
    val term1 = delta * delta_n * a.m0D
    val meanCombined = Moments.getCombinedMeanDouble(a.m0D, a.mean, 1.0, b)

    // writing them in reverse order like this will be useful in the mutable case, but doesn't matter here.
    val m4 = a.m4 + term1 * delta_n2 * (n * n - 3 * n + 3) + 6 * delta_n2 * a.m2 - 4 * delta_n * a.m3
    val m3 = a.m3 + term1 * delta_n * (n - 2) - 3 * delta_n * a.m2
    val m2 = a.m2 + term1

    new Moments(countCombined, meanCombined, m2, m3, m4)
  }
```

### DIFF
--- a/algebird-core/src/main/scala/com/twitter/algebird/MomentsGroup.scala
+++ b/algebird-core/src/main/scala/com/twitter/algebird/MomentsGroup.scala
@@ -208,21 +208,23 @@ class MomentsMonoid extends Monoid[Moments] with CommutativeMonoid[Moments] {
     if (countCombined == 0.0) zero
     else {
       val delta = b.mean - a.mean
+      val delta_n = delta / countCombined
+      val delta_n2 = delta_n * delta_n
+      val delta_n3 = delta_n2 * delta_n
+
       val meanCombined = Moments.getCombinedMeanDouble(a.m0D, a.mean, b.m0D, b.mean)
 
-      val m2 = a.m2 + b.m2 +
-        math.pow(delta, 2) * a.m0D * b.m0D / countCombined
+      val m2 = a.m2 + b.m2 + delta * delta_n * a.m0D * b.m0D
 
       val m3 = a.m3 + b.m3 +
-        math.pow(delta, 3) * a.m0D * b.m0D * (a.m0D - b.m0D) / math.pow(countCombined, 2) +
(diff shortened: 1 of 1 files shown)


johnynek (approved):
This is great!

Thanks for taking the time to do this.

One follow up if you are interested:

replace `math.pow(x, 2)` with `final def square(x: Double): Double = x * x` which should be inlined, but the `pow(x, 2)` looks dynamic I fear to the JIT and it may not be able to inline it as one multiplication.

johnynek:
As for the `plusDouble` I think a mutable Fold is the right way to do that since you can can allocate some mutable state to hold your variables `class MutMoments { var count: Double , ... }` then implement your `plusMoments` function there.

that way, you wouldn't allocate on each call, but you would wind up boxing the doubles (since these types are generic) so I don't know if that is going to kill the win (since boxing into a Moments is maybe almost as cheap).

But you could also add `plusDouble` to `Moments` and just add a law that `m.plusDouble(x) == m.plus(Moments(x))`.

That said, you would have to allocate a new `Moments` on each return anyway... Seems to me maybe the real win would be maybe making a `MomentsState` class that is mutable, and you could add a `Moments` into it, or a `Double`, and the only thing you can do with a `MomentsState` is convert to `Moments` (or from Moments).

so, something like:

```scala
final class MomentsState {
  // has private vars that keep track of the moments
  // we can then use this state in sumOption or in a mutable Fold.
  def +(m: Moments): this.type
  def +(d: Double): this.type
  
  def toMoments: Moments
  def resetFromMoments(m: Moments): this.type
}

object MomentsState {
  def newEmpty: MomentsState
}
```

sritchie:
Done in #1050

PR #1050, opened 2022-01-13
Mutable MomentsState, more efficient + for double added to Moments

sritchie:
This PR:

- fixes an issue with `build.sbt` preventing my local jdk17 from compiling
- adds `+` methods to `Moments` allowing addition with other `Moments` or doubles
- adds `MomentsState` for efficient mutable accumulation of a `Moments` instance from iterables of either doubles or other moments instances

`MomentsState` pulls the former mutable implementation from `sumOption` and adds a second one for mutably adding in new doubles.


algebird-core/src/main/scala/com/twitter/algebird/MomentsGroup.scala
```diff
@@ -53,7 +53,55 @@ sealed class Moments(val m0D: Double, val m1: Double, val m2: Double, val m3: Do
...
+      6 * delta_n2 * m2 - 4 * delta_n * m3
+
+    new Moments(n, meanCombined, m2combined, m3combined, m4combined)
+  }
+
+  /**
+    * Returns a [[Fold]] instance that uses `+` to accumulate deltas into this
+    * [[Moments]] instance.
+    */
+  def fold: Fold[Double, Moments] =
```
sritchie, line 93: This is the style we used in `ExpHist`, where the fold is an instance method that lets you fold into THIS `Moments`, implemented by

```scala
_ => Moments.MomentsState.fromMoments(this)
```

Let me know if you want to just have an empty fold on `Moments`.

algebird-core/src/main/scala/com/twitter/algebird/MomentsGroup.scala
```diff
@@ -102,6 +150,92 @@ sealed class Moments(val m0D: Double, val m1: Double, val m2: Double, val m3: Do
...
+        mean = 0.0
+        m2 = 0.0
+        m3 = 0.0
+        m4 = 0.0
+      } else {
+        val delta = b.mean - mean
+        val delta_n = delta / countCombined
+        val delta_n2 = delta_n * delta_n
+        val delta_n3 = delta_n2 * delta_n
+        val count_sq = count * count
```
sritchie, line 175: I also killed the `math.pow` calls everywhere.

algebird-core/src/main/scala/com/twitter/algebird/MomentsGroup.scala
```diff
@@ -111,6 +245,8 @@ object Moments {
 
   val aggregator: MomentsAggregator.type = MomentsAggregator
 
+  val fold: Fold[Double, Moments] = momentsMonoid.zero.fold
```
sritchie, line 248: here we go, the zero fold.

algebird-core/src/main/scala/com/twitter/algebird/MomentsGroup.scala
```diff
@@ -233,63 +371,12 @@ class MomentsMonoid extends Monoid[Moments] with CommutativeMonoid[Moments] {
   override def sumOption(items: TraversableOnce[Moments]): Option[Moments] =
     if (items.isEmpty) None
     else {
+      val state = Moments.MomentsState.newEmpty
       val iter = items.toIterator
-
```
sritchie, line 237: this implementation lives in `MomentsState` now.

algebird-test/src/test/scala/com/twitter/algebird/MomentsLaws.scala
```diff
@@ -100,7 +132,7 @@ class MomentsTest extends AnyWordSpec with Matchers {
    * Given a list of doubles, create a Moments object to hold the list's central moments.
    */
   def getMoments(xs: List[Double]): Moments =
-    MomentsAggregator(xs)
+    Moments.aggregator(xs)
```
sritchie, line 133: I switched this so we get coverage on the `aggregator` line :)

build.sbt
```diff
@@ -69,7 +69,7 @@ val sharedSettings = Seq(
         Nil
     }
   },
-  javacOptions ++= Seq("-target", "1.6", "-source", "1.6"),
+  javacOptions ++= Seq("-target", "1.8", "-source", "1.8"),
```
sritchie, line 72: here is the build change.

algebird-core/src/main/scala/com/twitter/algebird/MomentsGroup.scala
```diff
@@ -102,6 +146,96 @@ sealed class Moments(val m0D: Double, val m1: Double, val m2: Double, val m3: Do
...
+      m4 = m.m4
+      this
+    }
+  }
+
+  object MomentsState {
+    def fromMoments(m: Moments): MomentsState =
+      new MomentsState(m.m0D, m.m1, m.m2, m.m3, m.m4)
+
+    def newEmpty: MomentsState =
```
johnynek: nit: scala has developed a style of using `def newEmpty(): MomentsState` for non-pure things (in this case we are allocating mutable memory, so it is not pure, and the `newEmpty()` helps signal that.

algebird-core/src/main/scala/com/twitter/algebird/MomentsGroup.scala
johnynek: Huh... looking at it now, it seems a bit strange to me, but I can see the benefit of being able to start with a non empty value to resume some updates.

algebird-core/src/main/scala/com/twitter/algebird/MomentsGroup.scala
```diff
@@ -102,6 +146,96 @@ sealed class Moments(val m0D: Double, val m1: Double, val m2: Double, val m3: Do
...
+      var m4: Double
+  ) {
+
+    def +=(b: Moments): this.type = {
+      /*
+       * Unfortunately we copy the code from the monoid's plus implementation,
+       * but we do it to avoid allocating a new Moments on every item in the
+       * loop. the Monoid laws test that sum matches looping on plus
+       */
+      val countCombined = count + b.m0D
```
johnynek, line 163: I guess we could use a variant of Kahan summation on countCombined, right

algebird-core/src/main/scala/com/twitter/algebird/MomentsGroup.scala
```diff
@@ -53,7 +53,51 @@ sealed class Moments(val m0D: Double, val m1: Double, val m2: Double, val m3: Do
     math.sqrt(m0D) * m3 / math.pow(m2, 1.5)
```
johnynek: I wonder if this is slower than `math.sqrt(m0D) * m3 / (m2 * math.sqrt(m2))`

or even: `m3 / (m2 * stddev)`

seems like maybe sqrt and div are about the same:

https://stackoverflow.com/questions/6884359/c-practical-computational-complexity-of-cmath-sqrt

so I guess the `m3 / (m2 * stddev)` formula would be best because it is clearest. What do you think?

algebird-core/src/main/scala/com/twitter/algebird/MomentsGroup.scala
```diff
@@ -234,62 +371,19 @@ class MomentsMonoid extends Monoid[Moments] with CommutativeMonoid[Moments] {
...
-            delta * delta_n3 * count * b.m0D * (math.pow(count, 2) -
-              count * b.m0D + math.pow(b.m0D, 2)) +
-            6 * delta_n2 * (math.pow(count, 2) * b.m2 + math.pow(b.m0D, 2) * m2) +
-            4 * delta_n * (count * b.m3 - b.m0D * m3)
-
-          mean = meanCombined
-          m2 = m2Combined
-          m3 = m3Combined
-          m4 = m4Combined
+      // If there is only a single item, skip the MomentsState instantiation and
```
johnynek, line 376: this is nice, BTW.

algebird-test/src/test/scala/com/twitter/algebird/MomentsLaws.scala
```diff
@@ -75,6 +74,37 @@ class MomentsLaws extends CheckProperties {
...
+    }
+  }
+
+  property("adding Moment instances via +, sumOption should match") {
+    forAll(opGen, Gen.containerOf[Seq, Double](Gen.choose(0, 1000))) { (mom, ints) =>
+      val xs = ints.map(Moments(_)).toTraversable
+      val monoid = Moments.momentsMonoid
+
+      val fullViaAdd = xs.foldLeft(mom)(_ + _)
+      val fullViaMonoid = mom + monoid.sumOption(xs).getOrElse(monoid.zero)
```
johnynek: nit: you can do `monoid.sum(xs)` here.

algebird-core/src/main/scala/com/twitter/algebird/MomentsGroup.scala
sritchie: Yup, and if we had the fold version, or `KahanState`, it would fit nicely here.

algebird-core/src/main/scala/com/twitter/algebird/MomentsGroup.scala
sritchie: I agree, that's definitely clearest!

algebird-test/src/test/scala/com/twitter/algebird/MomentsLaws.scala
sritchie: whoopsies!

sritchie:
Okay, fixed up @johnynek !

sritchie:
Benchmark results... I'm a little surprised at the winner here.

```
[info] Benchmark                         (numElements)   Mode  Cnt    Score   Error  Units
[info] MomentsBenchmark.timeAggregate            10000  thrpt    3   7968.580 ±  625.872  ops/s
[info] MomentsBenchmark.timeFold                 10000  thrpt    3  12852.096 ±  760.259  ops/s
[info] MomentsBenchmark.timePlusDoubles          10000  thrpt    3  15150.855 ± 743.752  ops/s
[info] MomentsBenchmark.timePlusMoments          10000  thrpt    3  13478.708 ± 675.039  ops/s
[info] MomentsBenchmark.timeSumOption            10000  thrpt    3  13576.123 ± 1111.999  ops/s
```

Wow, and here are the benchmarks that could run BEFORE this PR and #1049. Optimizing `plus` really matters!!

```
[info] Benchmark                         (numElements)   Mode  Cnt    Score   Error  Units
[info] MomentsBenchmark.timeAggregate            10000  thrpt    3  227.477 ± 2.977  ops/s
[info] MomentsBenchmark.timePlusMoments          10000  thrpt    3  223.974 ± 7.554  ops/s
[info] MomentsBenchmark.timeSumOption            10000  thrpt    3  226.067 ± 3.714  ops/s
```

johnynek (approved):

Issue #1122, opened 2023-06-30
SketchMap is not commutative 


Peilin-Yang:
**Describe the bug**

**To Reproduce**
1. Add the following new test to the unit test `algebird/algebird-test/src/test/scala/com/twitter/algebird/SketchMapTest.scala` after line 57
```
    "plus should work commutatively" in {
      implicit val m = Monoid.longMonoid
      val valueTableLeft =
        DenseMatrix(
          PARAMS.width,
          PARAMS.depth,
          rowsByColumns = IndexedSeq(10L) ++ (1 until PARAMS.width * PARAMS.depth).map(_ => 0L)
        )
      val testLeft = SketchMap[Int, Long](
        valuesTable = valueTableLeft,
        heavyHitterKeys = List(1),
        totalValue = 10
      )

      val valueTableRight = SparseColumnMatrix(rowsByColumns =
        IndexedSeq(
          SparseVector(
            map = Map(1 -> 1L),
            sparseValue = 1L,
            length = PARAMS.width * PARAMS.depth
          )
        )
      )
      val testRight = SketchMap[Int, Long](
        valuesTable = valueTableRight,
        heavyHitterKeys = List(1),
        totalValue = 1
      )

      assert(MONOID.plus(testLeft, testRight) == MONOID.plus(testRight, testLeft))
    }
```
3. Run the test and it would fail

**Expected behavior**

The test is expected to pass

**Screenshots**
N/A

**Environment**
N/A

**Additional context**
This is mainly because the `algebird/algebird-core/src/main/scala/com/twitter/algebird/matrix/AdaptiveMatrix.scala` doesn't handle the (SparseColumnMatrix, DenseMatrix) input (in that order) correctly.

PR #1123, opened 2023-06-30
Simple Fix of SketchMap's non-commutative (issue-1122)

Peilin-Yang:
This PR introduces a simple fix to SketchMap's non-commutative, as illustrated in https://github.com/twitter/algebird/issues/1122 

The fix is simply to always put the DenseMatrix as the first param so that the update to it is correct.

### DIFF
--- a/algebird-core/src/main/scala/com/twitter/algebird/SketchMap.scala
+++ b/algebird-core/src/main/scala/com/twitter/algebird/SketchMap.scala
@@ -18,6 +18,7 @@ package com.twitter.algebird
 
 import algebra.CommutativeMonoid
 import com.twitter.algebird.matrix.AdaptiveMatrix
+import com.twitter.algebird.matrix.DenseMatrix
 
 /**
  * A Sketch Map is a generalized version of the Count-Min Sketch that is an approximation of Map[K, V] that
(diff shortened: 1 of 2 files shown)


algebird-core/src/main/scala/com/twitter/algebird/SketchMap.scala
```diff
@@ -50,7 +51,10 @@ class SketchMapMonoid[K, V](val params: SketchMapParams[K])(implicit
     SketchMap(AdaptiveMatrix.fill(params.depth, params.width)(monoid.zero), Nil, monoid.zero)
 
   override def plus(left: SketchMap[K, V], right: SketchMap[K, V]): SketchMap[K, V] = {
-    val newValuesTable = Monoid.plus(left.valuesTable, right.valuesTable)
+    val newValuesTable = right.valuesTable match {
+      case DenseMatrix(_, _, _) => Monoid.plus(right.valuesTable, left.valuesTable)
+      case _                    => Monoid.plus(left.valuesTable, right.valuesTable)
+    }
```
regadas, line 57:
```suggestion
    val newValuesTable = right.valuesTable match {
      case _: DenseMatrix[V] => Monoid.plus(right.valuesTable, left.valuesTable)
      case _                    => Monoid.plus(left.valuesTable, right.valuesTable)
    }
```

Also, perhaps it would make more sense to address this here https://github.com/twitter/algebird/blob/65a95ce440b9c30101db5528cf1a90ec02361142/algebird-core/src/main/scala/com/twitter/algebird/matrix/AdaptiveMatrix.scala#L56

algebird-test/src/test/scala/com/twitter/algebird/SketchMapTest.scala
```diff
@@ -53,6 +55,38 @@ class SketchMapTest extends AnyWordSpec with Matchers {
       assert(sm.totalValue == totalCount)
     }
 
+    "plus should work commutatively" in {
+      implicit val m = Monoid.longMonoid
```
regadas, line 59: I believe this is not required

regadas (commented):
thanks for the PR @Peilin-Yang

algebird-core/src/main/scala/com/twitter/algebird/SketchMap.scala
Peilin-Yang: > Also, perhaps it would make more sense to address this here

yeah, pretty much agreed. 
I was afraid to touch `AdaptiveMatrix` due to unfamiliarity of it, but will try to see if I can fix that directly.

